PR #8, opened 2023-01-11
Incorporate develop updates into main

douglasjacobsen:
This merge is to bring main and develop inline with each other.

This primarily fixes an issue earlier where a merge happened into main that was not in develop.


rfbgo (approved):
We should probably figure out some more formal structure around release to main in the future, but LGTM

PR #9, opened 2023-01-11
Add support for spack flags to be set in config files

douglasjacobsen:
This merge adds the tests and functionality for configuring flags for spack actions.


rfbgo (approved):
LGTM! 

We chatted about this offline, and agree the change is non breaking

PR #14, opened 2023-01-14
Add code coverage to PR validation

douglasjacobsen:


rfbgo (approved):
LGTM

We should capture somewhere that we want to make this "fail" under certain conditions not just report coverage

Issue #15, opened 2023-01-15
Software installation cache does not reflect spec_names


douglasjacobsen:
### Steps to reproduce

Using the full expansion example from (https://github.com/GoogleCloudPlatform/ramble/blob/develop/examples/full_expansion_config.yaml), the cache prevents openfoam-zen2 from installing.

This is because the cache houses `expander.workload_namespace`. And workload_namespace is defined as: `{application_name}.{workload_name}`.

This should either be:
`{spec_name}.{workload_name}`
or:
`{application_name}.{spec_name}.{workload_name}`

But, it needs `{spec_name}` in it to be fully unique. So, either we either need to change what is cached, or we need to change the definition of `{workload_namespace}`. Neither really has any consequences currently, but one of the two is necessary for the cache to function properly.

### Information on your system

Not relevant to this bug.

### General information

- [X] I have run `ramble debug report` and reported the version of Ramble/Python/Platform
- [X] I have searched the issues of this repo and believe this is not a duplicate
- [X] I have run the failing commands in debug mode and reported the output

PR #16, opened 2023-01-15
Add support for auxiliary software files

douglasjacobsen:
This merge adds support for $workspace/configs/auxiliary_software_files to contain arbitrary files which different application types can utilize.


douglasjacobsen:
This is currently blocked by #15 

Once we have a fix in, I'll rebase, which should fix the failing tests.

rfbgo (changes requested):
LGTM

As discussed let's add a little note/comment somewhere to help external readers to understand why we wanted to add this

douglasjacobsen:
Made issue: #23  to track the documentation bit. We'll add some text to that (eventual) documentation that says something like:

```
Workspaces allow auxiliary software files to be created in:
  $workspace_root/configs/auxiliary_software_files
  
These files will be copied into the Ramble managed software directories. As an example, these directories take the form of:
  $workspace_root/software/{spec_name}.{workload_name}

Auxiliary software files can be useful when an experiment requires additional configuration for the package manager. An example of this is when using spack on shared systems. Sometimes an experiment needs a unique package.yaml file to tell spack about the external dependencies. However, these files are not limited to spack config files, and any file placed in this directory is copied, to help improve compatibility with other package mangers.
```

rfbgo (approved):
LGTM, thanks for the changes!

PR #20, opened 2023-01-18
Change cache to use spec_namespace

### LINKED ISSUE #15: Software installation cache does not reflect spec_names

douglasjacobsen:
This merge adds spec_namespace as a property for the expander. This namespace is defined as: `'{spec_name}.workload_name'`

Fixes #15

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -239,6 +239,14 @@ class Expander(object):
             return '%s.%s' % (app_name, wl_name)
         return None
 
+    @property
+    def spec_namespace(self):
+        app_name = self.expand_var('{spec_name}')
+        wl_name = self.workload_name
+        if app_name and wl_name:
+            return '%s.%s' % (app_name, wl_name)
+        return None
+
     def set_experiment(self, exp_name):
         tty.debug('Expander: Setting exp name %s' % exp_name)
         if not self.workload_name:
--- a/lib/ramble/ramble/application_types/spack.py
+++ b/lib/ramble/ramble/application_types/spack.py
@@ -8,7 +8,7 @@
 
 import llnl.util.tty as tty
 
-from ramble.application import ApplicationBase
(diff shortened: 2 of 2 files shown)


lib/ramble/ramble/expander.py
```diff
@@ -239,6 +239,14 @@ def workload_namespace(self):
             return '%s.%s' % (app_name, wl_name)
         return None
 
+    @property
+    def spec_namespace(self):
+        app_name = self.expand_var('{spec_name}')
+        wl_name = self.workload_name
+        if app_name and wl_name:
+            return '%s.%s' % (app_name, wl_name)
+        return None
```
rfbgo, line 248: Do we want to check the returned value at the call sight? I'm not sure if it's actually possible, but a bug where we cache `('spack', None)` as meaning different iterations are the same doesn't sound fun to track down later. 

(given it's based on app and wl name being falsey, that seems unlikely but might be best to be safe?)

rfbgo (changes requested):

lib/ramble/ramble/expander.py
rfbgo: Thanks!

rfbgo (approved):
LGTM!

PR #21, opened 2023-01-18
Add a `debug` command

douglasjacobsen:
This merge creates a (previously referenced, but missing) `debug` command.


rfbgo (approved):
Good catch!

PR #27, opened 2023-01-31
Add create of simlink results.latest.txt during analysis

rfbgo:


lib/ramble/ramble/test/end_to_end.py
```diff
@@ -239,12 +239,14 @@ def test_wrfv4_dry_run():
 
         output = workspace('analyze', '-f', 'text',
                            'json', 'yaml', global_args=['-w', workspace_name])
+        text_simlink_results_files = glob.glob(os.path.join(ws1.root, 'results.latest.txt'))
```
douglasjacobsen, line 242: It might be nice to add a test that if `workspace analyze` is run multiple times, `results.latest.txt` still points to the (actual) latest file. Maybe we should add a task for tracking this.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1059,18 +1059,46 @@ def append_result(self, result):
...
 
+    def simlink_result(self, filename_base, latest_base, file_extension, overwrite=True):
+        """
+        Create simlink of result file so that results.latest.txt always points
+        to the most recent analysis. This clobbers the existing link
+        """
+        out_file = os.path.join(self.root, filename_base + file_extension)
+        latest_file = os.path.join(self.root, latest_base + file_extension)
+
+        if overwrite and os.path.exists(latest_file):
```
douglasjacobsen, line 1070: It might be good to check `os.is_symlink`, and provide a CLI argument to control the overwrite behavior.

It would be sad to clobber some file the user named made `results.latest.txt` (within the workspace). Admittedly though, this is pretty unlikely.

douglasjacobsen (changes requested):

lib/ramble/ramble/workspace/workspace.py
rfbgo: I initially thought we might want to expose the overwrite, but on reflection I don't think it's worth it (especially because you might want a 3-way semantic around clobber-none, clobber-symlinks, and clobber-all...)

Another thing I considered is only clobbering simlinks (instead of all files), I was 50/50 on which is better but eventually decided that given our original semantic was clobbering `results.txt` then clobbering files is fine too (and have no idea how simlinks on windows like platforms work..) 

It sounds like you prefer only clobbering if simlink?

lib/ramble/ramble/test/end_to_end.py
rfbgo: Good idea. I filed one onto our backlog

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: Yeah, I think I'd prefer it only clobber if it is a symlink. I think it's really unlikely that this will actually impact users, but I think I'd rather not accidentally clobber some file named `results.latest.txt` that the user copied in for some reason.

rfbgo: Pushed an update to address this

douglasjacobsen (approved):

PR #31, opened 2023-02-02
Fix incorrectly retained vars between experiments

### LINKED ISSUE #30: Expander deduced variables are wrong in vector/matrix expanded experiments
### Steps to reproduce

Make a workspace with a config that sets ppn/n_nodes and allows the deduction of `n_ranks`:

```
ramble:
  mpi:
    command: mpirun
    args:
      - "-np {n_ranks}"
  batch:
    submit: '{execute_experiment}'
  variables:
    processes_per_node: 56
  applications:
    hostname:
      workloads:
        parallel:
          experiments:
            with_name_{exp_name}_{n_nodes}:
              variables:
                exp_name: test
                n_nodes: [1, 2]
```

### Error message

This is actually easier to detect if you set the first value in `n_nodes` to be bigger then subsequent values and it triggers a sanity check in the expander: 

with `n_nodes: [2, 1]` gives:
```
==> Error: n_nodes in hostname.parallel.with_name_test_2 is 1 and should be 2
```
from
```
546         if n_ranks and ppn:
547             test_n_nodes = math.ceil(int(n_ranks) / int(ppn))
548
549             if n_nodes and n_nodes < test_n_nodes:
550                 tty.error('n_nodes in %s is %s and should be %s' %
551                           (self.experiment_namespace, n_nodes,
552                            test_n_nodes))
```

The basic problem is that the expander retains calculated values, and since we re-use the expander (and other) object for different experiments it pollutes the next one

A good long term solution to this would be to have each experiment (implemented as an instance an an application?) be fully independent, but for now it's sufficient to ensure the expander gets reset appropriately 

* **Ramble:** 0.0.1 (238696cd7f36caa0c54a9780874726cef961ad0e)
* **Python:** 3.10.7
* **Platform:** darwin-macos-m1

### General information

rfbgo:
Unsets deduced vars to allow them to fully re read from the experiment setup

Fixes #30


douglasjacobsen (changes requested):
LGTM.

There's a flake8 error in the PR validation though:
```
Step #2 - "ramble-tests": flake8: running flake8 code checks on ramble.
Step #2 - "ramble-tests": 
Step #2 - "ramble-tests": Modified files:
Step #2 - "ramble-tests":   lib/ramble/ramble/expander.py
Step #2 - "ramble-tests":   lib/ramble/ramble/workspace/workspace.py
Step #2 - "ramble-tests": =======================================================
Step #2 - "ramble-tests": lib/ramble/ramble/expander.py:448: [E303] too many blank lines (2)
```

douglasjacobsen (approved):

PR #32, opened 2023-02-03
Fix empty expvars

rfbgo:
Previously the following `ramble.yaml` would give a hard to understand error, and users were unable to not specify experiment level variables (which are not always needed):

```
 31   applications:
 32     hostname:
 33       workloads:
 34         parallel:
 35           experiments:
 36             with_name_{exp_name}_{n_nodes}:
 37                 {}
```

This PR allows for an empty value to be passed and a default empty dict to be created

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -746,11 +746,14 @@ class Workspace(object):
         # Iterate over applications in ramble.yaml first
         if application_namespace in ws_dict[ramble_namespace]:
             app_dict = ws_dict[ramble_namespace][application_namespace]
+
             for application, contents in app_dict.items():
                 application_vars = None
                 application_env_vars = None
+
                 if variables_namespace in contents:
                     application_vars = contents[variables_namespace]
+
                 if env_var_namespace in contents:
                     application_env_vars = contents[env_var_namespace]
                 self.extract_success_criteria('application', contents)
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -817,12 +820,18 @@ def all_experiments(self, workload):
 
         experiments = workload[experiment_namespace]
         for experiment, contents in experiments.items():
-            experiment_vars = None
+
+            # FIXME: what is the cleanest way to do this?
+            import ruamel
```
douglasjacobsen, line 825: I think we should use `spack.util.spack_yaml` instead of ruamel.

douglasjacobsen (changes requested):

douglasjacobsen:
There's also a flake8 error here:
```
Step #2 - "ramble-tests": =======================================================
Step #2 - "ramble-tests": flake8: running flake8 code checks on ramble.
Step #2 - "ramble-tests": 
Step #2 - "ramble-tests": Modified files:
Step #2 - "ramble-tests":   lib/ramble/ramble/expander.py
Step #2 - "ramble-tests":   lib/ramble/ramble/schema/applications.py
Step #2 - "ramble-tests":   lib/ramble/ramble/test/cmd/workspace.py
Step #2 - "ramble-tests":   lib/ramble/ramble/workspace/workspace.py
Step #2 - "ramble-tests": =======================================================
Step #2 - "ramble-tests": lib/ramble/ramble/test/cmd/workspace.py:1242: [E302] expected 2 blank lines, found 1
```

douglasjacobsen (approved):

lib/ramble/ramble/workspace/workspace.py
rfbgo: OK great, that's what I had but it wasn't clear to me that it would resolve to the same `CommentedDict`, changed!

PR #36, opened 2023-02-08
Allow workspace_path to be set via config:user:workspace_dir

dodecatheon:
This is the simplest possible change -- just change the workspace_path definition to config:user:workspace_dir, if found, otherwise use the old default.


douglasjacobsen:
@dodecatheon Thanks for the contribution!

I think we should define a default config value here: https://github.com/GoogleCloudPlatform/ramble/blob/develop/etc/ramble/defaults/config.yaml

And then remove the one from `ramble.paths`. You'd also need to add the definition of the config into the schema (which should be in the PR regardless).

I'd also prefer that the name is something that could easily be changed into a list in the future. Maybe something like:
`config:workspace_dirs`. Where the current value is a string (or maybe even a list already?).

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -62,7 +62,14 @@
 _active_workspace = None
 
 #: Path where workspaces are stored
-workspace_path = os.path.join(ramble.paths.var_path, 'workspaces')
+workspace_path = ramble.config.get('config:user:workspace_dir')
```
douglasjacobsen, line 65: Can we remove the `user` portion?

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -62,7 +62,14 @@
...
+if workspace_path:
+    try:
+        workspace_path = os.path.expanduser(workspace_path)
+    except KeyError:
+        workspace_path = os.path.join(ramble.paths.var_path, 'workspaces')
```
douglasjacobsen, line 70: I think we should remove the default from paths, and push it into the default configs instead.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -62,7 +62,14 @@
 _active_workspace = None
 
 #: Path where workspaces are stored
-workspace_path = os.path.join(ramble.paths.var_path, 'workspaces')
+workspace_path = ramble.config.get('config:user:workspace_dir')
+if workspace_path:
+    try:
```
douglasjacobsen, line 67: Given this try block, maybe this warrants a test to ensure we're trapping what we think we are.

douglasjacobsen (changes requested):

lib/ramble/ramble/workspace/workspace.py
dodecatheon: Sure

lib/ramble/ramble/workspace/workspace.py
dodecatheon: Tell me what you think of the latest change. Should I add a ramble.paths.default_workspace_dirs setting?

lib/ramble/ramble/workspace/workspace.py
dodecatheon: Outdated? I tried a different way of setting it.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -62,7 +62,10 @@
 _active_workspace = None
 
 #: Path where workspaces are stored
-workspace_path = os.path.join(ramble.paths.var_path, 'workspaces')
+workspace_path = ramble.config.get('config:workspace_dirs')
+if not workspace_path:
+    workspace_path = os.path.join(ramble.paths.var_path, 'workspaces')
```
douglasjacobsen, line 67: We should revert to building a path from `ramble.paths.var_path` if `config:workspace_dirs` is not present. Instead, we should only rely on it being defined in `config:workspace_dirs`. And then some commands might need to be updated to properly handle the case where the directory might not exist.

douglasjacobsen (changes requested):
That's closer to what I was thinking. You still need to update the schema for config to define this option though.

And I think we probably need at least one test, to make sure changing the workspace dir works in the future.

Also, something in this PR made the dryrun_setup test timeout. So, you might need to look at that.

To run it locally, you can `ramble unit-test -k test_dryrun_setup`

lib/ramble/ramble/workspace/workspace.py
dodecatheon: Did you mean "We should *not* revert to building a path ..."?

IIUC, you mean that since ramble.config.get('config:workspace_dirs') is now getting the default location unless redefined by the user, we don't have to assemble the default location here also.

Currently, there is no way to handle the case where there is no workspaces under var_path, say if someone has deleted the directory. So do you want to add new functionality that will handle that also, or at least report an error?

douglasjacobsen: Yeah, sorry. We should *not* revert to building a path.

Yes, basically, we should make the only path for getting the workspace_dir through a config, and not have a hardcoded path in `ramble.paths` in addition to a config set one.

I don't know that we need to throw an error. If the directory `config:workspace_dirs` points to doesn't exist, there wouldn't be any workspaces in it. The only place that an error would occur is if ramble tried to create this directory, and didn't have permissions. But, I think that's handled when trying to create a workspace the first time anyway.

It's worth testing at least.

lib/ramble/ramble/test/conftest.py
```diff
@@ -384,11 +384,11 @@ def _factory(name, output, subdir=('bin',)):
 @pytest.fixture()
 def mutable_mock_workspace_path(tmpdir_factory):
     """Fixture for mocking the internal ramble workspaces directory."""
-    saved_path = ramble.workspace.workspace_path
+    # saved_path = ramble.workspace.get_workspace_path()
```
douglasjacobsen, line 387: Remove?

lib/ramble/ramble/test/conftest.py
```diff
@@ -384,11 +384,11 @@ def _factory(name, output, subdir=('bin',)):
...
     mock_path = tmpdir_factory.mktemp('mock-workspace-path')
-    ramble.workspace.workspace.workspace_path = str(mock_path)
+    ramble.workspace.set_workspace_path(str(mock_path))
     yield mock_path
-    ramble.workspace.workspace.workspace_path = saved_path
+    # ramble.workspace.set_workspace_path(saved_path)
```
douglasjacobsen, line 391: Remove?

lib/ramble/ramble/test/cmd/workspace.py
```diff
@@ -327,6 +385,8 @@ def test_workspace_flag_named():
     with ramble.workspace.read(ws_name) as ws:
         check_no_basic(ws)
 
+    # workspace('remove', '-y', ws_name)
```
douglasjacobsen, line 388: Why is this commented out?

douglasjacobsen (changes requested):

share/ramble/qa/run-unit-tests
```diff
@@ -42,4 +42,4 @@ bin/ramble help -a
 #-----------------------------------------------------------
 # Run unit tests with code coverage
 #-----------------------------------------------------------
-$coverage_run $(which ramble) unit-test -x --verbose
+$coverage_run $(which ramble) unit-test -x # -s --verbose -vv --full-trace
```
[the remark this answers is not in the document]
dodecatheon: Sorry, I'll take that out

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -281,9 +279,30 @@ def active_workspace():
     return _active_workspace
 
 
+def get_workspace_path():
```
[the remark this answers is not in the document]
dodecatheon: You're correct. will change.

douglasjacobsen (approved):

PR #37, opened 2023-02-09
Fix bash completion for workspaces

rfbgo:
Add missing `_workspaces` function

### DIFF
--- a/share/ramble/ramble-completion.bash
+++ b/share/ramble/ramble-completion.bash
@@ -154,6 +154,15 @@ _repos() {
     RAMBLE_COMPREPLY="$RAMBLE_REPOS"
 }
 
+_workspaces() {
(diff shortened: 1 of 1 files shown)


dodecatheon (commented):
At the top of the file, it says:
```
# NOTE: ramble-completion.bash is auto-generated by:
#
#   $ ramble commands --aliases --format=bash
#       --header=bash/ramble-completion.in --update=ramble-completion.bash
#
# Please do not manually modify this file.
```
I find this a little confusing -- are you meant to incorporate the _workspaces definition into a .py file somewhere? It seems like putting it in this file would ensure that it would be overwritten by another process.

rfbgo:
Ah, you're right I did not see that! Let me fix and update

PR #38, opened 2023-02-10
Update mirror logic, and mirror software

douglasjacobsen:
This merge updates the mirror logic to be owned by the workspace. It also adds mirroring of the defined software stack in addition to the inputs.

### DIFF
--- a/lib/ramble/ramble/application_types/spack.py
+++ b/lib/ramble/ramble/application_types/spack.py
@@ -44,6 +44,7 @@ class SpackApplication(ApplicationBase):
     def __init__(self, file_path):
         super().__init__(file_path)
         self._setup_phases = [
+            'create_spack_env',
             'install_software',
             'get_inputs',
             'make_experiments'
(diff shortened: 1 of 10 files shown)


lib/ramble/ramble/application_types/spack.py
```diff
@@ -84,39 +92,43 @@ def _add_expand_vars(self, expander):
         except ramble.spack_runner.RunnerError as e:
             tty.die(e)
 
-    def _install_software(self, workspace, expander):
-        import os
+    def _extract_specs(self, workspace, expander, spec_name, app_name):
+        """Build a list of all specs the named spec requires
```
rfbgo, line 96:
```suggestion
        """Build a list of all specs which the named spec requires
```

lib/ramble/ramble/application_types/spack.py
```diff
@@ -84,39 +92,43 @@ def _add_expand_vars(self, expander):
...
+                spec_list.extend(
+                    self._extract_specs(workspace,
+                                        expander,
+                                        dep, app_name))
+        spec['application_name'] = app_name
+        spec_list.append((spec_name, spec))
+        return spec_list
+
+    def _create_spack_env(self, workspace, expander):
+        """Write create the spack environment for this experiment
```
rfbgo, line 114:
```suggestion
        """Create the spack environment for this experiment
```

lib/ramble/ramble/application_types/spack.py
```diff
@@ -84,39 +92,43 @@ def _add_expand_vars(self, expander):
...
+
+        Extract all specs this experiment uses, and write the spack environment
+        file that for it.
```
rfbgo, line 117:
```suggestion
        file for it.
```

lib/ramble/ramble/spack_runner.py
```diff
@@ -345,6 +351,28 @@ def get_package_path(self, package_spec):
             self._dry_run_print(args)
             return '/dry-run/path/to/%s' % package_spec.split()[0]
```
rfbgo, line 352: Is it worth changing this to `os.path` for portability? (even though it's not a real path...)

lib/ramble/ramble/spack_runner.py
```diff
@@ -345,6 +351,28 @@ def get_package_path(self, package_spec):
...
+            "-d",
+            mirror_path
+        ]
+
+        if not self.dry_run:
+            return self.exe(*args, output=str).strip()
+        else:
+            self._dry_run_print(args)
+            return """
+0    already present
```
rfbgo, line 372: If we're going not DRY / repeat this print perhaps we should use the same logic in the places it repeats so we're less likely to miss it in a future cleanup? 

It is a bit of a pain to DRY though..
```
                "  %-4d already present"  % len(self._software_mirror_stats.present),
                "  %-4d added"            % len(self._software_mirror_stats.new),
                "  %-4d failed to fetch." % len(self._software_mirror_stats.errors))
```

rfbgo (changes requested):

lib/ramble/ramble/spack_runner.py
douglasjacobsen: I updated the logic, but the spack runner can't reference `self._software_mirror_stats`. That's only a member of the workspace. Here we're just passing the string that spack would have printed (if we called it directly) to give the stats of the software mirror creation process.

Ideally, this would call spack methods directly, instead of calling spack as a sub-process. But for now we still have spack as an external dep.

rfbgo: Makes sense, thanks!

rfbgo (approved):
LGTM!

PR #42, opened 2023-02-14
Fix bug where `-t` was not working during workspace create

### LINKED ISSUE #41: Passing `-t` to workspace create fails
### Steps to reproduce

```
ramble workspace create -t <input_tpl> <workspace_name>
```

This calls a now removed function

### Error message

```
==> Error: 'Workspace' object has no attribute '_read_execution_template'
```

-

### General information

rfbgo:
Also add a test to cover this in the future

Fixes #41

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -243,8 +243,8 @@ def _workspace_create(name_or_path, dir=False,
 
     if template_execute:
         with open(template_execute, 'r') as f:
-            workspace._read_execution_template(f.read())
-            workspace._write_execution_template()
+            workspace._read_template('execute_experiment', f.read())
+            workspace._write_templates()
 
     return workspace
 
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -243,7 +243,9 @@ def _workspace_create(name_or_path, dir=False,
 
     if template_execute:
         with open(template_execute, 'r') as f:
-            workspace._read_template('execute_experiment', f.read())
+            _, file_name = os.path.split(template_execute)
+            template_name = os.path.splitext(file_name)[0]
+            workspace._read_template(template_name, f.read())
             workspace._write_templates()
 
     return workspace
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/test/cmd/workspace.py
```diff
@@ -182,6 +182,34 @@ def test_workspace_dir(tmpdir):
...
+    with tmpdir.as_cwd():
+        tpl_in = """
+cd "{experiment_run_dir}"
+
+cmake -DTEST=1 -h
+{spack_setup}
+
+{command}
+        """
+        print(tmpdir)
```
douglasjacobsen, line 195: Nit; I think we don't need this anymore.

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -243,8 +243,8 @@ def _workspace_create(name_or_path, dir=False,
 
     if template_execute:
         with open(template_execute, 'r') as f:
-            workspace._read_execution_template(f.read())
-            workspace._write_execution_template()
+            workspace._read_template('execute_experiment', f.read())
```
douglasjacobsen, line 246: Does it make sense to get the name of the template from the file, instead of creating it as `execute_experiment`?

It would be kind of sad if someone used:
`ramble workspace create -t /tmp/my_awesome_template.tpl -c ramble.yaml`

and referenced `{my_aweseom_template}` in `ramble.yaml`, but this broke that. I'm not sure I feel like this is an important thing to do right now. So, if it's difficult, it might not be worth the effort.

douglasjacobsen (changes requested):
I think this looks good. I left two comments.

If possible, it might be nice to diff the created template with the provided one to make sure they are the same, but I don't see why they wouldn't be (alternatively, if the names match that might be good enough?)

lib/ramble/ramble/cmd/workspace.py
rfbgo: Yea that's a good idea, I'll add it

rfbgo:
Updated to address review comments and slightly improve tests (could be more robust, for imo this is OK for now). PTAL

douglasjacobsen (approved):

PR #48, opened 2023-02-23
Fix bug introduced by PR #36

dodecatheon:
PR #36, configurable workspace_dirs, introduced a bug with the following test sequence:
ramble config blame config | grep workspace_dirs

ramble workspace create test1
ramble workspace list
ramble config blame config | grep workspace_dirs
ramble workspace activate test1
ramble workspace list
ramble config blame config | grep workspace_dirs
ramble workspace create test2
ramble workspace list
ramble config blame config | grep workspace_dirs

The file ~/.ramble/<ostype>/config.yaml was getting written with an incorrect setting for workspace_dirs.


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -137,8 +134,7 @@ def workspace_activate(args):
         env_mods = ramble.workspace.shell.deactivate()
 
     # Activate new workspace
-    workspace_path_dir = ramble.workspace.get_workspace_path()
-    active_workspace = ramble.workspace.Workspace(workspace_path_dir)
+    active_workspace = ramble.workspace.Workspace(workspace_path)
```
douglasjacobsen, line 137: I think there should be an override context right above this.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -281,17 +281,17 @@ def active_workspace():
 
 def get_workspace_path():
     """Returns current directory of ramble-managed workspaces"""
-    path_in_config = ramble.config.get('config:workspace_dirs')
+    path_in_config = ramble.config.get('config:workspace_dirs', scope='user')
```
douglasjacobsen, line 284: I don't love hard-coding the scope here. I think we should get whatever the correct definition is instead.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -281,17 +281,17 @@ def active_workspace():
...
     if not path_in_config:
-        path_in_config = '$ramble/var/ramble/GET_WORKSPACE_ERROR/'
+        path_in_config = ramble.config.get('config:workspace_dirs', scope='_builtin')
```
douglasjacobsen, line 286: If we get the correct definition is above, and path is not set in the config, I think this should be considered an error. In that case, we should call `tty.die` instead to kill the program.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -281,17 +281,17 @@ def active_workspace():
...
 
     wspath = ramble.util.path.canonicalize_path(path_in_config)
     return wspath
 
 
 def set_workspace_path(dirname):
     """Sets the parent directory of ramble-managed workspaces"""
-    ramble.config.set('config:workspace_dirs:', dirname)
+    ramble.config.set('config:workspace_dirs', dirname, scope='user')
```
douglasjacobsen, line 294: Here I also don't love setting the scope explicitly. Maybe we don't need `set_workspace_path` anymore, and we should just use contextual config overrides where ever we have this.

douglasjacobsen (changes requested):

lib/ramble/ramble/cmd/workspace.py
dodecatheon: lib/ramble/ramble/cmd/workspace.py was reverted to the exact form it had before PR#36. So if this change needs to be made, maybe it should be made as part of a separate PR?

lib/ramble/ramble/workspace/workspace.py
dodecatheon: changed as requested

douglasjacobsen (approved):

PR #53, opened 2023-02-27
Allow ramble workspace edit to open more files

rfbgo:
It now opens the config and all avaialble templates by default

As a side affect I removed the ability to specify which template to open (and create), which is techincally a UI breaking change


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -336,6 +336,18 @@ def template_path(ws_path, requested_template_name):
     return template_path
 
 
+def all_templates(path):
```
douglasjacobsen, line 339: Two nits;

1. This method has the same name as the `all_templates` method in the workspace class, but the functionality is different. Maybe this could be `all_template_paths` ?

2. Does it make more or less sense for this method to be inside the workspace class instead of global to the workspace module? The functionality is basically the same, I just can't decide if it should be module or class level...

douglasjacobsen (approved):
LGTM. Two nits, but feel free to ignore if you have a strong preference.

lib/ramble/ramble/workspace/workspace.py
rfbgo: For bullet 1, I agree, I will change it

For bullet 2, I did actually consider that during dev but there was interface issue why I decided that was not as good idea as it initially seemed. I think it was that when doing the `ramble workspace edit` stuff tries to avoid operations on the workspace as we wanted to be less dependent on having a valid workspace (so you can edit an invalid workspace)  -- as such I thought a static method made more sense

douglasjacobsen (approved):

PR #67, opened 2023-03-06
Update OpenFOAM to support explicit decomp

rfbgo:
OpenFOAMv10 reportedly added the need for manual decomposition, as reported by a user. This change adds variables to support the user in specifying a decomp

I confirmed manually that this fixes the reported issue, and checked that it should be backwards compatible with v7 but have been having a hard time testing this change on my local system


rfbgo:
I just managed to try this against OpenFoam v7 to check it's backwards compatible, and it worked fine. I think we're good to go

douglasjacobsen (approved):

PR #69, opened 2023-03-07
Add Experimental Upload of experiment results

rfbgo:
This PR adds the `ramble workspace analyze --upload` command to push results into a database (targeting bigQuery first). I've added most of this logic in the experimental namespace, as it's likely to change

It includes a few other changes:

- Adds general uploader logic / command
- Restructure the main results dict to be more list centric and database-like 
- Adds config:upload block 

There are some design decisions around how we want to integrate this into main line (should it live along side the code? should it be a submodule?) so opening this kick off those discussions

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -31,6 +31,8 @@ import ramble.fetch_strategy
 import ramble.util.install_cache
 import ramble.success_criteria
 from ramble.mirror import MirrorStats
+from ramble.config import ConfigError
+import ramble.experimental.uploader
 
 import spack.util.spack_yaml as syaml
 import spack.util.spack_json as sjson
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1107,11 +1109,38 @@ def concretize(self):
...
+    def write_json_results(self):
+        out_file = os.path.join(self.root, 'results.json')
+        with open(out_file, 'w+') as f:
+            sjson.dump(self.results, f)
+        return out_file
+
+    def upload_results(self):
+        if ramble.config.get('config:upload'):
+            # Read upload type and push it there
+            if ramble.config.get('config:upload:type') == 'BigQuery':  # TODO: enum?
```
douglasjacobsen, line 1121: Can you add this config into the schema?

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1107,11 +1109,38 @@ def concretize(self):
...
+                formatted_data = ramble.experimental.uploader.format_data(self.results)
+
+                # TODO: strategy object?
+                uploader = ramble.experimental.uploader.BigQueryUploader()
+
+                uri = ramble.config.get('config:upload:uri')
```
douglasjacobsen, line 1127: Can you add this config into the schema?

douglasjacobsen (changes requested):
I like this a lot.

I think we should make tasks for unifying the experiment class with the new experiment_set class.

In the future, I think we should add tests for this, but since it's in experimental it's probably better to omit them for now.

The only changes I'd like is to add the config options to the schema. Maybe those should also be scoped as experimental though?

douglasjacobsen (approved):

PR #71, opened 2023-03-15
Convert spack_setup to builtins

douglasjacobsen:
This merge migrates the previously defined `spack_setup` variable to use builtins that are registered in the spack application class.

These are automatically added to all spack application instances.

Additionally, a `builtin::spack_deactivate` builtin is added, to help with more complicate spack experiments that might need to deactivate and reactivate spack environments.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -356,6 +356,9 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
                 del self.variables['executable_name']
 
         self.variables['command'] = '\n'.join(command)
+
+        # TODO (dwj): Remove this after we validate that 'spack_setup' is not in templates.
+        #             this is no longer needed, as spack was converted to builtins.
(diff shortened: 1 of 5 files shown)


dodecatheon (changes requested):
Fails flake8 test

dodecatheon (changes requested):
Overall, looks good to me. One question: do the current tests cover new lines of code? If so, you can mark as addressed once flake8 error is fixed.

douglasjacobsen:
@dodecatheon With the exception of the `spack_deactivate` builtin yes.

coverage report says:
`lib/ramble/ramble/application_types/spack.py                                      171     62    64%   23, 27, 68-81, 110, 160-164, 179, 187, 222-223, 227-281, 298`

dodecatheon (approved):

rfbgo:
Pretty sure this breaks the location in the rendered execution where the `spack_setup` is emitted, which may be troublesome. Will follow up offline

PR #75, opened 2023-03-16
Revive unused code path in workspace for includes

rfbgo:
Previously this code path was not used anywhere (either in known prod or testing) and thus got broken during updates

This PR both fixes it and adds a test to cover it (woo increased coverage!)

### DIFF
--- a/lib/ramble/ramble/schema/merged.py
+++ b/lib/ramble/ramble/schema/merged.py
@@ -20,7 +20,7 @@ import ramble.schema.workspace
 
 #: Properties for inclusion in other schemas
 properties = union_dicts(
-    ramble.schema.applications.properties,
+    ramble.schema.applications.applications_schema,
     ramble.schema.config.properties,
     ramble.schema.repos.properties,
     ramble.schema.workspace.properties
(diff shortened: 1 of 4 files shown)


douglasjacobsen:
@rfbgo Open question that we discussed offline, but is useful to document here:

What happens if you have something like this:

`include.yaml`
```
ramble:
  variables:
    my_var: 'from_include'
```

`ramble.yaml`
```
ramble:
  variables:
    my_var: 'from_workspace'
  include:
  - include.yaml
```

Which value does `{my_var}` give in the resulting workspace?

rfbgo:
Right now this isn't super useful (to the point where we might disable). Right now this does get read and propagates down to scopes, but nothing on the consuming side checks to see if any "include" scopes exists to consume them

As such this doesn't really have an affect... (my main motivation with this PR was to avoid calling now dead code, but I'm open to throwing an error etc)

douglasjacobsen:
I think it's good to keep it in. It's just useful to know how it can be used currently (and it's good that it now has a test).

Maybe in the future we can properly consume these included scopes.

douglasjacobsen (approved):

PR #77, opened 2023-03-17
Add a class to manage ramble reserved keywords

douglasjacobsen:
This merge introduces a new class to encapsulate all of the ramble known keywords, and their types.

Tests are added to make sure incorrectly defined keywords throw reasonable errors.

Additionally, past references to the `*_key` type keyword definitions are updated to use this class now.


lib/ramble/ramble/keywords.py
```diff
@@ -0,0 +1,119 @@
...
+        'processes_per_node': key_types.required,
+        'n_threads': key_types.required,
+        'batch_submit': key_types.required,
+        'mpi_command': key_types.required,
+    }
+
+    def __init__(self):
+        # Define class attributes for all of the keys
+        for key in self.keys.keys():
+            setattr(self, key, key)
```
rfbgo, line 63: Can you remind me what the motivation was to do it this way instead of base classing enum? (ie `class Keywords(Enum):`. I'm fine either way, I was just surprised to see it this way round, but I like that we avoid the `.name`

lib/ramble/ramble/keywords.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Current key types are:
+    - Reserved: Ramble defines these, and a user should not be allowed to define them
+    - Optional: Ramble can function with a definition from the user but it isn't required
+    - Required: Ramble requires a definition for these. Ramble will try to set sensible defaults,
+                but it might not be possible always.
+
+    """
+
+    key_types = Enum('types', ['reserved', 'optional', 'required'])
```
rfbgo, line 33: tiny nit: imo this would be better as a singular

lib/ramble/ramble/keywords.py
```diff
@@ -0,0 +1,119 @@
...
+
+from enum import Enum
+
+import llnl.util.tty as tty
+
+import ramble.error
+
+
+class Keywords(object):
+    """Class to represent ramble known keywords.
```
rfbgo, line 17:
```suggestion
    """Class to represent known ramble keywords.
```?

File: lib/ramble/ramble/keywords.py [hunk 4]
```diff
@@ -0,0 +1,119 @@
...
+        'batch_submit': key_types.required,
+        'mpi_command': key_types.required,
+    }
+
+    def __init__(self):
+        # Define class attributes for all of the keys
+        for key in self.keys.keys():
+            setattr(self, key, key)
+
+    def is_valid(self, key):
```
rfbgo, line 65: If we use a more traditional enum then can this class be static? Right now we use a lot of `self` but don't really hold state. That would let us do something more like:

```
Keywords.is_reservered(...)
```
Rather than the current
```
keywords = Keywords()
keywords.is_reservered(...)
```

rfbgo (changes requested):
Really nice PR. Few in the weeds questions/suggestions but I really like it overall. Validator here we come!

rfbgo:
@douglasjacobsen can you fix up the conflict if you chose to implement some of my suggested changes too, please?

lib/ramble/ramble/expander.py
```diff
@@ -63,21 +66,24 @@ def __init__(self, variables, experiment_set):
     @property
     def application_name(self):
         if not self._application_name:
-            self._application_name = self.expand_var('{application_name}')
+            var = '{' + self._keywords.application_name + '}'
```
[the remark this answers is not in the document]
douglasjacobsen: Yes, I had thought about this, but I didn't have a great name for the expansion variable syntax.

Like... is it better to do:

`self._keywords.application_name_expander` ? or `self.for_expander(self._keywords.application_name)` ?

lib/ramble/ramble/keywords.py
douglasjacobsen: That was really just it. If I used enum, the default is to give the ID, and I didn't love having to append `.name` to all of the key names just to get the name.

The enum didn't really buy me anything beyond having a standard class for them, but I needed to wrap the logic anyway so I decided it would just confuse the implementation more.

lib/ramble/ramble/expander.py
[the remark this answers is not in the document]
rfbgo: I prefer wrapping it in a functional call

lib/ramble/ramble/keywords.py
douglasjacobsen: Also, it doesn't seem like there is a nice way to union enums, and with a single enum it's kind of gross to manage the types associated with each key.

lib/ramble/ramble/keywords.py
douglasjacobsen: Agreed.

I can make most of these static methods.

I don't actually think I can use traditional enums after looking at this more, at least not without making the code more complex.

One option to help with this `Keywords.is_reserved(...)` aspect is to create singleton instance of the class (as is done in config), and change the references to that instead of instantiating the class multiple times.

douglasjacobsen: FWIW, I made these class methods. But then added the singleton of the class so we don't need to refer to it by instantiating it anymore.

douglasjacobsen:
OK, Conflict is fixed now.

rfbgo (approved):

lib/ramble/ramble/experiment_set.py
```diff
@@ -82,16 +61,25 @@ def __init__(self, workspace):
...
         # Set some base variables from the workspace definition.
-        self.set_base_var(self.mpi_key, workspace.mpi_command)
-        self.set_base_var(self.log_dir_key, workspace.log_dir)
-        self.set_base_var(self.batch_submit_key, workspace.batch_submit)
-        self.set_base_var(self.spec_name_key, '{application_name}')
+        self.set_base_var(self.keywords.mpi_command, workspace.mpi_command)
+        self.set_base_var(self.keywords.log_dir, workspace.log_dir)
+        self.set_base_var(self.keywords.batch_submit, workspace.batch_submit)
+        self.set_base_var(self.keywords.spec_name,
+                          ramble.expander.Expander.expansion_str(self.keywords.application_name))
```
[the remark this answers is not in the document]
douglasjacobsen: Mostly because I find the short-hand hard for a newcomer to understand.

i.e. a fully qualified bit tells them where it's coming from without having to cross reference the imports.

However, I was not consistent in this PR. Even in this file I have a mix of the two, because in one place the fully qualified makes the lines too long (and pretty gross looking) and flake complains about it.

I'm happy to use one or the other though, if that would help readability.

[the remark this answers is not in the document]
rfbgo: I think #1 prio is to be consistent, and #2 is to follow some "nice" style. Let's change this to make them consistent in whatever style you prefer / think is more consistent with the rest of the codebase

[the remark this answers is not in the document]
douglasjacobsen: OK, I pushed an update. Should be ready once the CI finishes.

rfbgo (approved):
LGTM, thanks for the improvements

PR #78, opened 2023-03-21
Add test hostname workload with no inputs

rfbgo:
Inputs are broken, making non of the currently workloads easily usable


douglasjacobsen:
Is it better to add a new test, or just remove the fake input from the existing workloads?

Technically, they are broken, so I might prefer having the input removed from the other ones instead (or in addition to, but if it's done, this test workload isn't super useful anymore).

rfbgo:
I'm fine with removing the workloads or the input url, anything that makes this usable

douglasjacobsen:
OK, then I think we should remove the input file from all of the workloads, and then only have the serial / parallel ones.

douglasjacobsen (approved):

PR #79, opened 2023-03-21
Allow site variables to be read by experiment

rfbgo:
This is working, but a few design decisions to confirm before we put it into mainline:

1. The current interface / expected format in the site config is:
```
config:
  variables:
    site_test: 1
  env-vars:
    set:
      site_env: 2
```
(ie 
```    
    config('--scope', 'site', 'add', f'config:variables:{var_name1}:{var_val}')
    config('--scope', 'site', 'add', f'config:env-vars:set:{var_name2}:{var_val}')
```
)

Are we ok with that? 

2.  I moved all of the `_namespace` vars in lib/ramble/ramble/workspace/workspace.py into a class, it's somewhat irrelevant to this PR but I included it since I needed to access them outside of workspace. Are we OK with that change and should we move that out to a more common area? 

3. I left a FIXME to dry out all the magic string use around `'site'`. I think that's something we want to do but perhaps out of scope for this PR? I'm happy to do that as a follow on (but it's not urgent by any means)


douglasjacobsen:
1) I think that's fine. The only thing I can think of as a suggestion is to make `variables` (or something) a separate config section. But, given we're also doing env-vars, it's unclear if that helps, or hurts what this is trying to do.

2) Yes, I think that's a good change. I think we should consider where they should live. It looks to me like the `_namespace` vars were only ever used in the workspace module, so I think where you have them is fine for now. If we start to re-use them outside of the workspace context, we should revisit this.

3) Yeah, I don't think it's urgent. These are technically in lib/ramble/ramble/config.py, but not in a "let's not use magic strings" way.

lib/ramble/ramble/test/end_to_end.py
```diff
@@ -534,3 +535,87 @@ def test_env_var_builtin(mutable_config, mutable_mock_workspace_path, mock_appli
                 if export_found and cmd3_regex.search(line):
                     cmd_found = True
             assert cmd_found and export_found
+
+
+def test_sitevar_dry_run(mutable_config, mutable_mock_workspace_path):
```
douglasjacobsen, line 540: Does it make sense to test multiple scopes here?

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -47,19 +47,22 @@
...
-experiment_namespace = 'experiments'
-variables_namespace = 'variables'
-success_namespace = 'success_criteria'
-env_var_namespace = 'env-vars'
-mpi_namespace = 'mpi'
-batch_namespace = 'batch'
-compiler_namespace = 'compilers'
-mpi_lib_namespace = 'mpi_libraries'
+
+class namespace:
```
douglasjacobsen, line 51: Thanks for doing this!

lib/ramble/ramble/experiment_set.py
```diff
@@ -29,7 +30,7 @@ class ExperimentSet(object):
     """
 
     # In order of lowest to highest precedence
-    _contexts = Enum('contexts', ['base', 'application',
+    _contexts = Enum('contexts', ['site', 'base', 'application',
```
douglasjacobsen, line 33: It's nice to see how easy adding a layer of variables is. :)

lib/ramble/ramble/experiment_set.py
```diff
@@ -93,6 +96,29 @@ def __init__(self, workspace):
...
+        site_vars = self.get_site_vars(workspace)
+        site_env_vars = self.get_site_env_vars(workspace)
+        site_name = 'site'  # FIXME: DRY out these magic strings
+        self._set_context(self._contexts.site,
+                          site_name,
+                          site_vars,
+                          site_env_vars)
+
+    def get_site_vars(self, workspace):
+        site_conf = ramble.config.config.get_config('config', scope='site')
```
douglasjacobsen, line 109: What happens here if you leave off `scope='site'` and just get all of the config defined variables instead?

If that works, I think it would be preferable (i.e. this could be just config file defined variables, instead of specifically variables defined in the site config scope).

Then, the config parser will handle merging them before they are passed to this code.

lib/ramble/ramble/experiment_set.py
```diff
@@ -93,6 +96,29 @@ def __init__(self, workspace):
...
+        if site_conf and ramble.workspace.namespace.variables in site_conf:
+            site_vars = site_conf[ramble.workspace.namespace.variables]
+            return site_vars
+        return None
+
+    def get_site_env_vars(self, workspace):
+        site_conf = ramble.config.config.get_config('config', scope='site')
```
douglasjacobsen, line 116: Same thing here about `scope='site'`

douglasjacobsen (changes requested):

lib/ramble/ramble/test/end_to_end.py
rfbgo: I thought about this and I decided not. I like the idea in general but it wasn't clear to me that we actually wanted to enable variables to come in from every possible scope or just site? I was taking the scope of this to be site

douglasjacobsen: I think limiting it to 'site' makes for a kind of awkward user experience. i.e. this works:
`ramble config --scope=site add 'config:variables:test_var:"test"'`

but this doesn't:
`ramble config add 'config:variables:test_var:"test"'`

neither does:
`ramble config --scope=defaults add 'config:variables:test_var:"test"'`

And I'm pretty sure if we just remove the scope specification on these lines it'll "just work(TM)".

In lieu of that, I think we need some clear user facing documentation that this is only expected to work within the site scope. If there is a technical reason for this to only be at the site scope though, I think adding the documentation is sufficient. However, I think it would be helpful to know that as well.

douglasjacobsen: Awesome, thanks!

douglasjacobsen (approved):

PR #81, opened 2023-03-21
Add yaml support for custom executables and executable order control

douglasjacobsen:
This merge adds the ability to define custom executables in the YAML workspace configuration file.

Additionally, this adds the ability to control the executables used for creating the '{command}' variable within experiments.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -30,6 +30,8 @@ import ramble.mirror
 import ramble.fetch_strategy
 import ramble.expander
 
+from ramble.workspace import namespace
+
 from ramble.language.application_language import ApplicationMeta, register_builtin
 from ramble.error import RambleError
 
@@ -283,8 +285,6 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
         - command: set to the commands needed to execute the experiment
         - spack_setup: set to an empty string, so spack applications can override this
         """
-        from ramble.workspace import namespace
-
         executables = self.workloads[self.expander.workload_name]['executables']
         inputs = self.workloads[self.expander.workload_name]['inputs']
 
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -70,6 +70,7 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
         self.expander = None
         self.variables = None
         self.experiment_set = None
+        self.internals = None
         self._env_variable_sets = None
 
         self._file_path = file_path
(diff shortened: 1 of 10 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -782,8 +785,8 @@ def all_specs(self):
         import ramble.spec
 
         specs = []
-        for app, workloads, _, _ in self.all_applications():
-            for workload, _, _, _ in self.all_workloads(workloads):
+        for app, workloads, _, _, _ in self.all_applications():
```
rfbgo, line 788: Should this be `*_` too?

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1074,7 +1094,7 @@ def concretize(self):
         mpi_dict = spack_dict[namespace.mpi_lib]
         applications_dict = spack_dict[namespace.application]
 
-        for app_name, _, _, _ in self.all_applications():
+        for app_name, _, _, _, _ in self.all_applications():
```
rfbgo, line 1097: Same *_ here?

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1688,6 +1709,14 @@ def get_workspace_env_vars(self):
             if namespace.env_var in \
             workspace_dict[namespace.ramble] else None
 
+    def get_workspace_internals(self):
```
rfbgo, line 1712: these `get_workspace_xxx` blocks are very similar, do you think we should abstract the out ?

rfbgo (approved):
LGTM. Great to see how this was actually simple at it's core (but required adding lots of small changes)

lib/ramble/ramble/application.py
```diff
@@ -276,9 +283,24 @@ def add_expand_vars(self, workspace):
         - command: set to the commands needed to execute the experiment
         - spack_setup: set to an empty string, so spack applications can override this
         """
+        from ramble.workspace import namespace
```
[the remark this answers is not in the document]
douglasjacobsen: I hoisted it. Thanks

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: Nice catch, thanks!

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: Done

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: Done!

rfbgo (approved):
Looks good! Thanks for the changes

PR #89, opened 2023-04-07
Add install_compilers as a phase to spack applications

douglasjacobsen:
This merge allows spack applications to control their own compiler installation. This changes the behavior so that only used spack compilers are installed, rather than all of the compilers that are listed.

A test is added to ensure unused compilers are not installed.


dodecatheon (approved):
LGTM, can approve once tests pass

dodecatheon (approved):
LGTM

PR #90, opened 2023-04-07
Add support for external spack environments

douglasjacobsen:
This merge adds functionality to allow Ramble to refer to external spack environments instead of generating its own.

Documentation and tests are added for this functionality.

### DIFF
--- a/lib/ramble/docs/workspace_config.rst
+++ b/lib/ramble/docs/workspace_config.rst
@@ -477,6 +477,47 @@ is as follows:
 Not all of the options are required, but generally a spec object should contain
 at least ``base``, and ``version``.
 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+External Spack Environment Support:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**NOTE**: Using external Spack environments is an advanced feature.
+
+Some experiments will want to use an externally defined Spack environment
+instead of having Ramble generate its own Spack environment file. This can be
+useful when the Spack environment a user wants to experiment with is
+complicated.
+
+This section shows how this feature can be used.
+
+.. code-block:: yaml
+
+    spack:
+      applications:
+        gromacs:
+          external_spack_env: name_or_path_to_spack_env
+          gromacs:
+            base: gromacs
+
+In the above example, the ``external_spack_env`` keyword refers an external
+Spack environment. This can be the name of a named Spack environment, or the
+path to a directory which contains a Spack environment. Ramble will copy the
+``spack.yaml`` file from this environment, instead of generating its own.
+
+This allows users to describe custom Spack environments and allow them to be
+used with Ramble generated experiments. However, users will still need to
+define some aspects of the ``spack:applications:<app_name>`` dictionary in the
+``ramble.yaml`` file. Specifically, any software specs that are required by the
+application.py file will need to have their ``base`` defined at a minimum.
+Ramble won't use any other information in the spec definition (i.e. ``version``
+or ``variants``), but adding the information to the ``ramble.yaml`` will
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/spack_runner.py
```diff
@@ -326,14 +327,51 @@ def add_include_file(self, include_file):
...
+
+        named_location_args = [
+            'location',
+            '-e',
+            env_name_or_path
+        ]
+
+        # Try to use a named environment first
+        try:
+            path = self.exe(*named_location_args, output=str).strip('\n')
```
rfbgo, line 352: Should this be a `try` or more of a "check if we think it exists"? I don't think it matters too much either way, but I typically find it more graceful to check existence than generate an error

lib/ramble/ramble/spack_runner.py
```diff
@@ -326,14 +327,51 @@ def add_include_file(self, include_file):
...
+        lock_file = os.path.join(path, 'spack.lock')
+        if os.path.exists(lock_file):
+            found_lock = True
+            shutil.copyfile(lock_file, os.path.join(self.env_path, 'spack.lock'))
+
+        conf_file = os.path.join(path, 'spack.yaml')
+        if not os.path.exists(conf_file):
+            raise InvalidExternalEnvironment(f'{path} is not a spack environment.')
+
+        shutil.copyfile(conf_file, os.path.join(self.env_path, 'spack.yaml'))
```
rfbgo, line 368: Are you worried about clobbering existing confs?

rfbgo (changes requested):
Looks good. Few clarifying questions

lib/ramble/ramble/spack_runner.py
douglasjacobsen: No, I purposefully want to clobber existing configs. Currently, ramble's semantics are that when you run `workspace setup` it regenerates the `spack.yaml` (even if it's identical). Also, it defaults to forcibly re-concretizing (clobbering a `spack.lock`).

So, I think this makes the external environments behave the same way internal ones behave.

douglasjacobsen: No, I purposefully want to clobber existing configs. Currently, ramble's semantics are that when you run `workspace setup` it regenerates the `spack.yaml` (even if it's identical). Also, it defaults to forcibly re-concretizing (clobbering a `spack.lock`).

So, I think this makes the external environments behave the same way internal ones behave.

lib/ramble/ramble/spack_runner.py
rfbgo: I guess one option would be to flip it and check if the thing they gave us is indeed a file? What are the changes something is both a named env and a valid file?

douglasjacobsen: That's a good suggestion. I just updated this, let me know if you like that more.

lib/ramble/docs/workspace_config.rst
```diff
@@ -477,6 +477,42 @@ is as follows:
...
+``spack.yaml`` file from this environment, instead of generating its own.
+
+This allows users to describe custom Spack environments and allow them to be
+used with Ramble generated experiments. However, users will still need to
+define some aspects of the ``spack:applications:<app_name>`` dictionary in the
+``ramble.yaml`` file. Specifically, any software specs that are required by the
+application.py file will need to have their ``base`` defined at a minimum.
+Ramble won't use any other information in the spec definition (i.e. ``version``
+or ``variants``), but adding the information to the ``ramble.yaml`` will
+enhance the provenance information that Ramble is able to track.
```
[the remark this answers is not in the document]
douglasjacobsen: K, this is updated. Let me know if the new text helps.

rfbgo (approved):
Thanks for the changes

PR #91, opened 2023-04-07
Add dry run tests for all applications

douglasjacobsen:
This merge adds two new tests that collectively dry-run every application/workload in Ramble. This includes applications in both the builtin and builtin.mock repositories.

HPL and LULESH both had issues in their application.py files, so this merge fixes those.

### DIFF
--- a/lib/ramble/ramble/test/end_to_end.py
+++ b/lib/ramble/ramble/test/end_to_end.py
@@ -8,6 +8,7 @@
 
 import os
 import glob
+import re
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/test/end_to_end.py
```diff
@@ -800,3 +801,125 @@ def test_custom_executables(mutable_config, mutable_mock_workspace_path, mock_ap
...
+
+            for line in app_info.split('\n'):
+                match = workload_regex.search(line)
+                if match:
+                    workloads.append(match.group('wl_name').replace(' ', ''))
+
+            ws_name = f'test_all_workloads_{app}'
+            with ramble.workspace.create(ws_name) as ws:
+                ws.write()
+                config_path = os.path.join(ws.config_dir, ramble.workspace.config_file_name)
```
rfbgo, line 902: I don't how much of a duplication this is compared to the other test. Is there a way we can make the differences/distinctions more concise? It feels like we should be able to somehow get fancy with the args

share/ramble/qa/run-unit-tests
```diff
@@ -42,4 +42,4 @@ bin/ramble help -a
 #-----------------------------------------------------------
 # Run unit tests with code coverage
 #-----------------------------------------------------------
-$coverage_run $(which ramble) unit-test -x --verbose
+$coverage_run $(which ramble) unit-test -x --verbose -m "not long"
```
rfbgo, line 45: Do you have strong thoughts about which way round we want the unit test behavior by default? Should it do `long` or `not long` by default for a developer testing on their local machine?

rfbgo (approved):
LGTM. Few questions but none of them hard blockers

lib/ramble/ramble/test/end_to_end.py
douglasjacobsen: I parameterized it, so it isn't duplicated. Let me know if you like that more.

rfbgo:
Very nice! Can you resolve the conflict?

douglasjacobsen:
Should be resolved now. Thanks!

PR #99, opened 2023-04-12
Ensure RAMBLE_VARIABLES contains only expanded variables

douglasjacobsen:
This merge updates RAMBLE_VARIABLES (in results .json and .yaml files) to only contain expanded forms of the variables (i.e. not `{processes_per_node}*{n_nodes}`).

RAMBLE_RAW_VARIABLES is added to only contain the un-expanded forms.

Additionally, experiment_template_name is added to tracked variables, to help connect rendered experiments with the series that generated them.


rfbgo (approved):
Thanks for the fix!

PR #104, opened 2023-04-17
Allow repo.yaml to specify a subdir for applications

douglasjacobsen:
This merge allows the repo.yaml in an application repository to specify they subdirectory the applications should live in.

The default is `applications`. Empty strings can be passed in to allow no subdirectory.

This replicates the functionality from a PR in spack https://github.com/spack/spack/pull/36643

### DIFF
--- a/lib/ramble/ramble/repository.py
+++ b/lib/ramble/ramble/repository.py
@@ -27,6 +27,7 @@ except ImportError:
 
 
 import six
+from enum import Enum
 
 import ruamel.yaml as yaml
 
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/test/repository.py
```diff
@@ -12,18 +12,20 @@
 import ramble.paths
 
 
-@pytest.fixture()
-def extra_repo(tmpdir_factory):
+@pytest.fixture(params=["applications", "", "foo"])
```
rfbgo, line 15: Is it easy to add an end_to_end style test too? I don't love that this adds a user exposed API/CLI surface (`-d`) and doesn't explicitly use it

lib/ramble/ramble/repository.py
```diff
@@ -1052,7 +1055,7 @@ def __contains__(self, app_name):
         return self.exists(app_name)
 
 
-def create_repo(root, namespace=None):
+def create_repo(root, namespace=None, subdir=applications_dir_name):
```
rfbgo, line 1058: Should we make `applications_dir_name` more obviously `const` (ie enum member or similar) if we're making such that a user changing the dir does not update the var?

rfbgo (changes requested):
LGTM, 2 questions

lib/ramble/ramble/test/repository.py
douglasjacobsen: This should be fixed. I added some tests in `cmd/repo.py` to actually use this feature.

rfbgo (approved):
Thanks for the changes, this is really nice now

PR #106, opened 2023-04-19
Change workspace name to basename for all workspaces

douglasjacobsen:
This merge changes the name property of workspaces to only return the bsaename of the path.

Previously, this was the full path for anonymous workspaces.


rfbgo (approved):
Nice :)

PR #108, opened 2023-04-21
Add minixyce application definition

dodecatheon:


var/ramble/repos/builtin/applications/minixyce/application.py
```diff
@@ -0,0 +1,145 @@
...
+    def _make_experiments(self, workspace):
+        super()._make_experiments(workspace)
+
+        input_path = self.expander.expand_var('{experiment_run_dir}/params.txt')
+
+        settings = ['t_start', 't_step', 't_stop', 'tol', 'k']
+
+        with open(input_path, 'w+') as f:
+            for setting in settings:
+                f.write(setting + ' = ' + self.expander.expand_var('{' + setting + '}') + '\n')
```
douglasjacobsen, line 126: Please change the '{' + ... + '}' syntax tp use this" https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/expander.py#L190

var/ramble/repos/builtin/applications/minixyce/application.py
```diff
@@ -0,0 +1,145 @@
...
+                      description='Number of Ladder stages for RC|RLC ladder|ladder2 test generation',
+                      workloads=rc_workloads)
+
+    log_file = '{experiment_run_dir}/{workload_name}_tran_results.prn'
+    processed_output = '{experiment_run_dir}/processed_output.txt'
+
+    result_regex = r'.*\s+(?P<num_iters>[0-9]+)\s+(?P<num_restarts>[0-9]+)'
+
+    floating_point_regex = r'[\+\-]*[0-9]*\.*[0-9]+E*[\+\-]*[0-9]*'
+    close_paren = r')'
```
douglasjacobsen, line 79: I'm not sure if `close_paren` adds much. It's usage isn't consistent here, and the actually string is shorter than the variable name.

I'm not opposed to the idea of using the variable instead of the string, but I don't think we should mix and match the semantics in this file.

douglasjacobsen (changes requested):
Thanks! This looks good overall. Two inline comments, and one other question:

What was the discussion around adding `success_criteria` to this?

var/ramble/repos/builtin/applications/minixyce/application.py
dodecatheon: success_criteria:

Sample tests cir1 through cir4 can be run and then checked using a provided perl script. This simply matches exact numeric results.

In general, changing platforms should have the same final FoM values on any platform.  Should we provide comparisons to be checked for certain workloads?

douglasjacobsen: I'm curious what this would look like. Is it just running a perl script, taking the expected output, and hard-coding it into the python file?

If so, that seems like a reasonable thing to try. Though we might want to test that the values don't change on a handful of platforms before we commit to something like that.

dodecatheon: Added success_criteria that verifies the job ran.

douglasjacobsen (approved):

PR #109, opened 2023-04-21
Remove mpi and batch dictionaries

douglasjacobsen:
This merge removes the two primary dictionaries in the ramble.yaml file `ramble:mpi` and `ramble:batch` in favor of using the variables `mpi_command` and `batch_submit` respectively.

Tests and documenatation is updated to reflect this too.

### DIFF
--- a/lib/ramble/ramble/schema/workspace.py
+++ b/lib/ramble/ramble/schema/workspace.py
@@ -65,39 +65,6 @@ properties = {
                 'type': 'object',
                 'additionalProperties': {'type': 'string'},
             },
-            'mpi': {
-                'type': 'object',
-                'default': {
-                    'command': {
-                        'type': 'string',
-                        'default': 'mpirun'
-                    },
-                    'args': {
-                        'type': 'array',
-                        'default': []
-                    }
-                },
-                'properties': {
-                    'command': {
-                        'type': 'string',
-                        'default': 'mpirun'
-                    },
-                    'args': {
-                        'type': 'array',
-                        'default': []
-                    }
-                },
-            },
-            'batch': {
-                'type': 'object',
-                'default': {
-                    'submit': ''
-                },
-                'properties': {
-                    'submit': {'type': 'string'}
-                },
-                'additionalProperties': False
-            },
             'variables': ramble.schema.applications.variables_def,
             'internals': ramble.schema.applications.internals_def,
             'success_criteria': ramble.schema.applications.success_list_def,
--- a/examples/README.md
+++ b/examples/README.md
@@ -106,5 +106,6 @@ The slurm execute experiment script shows an example of creating an
 `execute_experiment` script for use in a slurm cluster.
 
 To use this, place it in the configs directory for your workspace, and change
-the `ramble:batch:submit` definition in your ramble.yaml to: `sbatch {slurm_execute_experiment}`.
-Then when performing `ramble workspace setup`, the `all_experiments` script will submit to slurm.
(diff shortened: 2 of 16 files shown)


douglasjacobsen:
@rfbgo  and @dodecatheon I'd like you to both review this, since it's a pretty major breaking change relative to the previous way this worked. I also think after this gets merged, we should make a new tag.

dodecatheon (approved):
LGTM

lib/ramble/ramble/schema/workspace.py
```diff
@@ -65,39 +65,6 @@
                 'type': 'object',
                 'additionalProperties': {'type': 'string'},
             },
-            'mpi': {
```
rfbgo, line 68: I think it would be nice if we can temporarily keep some code in that catches when a user is using the "old style" and warn them 

If I understood the PR correctly there is a chance where they end up with the default values for the variables and get a bit confused about what is going on

rfbgo (changes requested):
LGTM, but I wonder if we should do a "software transition" and help the users out a bit more?

lib/ramble/ramble/schema/workspace.py
douglasjacobsen: OK, I've added a deprecation check now after our offline discussion. Let me know if this seems to help.

rfbgo: Looks go to me. What do you see as the path to deprecating things in the future, and does it include `@deprecated`? If so it might make sense to use it here too?

douglasjacobsen: My current plan has been to mark things as deprecated in the code, and use `tty.warn` to throw a warning to the user.

I'm then going to make tasks to review these in a few releases, so we can remove the deprecated functionality before 1.0 is released.

Where possible, we can use `@deprecated`, but some of this type of stuff won't be nice with that (since we're not deprecating a function, we're really deprecating part of a function).

But maybe we should push these things into functions more clearly so we can use the deprecated decorator. What are your thoughts?

rfbgo: My only "requirement" for this PR is that we help the user know when they are using functionality that is going away, and give them a smooth transition, I think you've already achieved this :) 

I like everything you propose, and see this as a good chance to kick the tires on whatever we think our post v1.0 deprecation process is (I don't really mind what that looks like, but I too like the idea of a function level decorator that calls `tty.warn` so pulling the deprecated portion into functions sounds very nice and tidy to me)

Up to you if you want to capture that here or later

douglasjacobsen: Spending some time looking into how deprecate works, I don't think it's going to end up being any better than what I've implemented currently.

So, I think the following are kind of my current thinking as far as how we should go about handling deprecated thing:
1) Make an obvious 'warning' print in the code to let people know their functionality is deprecated. I don't really love the way this is handled by python currently, because it's printed when the functionality is used, rather than warning at the end of execution. But maybe we can explore ways to fix that outside of this PR. Ultimately this is done with `tty.warn` calls.
2) Mark the code with `# DEPRECATED` and list something that describes when it can be removed
3) Make a task tracking when we will remove the functionality.

rfbgo: Yea, I agree. Lets merge this and possibly follow up with another task when needed?

douglasjacobsen: SGTM

rfbgo (approved):

PR #110, opened 2023-04-28
Add pre-commit setup for pre-push hooks

douglasjacobsen:
This merge adds a pre-commit configuration file to manage git hooks.

Currently, there are two hooks
 - The flake8 tests are added as a pre-commit hook
 - The unit-tests (short only) are added as a pre-push hook.

It also updates the requirements.txt and README based on that.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+repos:
+- repo: local
+  hooks:
+  - id: flake8
+    name: flake8
+    entry: share/ramble/qa/run-flake8-tests
+    language: system
+  - id: short-unit-tests
+    name: short-unit-tests
+    entry: share/ramble/qa/run-unit-tests
+    language: system
+    stages: [push]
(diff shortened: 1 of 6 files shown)


rfbgo (changes requested):
I'm happy with the changes, but one sharp edge.. this is annoying locally if you don't have a branch called `develop`, since it compares against it (also true of the main flake8)

I sometimes find myself working in a fork with no develop (or even delete local develop). I don't have a great suggestion, and only really see two options:

1. Catch that and run a full flake8 test
2. Document how to bypass the checks one time if you have the hooks install

Thoughts?

douglasjacobsen:
OK, I've added several fallback options to the file list logic in `flake8`. This should go something like:

1. develop
2. origin/develop (to catch a repo with only a missing local develop, but still an upstream)
3. origin/main (Same, but for upstreams that only have main)
4. HEAD but list all files

rfbgo (approved):
Thanks for the updates, I really like this!

PR #112, opened 2023-05-01
Allow users to specify output_capture for executables

rfbgo:
This includes both applications via application.py, and custom exectuables defined in yaml

### DIFF
--- a/var/ramble/repos/builtin/applications/hostname/application.py
+++ b/var/ramble/repos/builtin/applications/hostname/application.py
@@ -18,9 +18,9 @@ class Hostname(ExecutableApplication):
 
     input_file('test', url="https://no.domain/file/test_dir.tgz", description="Example input file...")
 
-    executable('local', 'time hostname', use_mpi=False, output_capture="&>")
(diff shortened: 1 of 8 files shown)


var/ramble/repos/builtin/applications/hostname/application.py
```diff
@@ -18,9 +18,11 @@ class Hostname(ExecutableApplication):
 
     input_file('test', url="https://no.domain/file/test_dir.tgz", description="Example input file...")
 
-    executable('serial', '/usr/bin/time hostname', use_mpi=False)
-    executable('parallel', '/usr/bin/time hostname', use_mpi=True)
+    executable('local', 'time hostname', use_mpi=False, output_capture="&>")
```
douglasjacobsen, line 21: I'm not sure where we ended on this discussion, but can this use `OUTPUT.ALL` semantics instead of hard-coding `&>` ?

rfbgo: updated

douglasjacobsen (approved):

PR #119, opened 2023-05-10
Compiler Specs were previously added to spack.yaml

rfbgo:
This fixes a bug where compilers were added to the generated spack env, and was causing a whole host of issues including breaking platform detection and having odd mixing of gcc libs

Also removed no longer called add_compiler function


douglasjacobsen:
Internally the style checks passed. So I'm going to merge.

PR #120, opened 2023-05-11
Add an application definition for Quantum Espresso

douglasjacobsen:
This merge adds an application definition for the Quantum Espresso application. It includes several benchmark workloads that can be used to evaluate its performance.

### DIFF
--- a/var/ramble/repos/builtin/applications/quantum-espresso/application.py
+++ b/var/ramble/repos/builtin/applications/quantum-espresso/application.py
@@ -1,0 +1,99 @@
+# Copyright 2022-2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import os
+from ramble.appkit import *
+from ramble.expander import Expander
+
+
+class QuantumEspresso(SpackApplication):
+    '''Define Quantum-Espresso application.
+    https://www.quantum-espresso.org/
+    '''
+    name = 'quantum-espresso'
+
+    tags = ['electronic-structure', 'materials', 'dft', 'density-functional-theory', 'plane-waves', 'pseudopotentials']
+
+    default_compiler('gcc12', spack_spec='gcc@12.2.0', compiler_spec='gcc@12.2.0')
+
+    software_spec('impi2021p8', spack_spec='intel-oneapi-mpi@2021.8.0', compiler='gcc12')
+
+    software_spec('quantum-espresso', spack_spec='quantum-espresso@7.1', compiler='gcc12')
+
+    required_package('quantum-espresso')
+
+    input_file('AUSURF112', url='https://github.com/QEF/benchmarks/releases/download/bench0.0/AUSURF112.tgz',
+               sha256='2b71c27801f8abbde05f48297b56b3194be0f83be50c50433f44bb886cac5117',
+               description='Input file for AUSURF112 benchmark')
+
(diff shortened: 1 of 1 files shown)


dodecatheon:
LGTM

dodecatheon (approved):
LGTM

var/ramble/repos/builtin/applications/quantum-espresso/application.py
```diff
@@ -0,0 +1,99 @@
...
+# except according to those terms.
+
+import os
+from ramble.appkit import *
+from ramble.expander import Expander
+
+
+class QuantumEspresso(SpackApplication):
+    '''Define Quantum-Espresso application.
+    https://www.quantum-espresso.org/
```
rfbgo, line 16: nit: Is this needed? Since it relies on the spack package which supports `url` is seems a bit surplus?

var/ramble/repos/builtin/applications/quantum-espresso/application.py
```diff
@@ -0,0 +1,99 @@
...
+    input_file('GRIR686', url='https://github.com/QEF/benchmarks/releases/download/bench0.0/GRIR686.tgz',
+               sha256='4ade7691f88dd322b41abecbc08cc55437f9d376142e62b615317598abe498d7',
+               description='Input file for GRIR686 benchmark')
+
+    input_file('PSIWAT', url='https://github.com/QEF/benchmarks/releases/download/bench0.0/PSIWAT.tgz',
+               sha256='698fbdac9f307b9ebc77e70e52574d67adb0f21bfd6874559d633471664b77d6',
+               description='Input file for PSIWAT benchmark')
+
+    # Set of small benchmarks to build a workload from.
+    # input_file('pwscf-small', url='https://github.com/QEF/benchmarks/releases/download/bench0.0/pwscf-small-benchmark.tar.gz',
```
rfbgo, line 51: If you want to keep this, shall we just uncomment it? (or remove it entirely?)

var/ramble/repos/builtin/applications/quantum-espresso/application.py
```diff
@@ -0,0 +1,99 @@
...
+    for wl_name, input_file in zip(workload_names, input_files):
+        workload(wl_name, executables=['copy_inputs', 'execute'], inputs=[wl_name])
+        workload_variable('input_path', default=Expander.expansion_str(wl_name),
+                          description='Path to inputs for ' + wl_name + ' workload',
+                          workloads=[wl_name])
+        workload_variable('input_file', default=input_file,
+                          description='Name of input file for ' + wl_name + ' workload',
+                          workloads=[wl_name])
+
+    log_str = Expander.expansion_str('log_file')
```
rfbgo, line 71: Is this something we should make consistent across the other packages?

rfbgo (approved):
Wasn't planning on leaving a review, but a few things caught my eye as I was passing :)

rfbgo (approved):
Wasn't planning on leaving a review, but a few things caught my eye as I was passing :)

var/ramble/repos/builtin/applications/quantum-espresso/application.py
douglasjacobsen: I can remove it as I'm not planning on adding support for it yet. But I had wanted to leave it in here because it's a set of smaller benchmarks that one can run. The problem with it is you need to generate input files (they have a bash script in the repo that I had considered porting to python, but it seemed unnecessary).

var/ramble/repos/builtin/applications/quantum-espresso/application.py
douglasjacobsen: I made a task to discuss this.

I think the only action for this PR is if we want this syntax or not. But I think if we want this, this PR would not make it consistent across all other files. (I think NAMD might also use this for part of it though)

rfbgo (approved):

PR #124, opened 2023-05-12
Fix success regex for Quantum Espresso

douglasjacobsen:
The success regex previously didn't properly identify success.


rfbgo (approved):
Good catch!

PR #127, opened 2023-05-16
split end_to_end.py into separate modules in end_to_end subdirectory

dodecatheon:


dodecatheon:
Not sure this is quite ready, on my laptop it fails unit-test in test_configvar_dry_run

dodecatheon:
Fixed errors, passes test, ready for review

lib/ramble/ramble/test/end_to_end/env_var_builtin.py
```diff
@@ -0,0 +1,214 @@
...
+            for line in f.readlines():
+                if not export_found and export_regex.search(line):
+                    assert not cmd_found
+                    export_found = True
+                if export_found and cmd3_regex.search(line):
+                    cmd_found = True
+            assert cmd_found and export_found
+
+
+def test_configvar_dry_run(mutable_config, mutable_mock_workspace_path):
```
douglasjacobsen, line 130: This test looks replicated below in test_configvar_dry_run. Do we still need it here too?

lib/ramble/ramble/test/end_to_end/hpl_dry_run.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+# everything here uses the mock_workspace_path
+pytestmark = pytest.mark.usefixtures('mutable_config',
+                                     'mutable_mock_workspace_path')
+
+workspace = RambleCommand('workspace')
+
+
+def test_hpl_dry_run(mutable_config, mutable_mock_workspace_path):
```
douglasjacobsen, line 26: We can probably remove this test, as it should be covered in the known_applications test.

douglasjacobsen (changes requested):
Thanks. Two comments, but otherwise this LGTM.

lib/ramble/ramble/test/end_to_end/env_var_builtin.py
dodecatheon: Oops, that was an earlier version that crept back in when I did `rebase -i`. Fixing, see next push.

lib/ramble/ramble/test/end_to_end/hpl_dry_run.py
dodecatheon: Removing.

douglasjacobsen (approved):

PR #133, opened 2023-05-25
Fix application info printing software definitions

douglasjacobsen:
This merge fixes an issue where `ramble info <app>` would not print the software configuration for applications.

A test is added to help ensure this is caught in the future.


dodecatheon (approved):
LGTM

PR #139, opened 2023-05-26
Reduce unit test warnings

douglasjacobsen:
Unit test warnings appear to be coming from the use of fnmatch.filter when testing chained experiments.

These warnings do not present themselves when running the test by hand, so this merge simply ignores them on this test.


douglasjacobsen:
FWIW, I tested multiple ways of removing the warnings.

We can escape the strings in the test_config, but this causes the test to not work (because we write an invalid ramble.yaml).

I also tested using `glob.escape` on the pattern in fnmatch, but that doesn't actually prevent the warnings (and also breaks the actual glob matching).

dodecatheon (approved):
LGTM

PR #146, opened 2023-05-26
Remove support for the deprecated mpi and batch attributes

douglasjacobsen:
This merge removes support for the deprecated mpi and batch attributes, in favor of the mpi_command and batch_submit variable definitions.


dodecatheon (approved):
LGTM

PR #149, opened 2023-05-31
Add attribute front-end commands

douglasjacobsen:
This merge adds support for `ramble attributes` to help manage tags and maintainers on applications and modifiers.


lib/ramble/ramble/language/modifier_language.py
```diff
@@ -391,3 +391,35 @@ def _env_var_modification(mod):
             mod.env_var_modifications[mode][method].append(append_dict.copy())
 
     return _env_var_modification
+
+
+@modifier_directive(dicts=())
+def maintainers(*names: str):
```
rfbgo, line 397: Similar comment to recent PR, I don't love how we duplicate so much stuff in the different "languages" but we can deal with that elsewhere

share/ramble/ramble-completion.bash
```diff
@@ -267,7 +267,16 @@ _ramble() {
     then
         RAMBLE_COMPREPLY="-h --help -H --all-help --color -c --config -C --config-scope -d --debug --timestamp --pdb -w --workspace -D --workspace-dir -W --no-workspace --use-workspace-repo -k --insecure -l --enable-locks -L --disable-locks -m --mock -p --profile --sorted-profile --lines -v --verbose --stacktrace -V --version --print-shell-vars"
     else
-        RAMBLE_COMPREPLY="clean commands config debug edit flake8 help info license list mirror mods on repo software-definitions unit-test workspace"
+        RAMBLE_COMPREPLY="attributes clean commands config debug edit flake8 help info license list mirror mods on repo software-definitions unit-test workspace"
```
rfbgo, line 270: I might be misremembering how this works, but don't we also need to update the `ramble-completion.in`?

share/ramble/ramble-completion.bash
```diff
@@ -267,7 +267,16 @@ _ramble() {
...
+    fi
+}
+
+_ramble_attributes() {
+    if $list_options
+    then
+        RAMBLE_COMPREPLY="-h --help --defined --undefined -a --all --by-attribute --applications --modifiers --maintainers --tags"
+    else
+        RAMBLE=""
```
rfbgo, line 279: I see that this is consistent with the other methods.. but why is the else `RAMBLE=` and not a `COMREPLY`?

var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
```diff
@@ -23,7 +23,9 @@ class IntelMpiBenchmarks(SpackApplication):
 
     name = 'IntelMpiBenchmarks'
 
-    tags = ['micro-benchmark', 'benchmark', 'mpi']
+    maintainers('rfbgo')
```
rfbgo, line 26: What if I don't want to be the maintainer? ;)

rfbgo (approved):
Few comments, but I think all can be deferred?

share/ramble/ramble-completion.bash
douglasjacobsen: I think we only need to update `ramble-completion.in` if we're expecting the ability to tab complete the actual tags or maintainers.

I think there's also a task to make sure this works properly.

var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
douglasjacobsen: Donate it to someone else? :)

lib/ramble/ramble/language/modifier_language.py
douglasjacobsen: Agreed (and I made a task for this already)

share/ramble/ramble-completion.bash
douglasjacobsen: I'm not sure. Maybe that's a bug in `lib/ramble/ramble/cmd/commands.py` ?

rfbgo: I'll make an (internal) task to look at that and we can merge this for now

PR #151, opened 2023-06-01
Remove support for deprecated env-vars

douglasjacobsen:
This merge removes support for the deprecated env-vars config sections.

Previously, environment variables were defined through the env-vars config section, but we transitioned to env_vars to allow the config section to be modifiable through `ramble config` commands. This merge completes this transition by removing support and warnings for the older format.


rfbgo (dismissed):
LGTM. I'm going to let you chose the right time to merge this @douglasjacobsen since it might break existing users workflows

rfbgo (dismissed):
Looks good. 

I'm a bit nervous about breaking any users on v1, but better to do this pre v1.0 release

rfbgo (dismissed):

rfbgo (approved):

rfbgo (dismissed):
Last attempt to un-confuse GH before I force the merge

PR #154, opened 2023-06-02
Fix workloads with multiple inputs

douglasjacobsen:
This merge fixes an issue where workloads with multiple inputs (archives or files) would not configure any inputs past the first properly.

Notable changes (primarily internal to Ramble and application definitions) are that inputs are now stored at:
$workspace/inputs/application/workload/input-name

The lammps application definition used the older format, and has been updated.

### DIFF
--- a/lib/ramble/ramble/test/mirror_tests.py
+++ b/lib/ramble/ramble/test/mirror_tests.py
@@ -64,25 +64,41 @@ def test_mirror_cache_symlinks(tmpdir):
 def create_archive(archive_dir, app_class):
     tar = spack.util.executable.which('tar', required=True)
 
-    for input_name, conf in app_class._inputs_and_fetchers().items():
-        archive_dir.ensure(input_name, dir=True)
-        archive_name = os.path.basename(conf['fetcher'].url)
-        test_file_path = str(archive_dir.join(input_name, 'input-file'))
-        with open(test_file_path, 'w+') as f:
-            f.write('Input File\n')
+    app_class._inputs_and_fetchers()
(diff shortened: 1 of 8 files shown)


lib/ramble/ramble/test/mirror_tests.py
```diff
@@ -64,25 +64,43 @@ def test_mirror_cache_symlinks(tmpdir):
...
-        archive_dir.ensure(input_name, dir=True)
-        archive_name = os.path.basename(conf['fetcher'].url)
-        test_file_path = str(archive_dir.join(input_name, 'input-file'))
-        with open(test_file_path, 'w+') as f:
-            f.write('Input File\n')
-
-        with archive_dir.as_cwd():
-            tar('-czf', archive_name, input_name)
-            with open(archive_name, 'rb') as f:
+    if not hasattr(app_class, '_input_fetchers'):
```
rfbgo, line 67: I don't have any great suggestions.. but this is kind of gross

I'd prefer that it was hidden under some common interface if we really need to support both.

I noticed we do this in a bunch of places, so perhaps it is worthwhile?

rfbgo (changes requested):
Nice enhancement! One (hopefully easy) request then it LGTM

lib/ramble/ramble/test/mirror_tests.py
douglasjacobsen: OK, this should look better now. Let me know if you have any other issues with it.

rfbgo (approved):

PR #160, opened 2023-06-06
add spell checking

vsoch:
Hi! I was trying this out today, and I noticed a few typos while browsing the source code. So this pull request will fix all found
typos (for code developed here) and add spell checking to the CI to ensure we catch any for future additions/changes!

There were a handful that were typos in variables or attributes, namely in testing an argument to `config(` and a typo of "global" and then an attribute that was something like "softwre_*" and I fixed them, and  hopefully they are correct in the fixed state.

### DIFF
--- a/.github/workflows/spelling.yaml
+++ b/.github/workflows/spelling.yaml
@@ -1,0 +1,18 @@
+name: ramble spelling
+
+on:
+  pull_request: []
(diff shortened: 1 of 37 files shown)


douglasjacobsen:
Thanks for taking the time to submit this!

PR #163, opened 2023-06-09
[WIP] fix spacing in gromacs config

vsoch:
I'm attempting to build this with spack (and run in flux) and I think to start the spacing of this section of the file might be a bit off? If I am looking at it, I think the named section should be at the same level as `pme_single_rank` and both be under experiments.


douglasjacobsen:
These should be fixed in https://github.com/GoogleCloudPlatform/ramble/pull/199

PR #170, opened 2023-06-13
Allow input files path to be expanded

rfbgo:
Previously the url for an application `input_file` could not be a variable. Something like:

```
input_file('CONUS_12km', url='{my_input}', ...)
workload_variable('my_input', default='https://url/to/resource'
```

Would not work. This PR enables expansion of the url

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -776,7 +776,7 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
             workload = self.workloads[workload_name]
 
             for input_file in workload['inputs']:
-                input_conf = self.inputs[input_file]
+                input_conf = self.inputs[input_file].copy()
 
                 # Expand input value as it may be a var
                 expanded_url = self.expander.expand_var(input_conf['url'])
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/application.py
```diff
@@ -778,6 +778,10 @@ def _inputs_and_fetchers(self, workload=None):
             for input_file in workload['inputs']:
                 input_conf = self.inputs[input_file]
 
+                # Expand input value as it may be a var
+                expanded_url = self.expander.expand_var(input_conf['url'])
```
douglasjacobsen, line 782: I'm not sure if this will actually matter, but would it be safer to create a copy of `input_conf` and overwrite that?

Just to make sure the change is constrained to this specific instance?

douglasjacobsen (changes requested):
LGTM over all. One small comment.

lib/ramble/ramble/application.py
rfbgo: Sure. Safety first! However, if that's the case then maybe it's better to do something more like:

```
                input_conf['expanded_url'] = expanded_url
                fetcher = ramble.fetch_strategy.URLFetchStrategy(**input_conf)
                file_name = os.path.basename(input_conf['expanded_url'])
```
(ie just had the file name consume the expanded one?)

douglasjacobsen: I think you'll need `input_conf['url']` to be set to the expanded one too, so that the fetcher is configured with the right URL.

rfbgo: Sure, makes sense. I can add a `.copy()`

douglasjacobsen (approved):

PR #174, opened 2023-06-15
Add support for application based success functions and FOM comparisons

douglasjacobsen:
This merge adds the ability for applications to define an `evaluate_success` function, which can perform application specific validation of an experiment on top of the built in success criteria.

This function is automatically added to all experiments if it is defined.

Additionally, this merge adds the ability for success criteria to have `mode='fom_comparison'` where a comparison formula can be defined to validate the value of a FOM.


lib/ramble/ramble/test/modifier_functionality/mock_env_var_modifiers.py
```diff
@@ -10,6 +10,7 @@
 
 import pytest
 
+from ramble.test.dry_run_helpers import *
```
rfbgo, line 13: It's not a big deal but why do you `from .. import *` in here not just `import module` ?

douglasjacobsen: Laziness mostly. 😏 I should at a minimum make it `from ... import <stuff>` instead of `*`.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1101,7 +1099,7 @@ def dump_results(self, output_formats=['text']):
                                 exp['name'])
                         f.write('  Status = %s\n' %
                                 exp['RAMBLE_STATUS'])
-                        if exp['RAMBLE_STATUS'] == 'SUCCESS':
+                        if exp['RAMBLE_STATUS'] == 'SUCCESS' or self.always_print_foms:
```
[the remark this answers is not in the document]
douglasjacobsen: Probably. Let me change that.

lib/ramble/ramble/test/modifier_functionality/mock_env_var_modifiers.py
rfbgo: Yea that's what caught my eye, I think that's generally considered "better" (or just import the entire module)

lib/ramble/ramble/workspace/workspace.py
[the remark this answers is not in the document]
douglasjacobsen: I believe this should be fixed now, but let me know if I missed some logic in the uploader.

rfbgo (approved):

PR #182, opened 2023-06-21
Add workspace / experiment hash functionality

douglasjacobsen:
This merge adds the ability for experiment instances to hash themselves in a reproducible way. Experiment instances also now create inventories of the various aspects of their definitions. Workspaces aggregate this information into a workspace inventory and hash.

Together these changes allow users to determine which aspects of a set of experiments in a workspace change between setup executions.

It is important to note however, that the hash of a spack environment is only determined after concretizing the spack environment. Outside of the spack differences, this hash is fully determined with an execution of `ramble workspace setup --dry-run`.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -35,6 +35,7 @@ import ramble.repository
 import ramble.modifier
 import ramble.util.executable
 import ramble.util.colors as rucolor
+import ramble.util.hashing
 
 from ramble.keywords import keywords
 from ramble.workspace import namespace
@@ -50,6 +51,7 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
     _mod_prefix_builtin = 'modifier_builtin::'
     _builtin_required_key = 'required'
     _workload_exec_key = 'executables'
+    _inventory_file_name = 'ramble_inventory.json'
 
     #: Lists of strings which contains GitHub usernames of attributes.
     #: Do not include @ here in order not to unnecessarily ping the users.
@@ -80,6 +82,14 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
         self._modifier_instances = []
         self._modifier_builtins = {}
         self._input_fetchers = None
+
+        self.hash_inventory = {
+            'attributes': [],
+            'inputs': [],
+            'software': [],
+            'templates': [],
+        }
+        self.experiment_hash = None
 
         self._file_path = file_path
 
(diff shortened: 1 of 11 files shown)


lib/ramble/ramble/application.py
```diff
@@ -865,17 +875,97 @@ def _make_experiments(self, workspace):
...
+        before hashing, to help give useful hashes.
+        """
+
+        # Purge workspace name, as this shouldn't affect the experiments
+        if 'workspace_name' in variables:
+            del variables['workspace_name']
+
+        # Remove the workspace path from variable definitions before hashing
+        for var in variables:
+            if isinstance(variables[var], six.string_types):
```
rfbgo, line 902: I'm a bit uneasy about the hashing actually changing the workspace (not that it's actually an issue or its not repeatable)

Is it perhaps better to either: a) clone and edit the attr dict at hash time; or b) separate out the things like root and subpath so we only delete not edit for hashing?

lib/ramble/ramble/util/hashing.py
```diff
@@ -0,0 +1,26 @@
...
+        bytes = f.read()
+        file_hash = hashlib.sha256(bytes).hexdigest()
+    return file_hash
+
+
+def hash_string(string):
+    return hashlib.sha256(string.encode('UTF-8')).hexdigest()
+
+
+def hash_json(in_json):
```
rfbgo, line 25: Does `dump` impose a strict ordering to the data? It would be neat if this was resilient to order changes 

Also, dump means things with different spacing etc should be fine, right?

rfbgo (approved):
Two quick questions

lib/ramble/ramble/application.py
douglasjacobsen: What is actually implemented is, I think, basically your (a): https://github.com/GoogleCloudPlatform/ramble/pull/182/files/155880a645977e10222d4f29e4bfc1cfa7c26cb6#diff-f442a3694bd08c70f90d7995078d7a5b62e2443faa816e9f97ada8389dee66c5R916-R917

Also, I wanted to edit rather than delete, since users could have paths in variables that they defined, instead of only Ramble managed ones (which would basically just re-create this issue).

lib/ramble/ramble/util/hashing.py
douglasjacobsen: It orders it based on the YAML order for ramble.yaml. Since it uses a spack_json dict, it's ordered based on creation order.

But we could change it to be ordered by name or something instead.

douglasjacobsen: I added some explicit sorting to help make this more resilient to order changes.

rfbgo (approved):

PR #183, opened 2023-06-21
Warn when a package and its compiler are added to an environment

douglasjacobsen:
This merge adds a warning when a software environment definition contains both a package, and the package's compiler. This is because package managers (like spack) might try to re-install the compiler within these environments even if they are installed outside of the environment.

This is just a warning, and is not harmful, but hopefully will help users catch issues that can be easily remedied by removing the compiler from the environment.


dodecatheon (approved):
LGTM

PR #185, opened 2023-06-23
Allow variables to be specified in spack config section

douglasjacobsen:
Previously, the `variables` attribute was not added to the spack schema. This prevented variables from being defined in standard spack config sections.

This merge adds `variables` to the spack schema.


dodecatheon (approved):
LGTM

PR #186, opened 2023-06-23
Fix an issue when merging variables in software environments

douglasjacobsen:
Previously, internally defined variables would destroy all variable definitions from higher level scopes.

This merge fixes this, and adds a test to help ensure this behavior works as expected.


dodecatheon (approved):
LGTM

PR #191, opened 2023-06-28
Update modifier langauge to use the language helpers

douglasjacobsen:
This merge updates the modifier language module to use the language helpers, such as `require_definition` to remove some duplicated logic.


dodecatheon (approved):
LGTM

PR #193, opened 2023-06-28
Add success criteria to modifier definitions

douglasjacobsen:
This merge allows modifiers to define success criteria. This allows modifiers to define additional success criteria that augment an experiment's list of success criteria, and can be more closely tied to output from the modifier.


dodecatheon (approved):
LGTM

PR #194, opened 2023-06-29
Fix calculation of P and Q for calculator workload

dodecatheon:
Fixes a small error in the calculation of P and Q for the calculator workload


dodecatheon:
Abandoning, implementing more comprehensive fixes in intel-hpl PR #195

PR #195, opened 2023-06-30
Intel MKL-optimized HPL

dodecatheon:
Uses provided binary from intel-oneapi-mkl spack package

calculator workload modified to correctly calculate N based on 80% of available memory, with P and Q based on 2 procs per node.

To change memory usage, e.g. for testing, set percent_mem for workload calculator in ramble.yaml.

### DIFF
--- a/var/ramble/repos/builtin/applications/hpl/application.py
+++ b/var/ramble/repos/builtin/applications/hpl/application.py
@@ -17,119 +17,179 @@ class Hpl(SpackApplication):
 
     maintainers('douglasjacobsen', 'dodecatheon')
 
-    tags('benchmark-app', 'mini-app', 'benchmark')
+    tags('benchmark-app', 'benchmark', 'linpack')
 
     default_compiler('gcc9', spack_spec='gcc@9.3.0')
 
-    software_spec('impi2018', spack_spec='intel-mpi@2018.4.274')
+    software_spec('impi_2018', spack_spec='intel-mpi@2018.4.274')
 
-    software_spec('hpl',
-                  spack_spec='hpl@2.3 +openmp',
-                  compiler='gcc9')
+    software_spec('hpl', spack_spec='hpl@2.3 +openmp', compiler='gcc9')
 
     required_package('hpl')
 
     executable('execute', 'xhpl', use_mpi=True)
 
     workload('standard', executables=['execute'])
+    workload('calculator', executables=['execute'])
 
-    workload_variable('output_file', default='HPL.out      output file name (if any)',
+    workload_variable('output_file',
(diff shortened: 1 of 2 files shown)


dodecatheon:
Default `calculator` memory usage set to 85% of memory.

var/ramble/repos/builtin/applications/hpl/application.py
```diff
@@ -17,120 +17,180 @@ class Hpl(SpackApplication):
...
     required_package('hpl')
 
     executable('execute', 'xhpl', use_mpi=True)
 
     workload('standard', executables=['execute'])
+    workload('calculator', executables=['execute'])
 
-    workload_variable('output_file', default='HPL.out      output file name (if any)',
+    workload_variable('output_file',
+                      default='{:13} output file name (if any)'.format('HPL.out'),
```
douglasjacobsen, line 36: What is this change for? I see it's fairly pervasive, but it also seems unnecessary. Is there a reason why this is pushed through everywhere?

var/ramble/repos/builtin/applications/hpl/application.py
```diff
@@ -142,55 +202,114 @@ class Hpl(SpackApplication):
...
-            self.variables['PFACTS'] = '2'
-            self.variables['N-NBMINs'] = '1'
-            self.variables['NBMINs'] = '4'
-            self.variables['N-RFACTs'] = '1'
-            self.variables['RFACTs'] = '1'
-            self.variables['N-BCASTs'] = '1'
-            self.variables['BCASTs'] = '1'
-            self.variables['N-DEPTHs'] = '1'
-            self.variables['DEPTHs'] = '1'
+            pfact = expander.expand_var('{pfact}')
```
douglasjacobsen, line 263: Please change these lines to:
```
pfact = expander.expand_var(expander.expansion_str('pfact'))
```

douglasjacobsen (commented):
Thanks for doing this! I think my comments relate to both application files. Let me know your thoughts.

var/ramble/repos/builtin/applications/hpl/application.py
dodecatheon: Changed in next commit.

var/ramble/repos/builtin/applications/hpl/application.py
douglasjacobsen: Can we pull this out into a `pad_value` function, so the `'{:13}'` is not repeated in all of the workload variables?

dodecatheon: Changed and updated.

douglasjacobsen (approved):

PR #196, opened 2023-06-30
Extract shared directives into a shared base class

douglasjacobsen:
This merge extracts directives that are shared between multiple object types into a shared_language module.


douglasjacobsen:
The unit tests are timing out. This is fixed by https://github.com/GoogleCloudPlatform/ramble/pull/197 I'll re-run the unit-tests with a longer timeout though

PR #197, opened 2023-07-01
Convert spack version to use git and import

douglasjacobsen:
This merge changes the spack version string printing to use git and a python import instead of trying to run spack commands directly.

This fixes a regression in testing introduced with the workspace hashing merge

### DIFF
--- a/lib/ramble/ramble/main.py
+++ b/lib/ramble/ramble/main.py
@@ -116,16 +116,29 @@ def get_version():
 
     The commit sha is only added when available.
     """
-    import spack.util.git
     version = ramble.ramble_version
-    git_path = os.path.join(ramble.paths.prefix, ".git")
+    git_hash = get_git_hash(path=ramble.paths.prefix)
+
+    if git_hash:
+        version += f' ({git_hash})'
+
+    return version
+
+
+def get_git_hash(path=ramble.paths.prefix):
+    """Get get hash from a path
+
+    Outputs '<git commit sha>'.
+    """
+    import spack.util.git
+    git_path = os.path.join(path, ".git")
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/main.py
```diff
@@ -116,29 +116,42 @@ def get_version():
...
         if git.returncode != 0:
-            return version
+            return None
         match = re.match(r"[a-f\d]{7,}$", rev)
         if match:
-            version += " ({0})".format(match.group(0))
+            return match.group(0)
 
-    return version
+    return None
```
rfbgo, line 154: Is the none value meaningful here? Should we just be returning or continuing?

lib/ramble/ramble/spack_runner.py
```diff
@@ -443,6 +460,7 @@ def inventory_hash(self):
         if self.dry_run:
             spack_file = os.path.join(self.env_path, 'spack.yaml')
 
+        return 'test'
```
rfbgo, line 463: Is this at the right indent, or should it be in the dryrun? Looks like a bit of an overly aggressive early exit?

rfbgo (changes requested):

lib/ramble/ramble/spack_runner.py
douglasjacobsen: Good catch, this shouldn't be here at all. :)

lib/ramble/ramble/main.py
douglasjacobsen: I changed it to just a return.

rfbgo (approved):
Thanks for the fixup!

PR #207, opened 2023-07-07
Raise an error when variable names contain decimals

douglasjacobsen:
This merge adds an exception that explains variable names cannot contain decimals when variable names are used that contain decimals.

A test is added to ensure this exception is raised properly.


rfbgo (approved):
Thanks!

PR #212, opened 2023-07-12
Update LAMMPS application definition

douglasjacobsen:
Update figure of merit regular expressions
Add several Intel workload definitions

### DIFF
--- a/var/ramble/repos/builtin/applications/lammps/application.py
+++ b/var/ramble/repos/builtin/applications/lammps/application.py
@@ -6,6 +6,7 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
+import os
 from ramble.appkit import *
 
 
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/lammps/application.py
```diff
@@ -22,7 +23,7 @@ class Lammps(SpackApplication):
     software_spec('impi2018', spack_spec='intel-mpi@2018.4.274')
 
     software_spec('lammps',
-                  spack_spec='lammps@20220623 +opt+manybody+molecule+kspace+rigid',
+                  spack_spec='lammps@20220623.3 +opt+manybody+molecule+kspace+rigid+intel+openmp+openmp-package+asphere+dpd-basic+dpd-meso+dpd-react+dpd-smooth',
```
rfbgo, line 26: We chatted about this offline, but `+intel` caught my eye. Can you double check this is the right default list of variants?

rfbgo (changes requested):
Small change/question, otherwise LGTM

var/ramble/repos/builtin/applications/lammps/application.py
douglasjacobsen: Is it, aside from Intel. I just removed the intel one, since we only need that if we want to use the intel package (which seems like it's used for KNL and other Intel specific optimizations)

rfbgo (approved):

PR #217, opened 2023-07-14
Remove some spack env logic from vendored spack

douglasjacobsen:
This merge fixes an issue where spack tries to look up an active environment. However, since this is spack inside Ramble, there is never an active environment. This causes errors in older versions of python, so we are temporarily removing it. This should be resolved in the future by updating the vendored version of spack.


dodecatheon (approved):
LGTM

rfbgo (approved):

PR #221, opened 2023-07-20
Add support for explicit zips

douglasjacobsen:
This merge adds support for explicit zips, which are a new feature in defining experiments. Zips are named groups of variables with the same length, and can be consumed as a dimension of a matrix.

This merge adds tests, and documentation to support this feature.

### DIFF
--- a/lib/ramble/ramble/renderer.py
+++ b/lib/ramble/ramble/renderer.py
@@ -32,12 +32,16 @@ class Renderer(object):
             tty.die(f'Object type {obj_type} is not valid to render.\n' +
                     'Valid options are: "experiment"')
 
-    def render_objects(self, variables, matrices=None):
+    def render_objects(self, variables, zips=None, matrices=None):
         """Render objects based on the input variables and matrices
 
         Internally collects all matrix and vector variables.
 
-        Matrices are processed first.
+        First, zips are created. Zips extract vector variables, and group them
+        into a higher level name.
+
+        Next, matrices are processed. Matrices consume vector variables, or
+        explicit zips.
 
         Vectors in the same matrix are crossed, sibling matrices are zipped.
         All matrices are required to result in the same number of elements, but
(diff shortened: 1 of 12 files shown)


dapomeroy (commented):
LGTM

lib/ramble/ramble/renderer.py
```diff
@@ -67,8 +71,44 @@ def render_objects(self, variables, matrices=None):
...
+        defined_zips = {}
+        consumed_zips = set()
         matrix_objects = []
 
+        if zips:
+            zipped_vars = set()
+
+            for zip_group, group_def in zips.items():
+                # Create a new defined zip
+                defined_zips[zip_group] = {'vars': {}, 'length': -1}
```
rfbgo, line 83: Is `-1` actually important here, over say using 0? 

I see later you check if it's -1, and are using it as a "I set this value not someone else", but I think 0 would be the same and more truthy?

lib/ramble/ramble/renderer.py
```diff
@@ -67,8 +71,44 @@ def render_objects(self, variables, matrices=None):
...
+                cur_zip = defined_zips[zip_group]
+                for var_name in group_def:
+                    if var_name in zipped_vars:
+                        tty.die(f'Variable {var_name} is used across multiple zips.\n'
+                                'Ensure it is only used in a single zip')
+
+                    # Ensure the variable is defined
```
rfbgo, line 90: nit: it functionally doesn't matter, but it feels like "is this a real var" makes sense as the first check not down in the list? Could be worth hoisting but I don't strictly care

lib/ramble/ramble/renderer.py
```diff
@@ -139,7 +187,28 @@ def render_objects(self, variables, matrices=None):
...
+                            for zip_var in defined_zips[name]['vars']:
+                                matrix_objects[obj_idx][zip_var] = \
+                                    defined_zips[name]['vars'][zip_var][entry[name_idx]]
+
+                            # Consume the defined zip
+                            consumed_zips.add(name)
+                        else:
+                            matrix_objects[obj_idx][name] = entry[name_idx]
+
+        # Remove all consumed zips and
```
rfbgo, line 201: any reason you broken the command lines like this?

rfbgo (changes requested):
Few small questions, but LGTM!

lib/ramble/ramble/renderer.py
douglasjacobsen: Well, I wanted to support:

`var_name: []` which was why I used `-1`. But maybe I should error if a vector is passed with a length of zero, since we don't have values to extract for the variable.

lib/ramble/ramble/renderer.py
douglasjacobsen: Done

lib/ramble/ramble/renderer.py
douglasjacobsen: Fixed!

rfbgo (approved):

PR #223, opened 2023-07-24
Improve logic preventing repeated tasks

douglasjacobsen:
This merge performs two main functions, the first is to cache which tasks have previously completed by the path the task would take place in. This mainly allows subsequent tasks that would take place in the same directory, but have different namespaces to skip the task.

The second function is to add logic that prevents reconcretizing an already concretized spack environment. The new logic checks the mtime of the spack.lock, and the content of the spack.yaml (against the spack.yaml that would be generated) to determine if it's safe to skip.

Note that installation is never skipped, unless the environment was cached as installed already.


lib/ramble/ramble/spack_runner.py
```diff
@@ -425,16 +427,40 @@ def generate_env_file(self):
...
+        spack_env_file = os.path.join(self.env_path, 'spack.yaml')
+        spack_lock_file = os.path.join(self.env_path, 'spack.lock')
+
+        # Check that a spack.yaml and spack.lock file exist already
+        if os.path.exists(spack_env_file) and os.path.exists(spack_lock_file):
+            existing_env_mtime = os.path.getmtime(spack_env_file)
+            existing_lock_mtime = os.path.getmtime(spack_lock_file)
+
+            # If the lock file was last modified after the yaml file...
+            if existing_lock_mtime >= existing_env_mtime:
```
rfbgo, line 449: I was thinking about what it means for these times to be equal, and it seems unlikely that in normal use they would be equal? 

One interesting case where they might be equal is if someone copy paste the dir structure from somewhere (I think?). In that case it might be worth regenerating 

Are there other reasonable reasons these would be equal?

rfbgo (approved):
LGTM. One small question with possible tiny change

lib/ramble/ramble/spack_runner.py
douglasjacobsen: Good point. I can't decide how I feel about the copy/paste. But I think reconcretizing wouldn't hurt in that case.

PR #225, opened 2023-07-25
Add `expand_var_name` function to expander

douglasjacobsen:
This merge adds a new function to the expander `expand_var_name` which takes a variable name as a string, converts it to an expansion string (currently '{var_name}') and then expands that.


dodecatheon (approved):
LGTM

PR #227, opened 2023-07-25
Add support to import & upload experiment results

dapomeroy:
Adds 'results' module with support to import Ramble experiment results from a JSON file and upload them to BigQuery.

### DIFF
--- a/lib/ramble/ramble/cmd/results.py
+++ b/lib/ramble/ramble/cmd/results.py
@@ -7,89 +7,54 @@
 # except according to those terms.
 
 import llnl.util.tty as tty
+import json
 
-from ramble.config import ConfigError
 import ramble.experimental.uploader
 
-
 description = "take actions on experiment results"
-section = "workspaces"
+section = "results"
 level = "short"
-
-subcommands = [
-    'file_upload',
-]
-
-#: Dictionary mapping subcommand names and aliases to functions
-subcommand_functions = {}
-
-
-def results_file_upload_setup_parser(subparser):
-    subparser.add_argument(
-        'filename', metavar='filename',
-        help='path of file to upload')
-
-
-def results_file_upload(args):
-    """Imports Ramble experiment results from JSON file and uploads them."""
-    imported_results = ramble.experimental.uploader.results_file_import(args.filename)
-
-    if ramble.config.get('config:upload'):
-        # Read upload type and push it there
-        if ramble.config.get('config:upload:type') == 'BigQuery':  # TODO: enum?
-            try:
-                formatted_data = ramble.experimental.uploader.format_data(imported_results)
-            except KeyError:
-                tty.die("Error parsing file: Does not contain valid data.")
-
-            # TODO: strategy object?
-            uploader = ramble.experimental.uploader.BigQueryUploader()
-
-            # Read workspace name from results for uploader, or use default.
-            try:
-                workspace_name = formatted_data[0].workspace_name
-            except KeyError:
-                workspace_name = "Default Workspace"
-
-            uri = ramble.config.get('config:upload:uri')
-            if not uri:
-                tty.die('No upload URI (config:upload:uri) in config.')
-
-            tty.msg('Uploading Results to ' + uri)
-            uploader.perform_upload(uri, workspace_name, formatted_data)
-        else:
-            raise ConfigError("Unknown config:upload:type value")
-
-    else:
-        raise ConfigError("Missing correct conifg:upload parameters")
 
 
 def setup_parser(subparser):
     sp = subparser.add_subparsers(metavar='SUBCOMMAND',
                                   dest='results_command')
 
-    for name in subcommands:
-        if isinstance(name, (list, tuple)):
-            name, aliases = name[0], name[1:]
+    # Upload
+    upload_parser = sp.add_parser('upload', help=results_upload.__doc__)
+    upload_parser.add_argument(
+        'filename', help='path of file to upload')
+
+
+def results_upload(args):
+    """Imports Ramble experiment results from JSON file and uploads them as
+    specified in the upload block of Ramble's config file."""
+    imported_results = import_results_file(args.filename)
+
+    ramble.experimental.uploader.upload_results(imported_results)
+
+
+def import_results_file(filename):
+    """
+    Import Ramble experiment results from a JSON file.
+    """
+    tty.debug("File to import:")
+    tty.debug(filename)
+
+    imported_file = open(filename)
+
+    try:
+        tty.msg("Import file...")
+        parsed_json_file = json.load(imported_file)
+        # Check if data contains an experiment
+        if parsed_json_file.get('experiments'):
+            return parsed_json_file
         else:
-            aliases = []
-
-        # add commands to subcommands dict
-        function_name = 'results_%s' % name
-        function = globals()[function_name]
-        for alias in [name] + aliases:
-            subcommand_functions[alias] = function
-
-        # make a subparser and run the command's setup function on it
-        setup_parser_cmd_name = 'results_%s_setup_parser' % name
-        setup_parser_cmd = globals()[setup_parser_cmd_name]
-
-        subsubparser = sp.add_parser(
-            name, aliases=aliases, help=setup_parser_cmd.__doc__)
-        setup_parser_cmd(subsubparser)
+            tty.die("Error parsing file: Does not contain valid data to upload.")
+    except ValueError:
+        tty.die("Error parsing file: Invalid JSON formatting.")
 
 
 def results(parser, args):
-    """Look for a function called environment_<name> and call it."""
-    action = subcommand_functions[args.results_command]
-    action(args)
+    action = {'upload': results_upload}
+    action[args.results_command](args)
--- a/lib/ramble/ramble/cmd/results.py
+++ b/lib/ramble/ramble/cmd/results.py
@@ -1,0 +1,95 @@
+# Copyright 2022-2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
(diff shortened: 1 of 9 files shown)


lib/ramble/ramble/experimental/uploader.py
```diff
@@ -161,6 +161,27 @@ def format_data(data_in):
     return results
 
 
+def results_file_import(filename):
+    """
+    Import Ramble experiment results from a JSON file.
+    """
+    tty.debug("Import file")
```
rfbgo, line 168: We print "import file" twice, can we reduce that somehow?

lib/ramble/ramble/experimental/uploader.py
```diff
@@ -161,6 +161,27 @@ def format_data(data_in):
...
+    tty.debug(filename)
+
+    imported_file = open(filename)
+
+    try:
+        tty.msg("Import file...")
```
rfbgo, line 174: Maybe this is better as a verb, ie `importing`?

lib/ramble/ramble/experimental/uploader.py
```diff
@@ -161,6 +161,27 @@ def format_data(data_in):
...
+        parsed_json_file = json.load(imported_file)
+        # Check if data contains an experiment
+        if parsed_json_file.get('experiments'):
+            return parsed_json_file
+        else:
+            tty.die("Error parsing file: Does not contain valid data.")
```
rfbgo, line 180:
```suggestion
            tty.die("Error parsing file: Does not contain valid data to upload.")
```

lib/ramble/ramble/cmd/results.py
```diff
@@ -0,0 +1,95 @@
...
+    'file_upload',
+]
+
+#: Dictionary mapping subcommand names and aliases to functions
+subcommand_functions = {}
+
+
+def results_file_upload_setup_parser(subparser):
+    subparser.add_argument(
+        'filename', metavar='filename',
```
rfbgo, line 29: Do we also want to support a short arg like `-f`?

lib/ramble/ramble/cmd/results.py
```diff
@@ -0,0 +1,95 @@
...
+from ramble.config import ConfigError
+import ramble.experimental.uploader
+
+
+description = "take actions on experiment results"
+section = "workspaces"
+level = "short"
+
+subcommands = [
+    'file_upload',
```
douglasjacobsen, line 20: Can you make this `file-upload` to be consistent with our other commands?

lib/ramble/ramble/cmd/results.py
```diff
@@ -0,0 +1,95 @@
...
+subcommand_functions = {}
+
+
+def results_file_upload_setup_parser(subparser):
+    subparser.add_argument(
+        'filename', metavar='filename',
+        help='path of file to upload')
+
+
+def results_file_upload(args):
```
douglasjacobsen, line 33: This logic looks like a replication of the upload logic in workspace.py. It might be nice to DRY it out and push it into the uploader class instead of having it exist in all of the individual classes. WDYT?

lib/ramble/ramble/cmd/results.py
```diff
@@ -0,0 +1,95 @@
...
+                                  dest='results_command')
+
+    for name in subcommands:
+        if isinstance(name, (list, tuple)):
+            name, aliases = name[0], name[1:]
+        else:
+            aliases = []
+
+        # add commands to subcommands dict
+        function_name = 'results_%s' % name
```
douglasjacobsen, line 78: This is super nit-picky since we aren't consistent in the whole code base yet, but I've been trying to use f-string formats for newer strings.

Can you change this (and other python 2 style string definitions) to the newer f-strings?

Or at a minimum use `str.format()` (but I prefer the f-string format personally)

share/ramble/ramble-completion.bash
```diff
@@ -595,6 +595,24 @@ _ramble_repo_rm() {
...
+        RAMBLE_COMPREPLY="file_upload"
+    fi
+}
+
+_ramble_results_file_upload() {
+    if $list_options
+    then
+        RAMBLE_COMPREPLY="-h --help"
+    else
+        RAMBLE=""
```
douglasjacobsen, line 612: This should be `RAMBLE_COMREPLY`. 

It's auto-generated, but you're using an older version of develop which is why it's generated wrong.

lib/ramble/ramble/test/cmd/results.py
```diff
@@ -0,0 +1,40 @@
...
+            py.path.local(INPUT_DATA).join('test2_not_json.txt.json'),
+            'Error parsing file: Invalid JSON formatting.',
+        ),
+        (
+            py.path.local(INPUT_DATA).join('test3_malformed_json.json'),
+            'Error parsing file: Invalid JSON formatting',
+        ),
+    ],
+)
+def test_file_import_rejects_invalid_files(filename, expected_output, capsys):
```
douglasjacobsen, line 36: This is super concise. :)

douglasjacobsen (changes requested):

lib/ramble/ramble/cmd/results.py
```diff
@@ -0,0 +1,95 @@
...
+
+def results_file_upload(args):
+    """Imports Ramble experiment results from JSON file and uploads them."""
+    imported_results = ramble.experimental.uploader.results_file_import(args.filename)
+
+    if ramble.config.get('config:upload'):
+        # Read upload type and push it there
+        if ramble.config.get('config:upload:type') == 'BigQuery':  # TODO: enum?
+            try:
+                formatted_data = ramble.experimental.uploader.format_data(imported_results)
```
rfbgo, line 41: Small nit, but given we call into the uploader multiple time, I think it's probably better if we move a lot of these upload calls outside of command (either into workspace or the uploader) 

You can see for the `analyze --upload` command we chose to do most of the logic outside the command, what do you think?

rfbgo (changes requested):
Awesome contribution, thank you! 

Looks good to me, I have a few little tweaks and changes to help make it even better. Take a look and let me know what you think

lib/ramble/ramble/experimental/uploader.py
dapomeroy: This is mirroring lines 242 ("Upload experiments...") and 247 ("Upload FOMs..."). I don't have strong feelings for which is better but just wanted to maintain consistency. Shall I update all of them to verbs?

lib/ramble/ramble/experimental/uploader.py
dapomeroy: Adjusted this to clarify that it's listing the filename for debuggin purpose (mirroring how it's done on lines 239-240)

lib/ramble/ramble/cmd/results.py
rfbgo: When I wrote this I forgot we don't have an active workspace, maybe the more appropriate thing to do is to push a lot of this logic down into the uploader class?

lib/ramble/ramble/experimental/uploader.py
rfbgo: Ah yea you're right, thanks for calling that out

I think this is probably fine as is then

lib/ramble/ramble/cmd/results.py
dapomeroy: I think that sounds right. Going to talk to Doug about how to do that and see if we need to update the 'workspace analyze upload' logic to use the same method

dapomeroy: Removed unused 'workspace name' parameter from perform_upload(), which enabled upload_results() to be moved into uploader module and used for both workspace upload and results import upload.

lib/ramble/ramble/cmd/results.py
dapomeroy: Completed this

lib/ramble/ramble/cmd/results.py
dapomeroy: Done

lib/ramble/ramble/cmd/results.py
dapomeroy: Since filename is the only argument for upload, it doesn't take a flag. Syntax is 'ramble results upload <filename>'

lib/ramble/ramble/cmd/results.py
```diff
@@ -0,0 +1,89 @@
...
+        tty.die("Error parsing file: Invalid JSON formatting.")
+
+
+def setup_parser(subparser):
+    sp = subparser.add_subparsers(metavar='SUBCOMMAND',
+                                  dest='results_command')
+
+    for name in subcommands:
+        if isinstance(name, (list, tuple)):
+            name, aliases = name[0], name[1:]
```
rfbgo, line 67: I know you copied this from elsewhere, but can we remove this unneeded alias stuff? 

We don't need aliases here (yet?), but when we do it feels like we would want to pull this out into a common function somewhere than repeat it for a third time

rfbgo (changes requested):
Nice changes! Small change and lets squash/clean the history a bit and we should be good to go!

lib/ramble/ramble/cmd/results.py
dapomeroy: Reworked this to use the parser setup pattern from repo.py

lib/ramble/ramble/experimental/uploader.py
```diff
@@ -118,7 +146,7 @@ def format_data(data_in):
 
     .. code-block:: text
 
-        { expierment_name:
+        { expiriment_name:
```
douglasjacobsen, line 149:
```suggestion
        { experiment_name:
```

douglasjacobsen (changes requested):
LGTM. One small comment though.

douglasjacobsen (approved):

rfbgo (approved):

PR #229, opened 2023-07-27
Allow Ramble to track the origin of a FOM

rfbgo:
This change allows ramble to check if a FOM came from an app or a mod (and later perhaps that it is in fact a derived FOM)

This is useful as it allow the user to tell if a fom came from an app or a mod during analysis, and can be decorate as such in text print

This change also allows for the merging of contexts when a modifier and application emit the same context. Previously the applications context was lost.

### DIFF
--- a/lib/ramble/ramble/language/shared_language.py
+++ b/lib/ramble/ramble/language/shared_language.py
@@ -201,7 +201,7 @@ def success_criteria(name, mode, match=None, file='{log_file}',
             'file': file,
             'fom_name': fom_name,
             'fom_context': fom_context,
-            'formula': formula,
+            'formula': formula
         }
 
     return _execute_success_criteria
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/language/shared_language.py
```diff
@@ -201,7 +201,7 @@ def _execute_success_criteria(obj):
             'file': file,
             'fom_name': fom_name,
             'fom_context': fom_context,
-            'formula': formula,
+            'formula': formula
```
douglasjacobsen, line 204: Is there a reason the comma had to be removed?

douglasjacobsen (commented):
This looks good. It's nice how easy it was to make the merging work!

Once some testing is added, it'll probably be ready to merge. :)

lib/ramble/ramble/language/shared_language.py
rfbgo: No, it just made it more consistent with the rest of the file. Happy to put it back / make all in the file have trailing `,` / leave as is

douglasjacobsen (approved):

PR #236, opened 2023-08-07
Cleanup tags on applications

dodecatheon:
Clean up tags on applications

### DIFF
--- a/var/ramble/repos/builtin/applications/cloverleaf/application.py
+++ b/var/ramble/repos/builtin/applications/cloverleaf/application.py
@@ -18,9 +18,7 @@ class Cloverleaf(SpackApplication):
 
     maintainers('dodecatheon')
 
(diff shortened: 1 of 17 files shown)


douglasjacobsen (changes requested):
We discussed this in one of our internal meetings.

I don't feel like this goes quite far enough at minimizing redundant tags, but also we need to decide how we want to deal with things like the `hcls` and `life-sciences` tags.

dodecatheon:
> We discussed this in one of our internal meetings.
> 
> I don't feel like this goes quite far enough at minimizing redundant tags, but also we need to decide how we want to deal with things like the `hcls` and `life-sciences` tags.

Good point. They could easily be added on non-public application repos if needed, as you said. Shall I remove them?

douglasjacobsen:
I think we should discuss this in our meeting tomorrow, and decide what we want to do with the tags moving forward.

douglasjacobsen:
Given the discussion we had a while ago, I'm going to close this PR.

PR #243, opened 2023-08-17
Clean up context attribute definitions

dapomeroy:
Creates a context object and simplifies how contexts are stored and passed.

### DIFF
--- a/lib/ramble/ramble/context.py
+++ b/lib/ramble/ramble/context.py
@@ -1,0 +1,142 @@
+# Copyright 2022-2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.namespace import namespace
+import ramble.util.matrices
+
+import spack.util.spack_yaml as syaml
+
+
+class Context(object):
+    """Class to represent a context
+
+    This class contains variable definitions to store any individual context
+    (such as application, workload, or experiment) and logic to merge in
+    additional contexts by order of precedence."""
+
+    def __init__(self):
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/context.py
```diff
@@ -0,0 +1,142 @@
...
+            self.is_template = in_context.templates
+        if in_context.exclude:
+            self.exclude = in_context.exclude
+        if in_context.zips:
+            self.zips = in_context.zips
+        if in_context.matrices:
+            self.matrices = in_context.matrices
+
+
+def from_dict(context_name, in_dict):
```
douglasjacobsen, line 82: I think this should either be a member function of `Context` (where it should set the context from a dictionary), i.e.:
```
new_context = Context()
new_context.from_dict(in_dict)
```

Or, the name should be more like `create_context_from_dict`.

douglasjacobsen (changes requested):
This is great. Thanks for this!

One small request, feel free to do it either way I suggested.

lib/ramble/ramble/context.py
dapomeroy: Great, I've updated the name. I might be overthinking this, but my reasoning was that having this as a non-member function would ensure that updates to an existing instance go through the 'merge_context' logic. That way, a context 'from_dict' would start with a clean instance and avoid any potential conflict with the 'from_dict' method being invoked after vars have already been set and either not overwriting or not properly merging the two contexts.

lib/ramble/ramble/context.py
```diff
@@ -0,0 +1,142 @@
...
+
+        Create a Context object, which holds context attributes.
+        """
+        self.env_variables = []
+        self.variables = syaml.syaml_dict()
+        self.internals = {}
+        self.templates = None
+        self.chained_experiments = []
+        self.modifiers = []
+        self.context_names = None
```
douglasjacobsen, line 33: One other small nit, that I just noticed when you changed the name of the `from_dict` function.

Is there a reason for this to be plural? It seems like it should be singular instead. But, if there is a reason then we can leave it.

PR #244, opened 2023-08-21
Add support for phase selections

douglasjacobsen:
This merge adds support for workspace commands to define which phases of their pipeline they should execute.

As an example:

```
ramble workspace setup --phases *software*
```
will execute only phases that contain the string `software` in their names.

Additionally, `ramble workspace info` is updated to print all of the phases any experiment within the workspace will execute.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -56,6 +56,7 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
     _builtin_required_key = 'required'
     _workload_exec_key = 'executables'
     _inventory_file_name = 'ramble_inventory.json'
+    _pipelines = ['analyze', 'archive', 'mirror', 'setup']
 
     #: Lists of strings which contains GitHub usernames of attributes.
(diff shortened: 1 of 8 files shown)


lib/ramble/ramble/application.py
```diff
@@ -64,6 +65,7 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
     def __init__(self, file_path):
         super().__init__()
 
+        self._pipelines = ['setup', 'analyze', 'archive', 'mirror']
```
rfbgo, line 68: Should we pull this out and validate calls to `run_pipeline` against it? I feel like we currently have the different pipelines strongly implied in different places and adding a new pipeline might be a bit odd/error prone right now?

rfbgo (approved):

lib/ramble/ramble/application.py
douglasjacobsen: Should be better now. Let me know if you have concerns with my latest update. Thanks for the suggestion!

rfbgo (approved):

PR #245, opened 2023-08-21
Add support for prefix on spack commands

douglasjacobsen:
This merge adds support for a config:

  config:spack:install:prefix

Which can define a prefix to all `spack install` executions (allowing parallel installs).

Additionally, this updates the format of the config config_section, which can be applied using `ramble config update config`. This change is also applied to the `concretize` command.

Tests and documentation are updated as well.


douglasjacobsen:
I don't think this is quite working how I intended. So I converted it to a draft for now.

douglasjacobsen:
It's working now.

etc/ramble/defaults/config.yaml
```diff
@@ -1,27 +1,14 @@
...
-# Users can override these settings by editing the following files.
-#
-# Per-ramble-instance settings (overrides defaults):
-#   $RAMBLE_ROOT/etc/ramble/config.yaml
-#
-# Per-user settings (overrides default and site settings):
-#   ~/.ramble/config.yaml
-# -------------------------------------------------------------------------
 config:
-    test_stage: ~/.ramble/test
```
rfbgo, line 17: Did you want to keep these comments?

rfbgo (dismissed):

etc/ramble/defaults/config.yaml
douglasjacobsen: Yes, thanks, I added them back in.

rfbgo (approved):

PR #257, opened 2023-08-31
Redirect prints from Spack to log file

dapomeroy:
Redirects prints from Spack to a log file to provide a cleaner command line experience for users. Logger wraps everything within the run_pipeline() method in workspace.py, which currently covers workspace setup, analyze, and mirror commands. All of stdout gets redirected into the logfile, unless inside a logger.force_echo() block. Added force_echo() blocks to all errors / tty.die within spack.py to ensure errors get printed to stdout.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -20,6 +20,7 @@ from typing import List
 import llnl.util.filesystem as fs
 import llnl.util.tty as tty
 import llnl.util.tty.color as color
+import llnl.util.tty.log
 from llnl.util.tty.colify import colified
 
 import spack.util.executable
@@ -105,6 +106,8 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
         self.license_path = ''
         self.license_file = ''
         self.license_inc_name = 'license.inc'
+
+        self.logger = llnl.util.tty.log.log_output(echo=False, debug=tty.debug_level())
 
     def copy(self):
         """Deep copy an application instance"""
@@ -278,6 +281,13 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
         """Set modifiers for this instance"""
         if modifiers:
             self.modifiers = modifiers.copy()
+
+    def experiment_log_file(self, logs_dir):
+        """Returns an experiment log file path for the given logs directory"""
+        return os.path.join(
+            logs_dir,
+            self.expander.experiment_namespace) + \
+            '.out'
 
     def get_pipeline_phases(self, pipeline, phase_filters=['*']):
         if pipeline not in self._pipelines:
--- a/lib/ramble/ramble/test/end_to_end/explicit_zips.py
+++ b/lib/ramble/ramble/test/end_to_end/explicit_zips.py
@@ -15,6 +15,7 @@ import ramble.workspace
 import ramble.config
 import ramble.software_environments
 from ramble.main import RambleCommand
+from ramble.test.dry_run_helpers import search_files_for_string
 
 
 # everything here uses the mock_workspace_path
@@ -133,8 +134,12 @@ licenses:
 
         ws1._re_read()
 
-        output = workspace('setup', '--dry-run', global_args=['-w', workspace_name])
-        assert "Would download https://www2.mmm.ucar.edu/wrf/users/benchmark/v422/v42_bench_conus12km.tar.gz" in output
+        workspace('setup', '--dry-run', global_args=['-w', workspace_name])
+
+        out_files = glob.glob(os.path.join(ws1.log_dir, '**', '*.out'), recursive=True)
+
+        assert search_files_for_string(
+            out_files, 'Would download https://www2.mmm.ucar.edu/wrf/users/benchmark/v422/v42_bench_conus12km.tar.gz') # noqa
 
         # Test software directories
         software_dirs = ['wrfv4.CONUS_12km', 'wrfv4-portable.CONUS_12km']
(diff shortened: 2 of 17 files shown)


lib/ramble/ramble/test/end_to_end/explicit_zips.py
```diff
@@ -24,6 +24,14 @@
 workspace = RambleCommand('workspace')
 
 
+def search_files_for_string(file_list, string):
```
douglasjacobsen, line 27: Remove and import

lib/ramble/ramble/test/end_to_end/phase_selection.py
```diff
@@ -23,6 +24,14 @@
 workspace = RambleCommand('workspace')
 
 
+def search_files_for_string(file_list, string):
```
douglasjacobsen, line 27: Remove and import

lib/ramble/ramble/test/end_to_end/unused_compilers_are_skipped.py
```diff
@@ -23,6 +24,14 @@
 workspace = RambleCommand('workspace')
 
 
+def search_files_for_string(file_list, string):
```
douglasjacobsen, line 27: Remove and import

lib/ramble/ramble/test/end_to_end/wrfv4_dry_run.py
```diff
@@ -24,6 +24,14 @@
 workspace = RambleCommand('workspace')
 
 
+def search_files_for_string(file_list, string):
```
douglasjacobsen, line 27: Remove and import

lib/ramble/ramble/test/modifier_functionality/mock_spack_modifier.py
```diff
@@ -18,6 +19,14 @@
 workspace = RambleCommand('workspace')
 
 
+def search_files_for_string(file_list, string):
```
douglasjacobsen, line 22: Remove and import

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1089,82 +1095,137 @@ def create_mirror(self, mirror_root):
...
+            experiment_set = self.build_experiment_set()
+
+            experiment_count = len(experiment_set.experiments) + \
+                len(experiment_set.chained_experiments)
+            with self.logger.force_echo():
+                tty.msg(f'  Working on {experiment_count} experiments:')
+
+        for exp, app_inst, count in experiment_set.all_experiments():
+            # Set up an experiment logger and a log file for each experiment
+            exp_log_file = exp + '.out'
```
douglasjacobsen, line 1160: The application instance should own this filename creation.

Maybe something like:
`app_inst.log_file_name`

douglasjacobsen (changes requested):
Thanks for this. Overall it looks good, I have a few things that should be changed in the PR, and one bug that I ran into while testing it.

It seems if you use debug mode, most of the debug output is just lost entirely and not written to any of the log files.

Verbose appears to work correctly, but debug doesn't for some reason.

dapomeroy:
All fixed up. Just needed to pass the global debug setting to the logger.

lib/ramble/ramble/application.py
```diff
@@ -106,6 +107,9 @@ def __init__(self, file_path):
         self.license_file = ''
         self.license_inc_name = 'license.inc'
 
+        self.logger = llnl.util.tty.log.log_output(echo=False, debug=tty.debug_level())
+        self.experiment_log_file = None
```
douglasjacobsen, line 111: I don't think we need this attribute at all here (I'll suggest the other changes below though, that make this not necessary)

lib/ramble/ramble/application.py
```diff
@@ -279,6 +283,10 @@ def set_modifiers(self, modifiers):
         if modifiers:
             self.modifiers = modifiers.copy()
 
+    def set_experiment_log_file(self, experiment_name):
```
douglasjacobsen, line 286: Here, instead of setting the attribute, you can do something like:

```
def experiment_log_file(self, logs_dir):
    return os.path.join(
        logs_dir
        self.expander.expand_var_name(keywords.experiment_name)) + \
        '.out'
```

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1089,82 +1095,137 @@ def create_mirror(self, mirror_root):
...
+
+            experiment_count = len(experiment_set.experiments) + \
+                len(experiment_set.chained_experiments)
+            with self.logger.force_echo():
+                tty.msg(f'  Working on {experiment_count} experiments:')
+
+        for exp, app_inst, count in experiment_set.all_experiments():
+            # Set up an experiment logger and a log file for each experiment
+            app_inst.set_experiment_log_file(exp)
+            exp_log_path = os.path.join(self.experiment_log_dir, app_inst.experiment_log_file)
```
douglasjacobsen, line 1161: Then here, the `exp_log_path` is just a single call to the app_inst function described in the previous comment.

douglasjacobsen (changes requested):
Thanks! Debug works now. A few more nits about the log file stuff though.

lib/ramble/ramble/application.py
dapomeroy: Used experiment_namespace instead to create separate files for each app.workload.experiment

douglasjacobsen (approved):

PR #270, opened 2023-10-02
Add 'Running an Experiment' tutorial to docs

dapomeroy:
Update 'Getting Started' doc and replace 'General Usage' with 'Running an Experiment' tutorial using basic GROMACS config.


lib/ramble/docs/getting_started.rst
```diff
@@ -12,32 +12,68 @@
...
+* Compiler flag sweeps
+
 --------------------
 System Requirements
 --------------------
 
-Ramble's dependencies are listed within the top level
-requirements.txt file.
+Ramble requires Python 3.6.8 or greater and
+`Spack <https://spack.io/about/#install-spack>`_, which is used for managing
```
douglasjacobsen, line 38: Ramble doesn't *require* spack, so if this is written in the readme, we should change it.

Some applications within ramble require spack, but other ones (i.e. `hostname`) do not

lib/ramble/docs/tutorials.rst
```diff
@@ -19,5 +19,5 @@ various features.
     :maxdepth: 1
     :caption: Tutorials
 
-    tutorials/general_usage
+    tutorials/running_an_experiment
```
douglasjacobsen, line 22: Maybe the name should have gromacs in it, just in case we want to add other apps later.

like:
`running_a_simple_gromacs_experiment`

lib/ramble/docs/tutorials/running_an_experiment.rst
```diff
@@ -0,0 +1,270 @@
...
+   Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+   https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+   <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+   option. This file may not be copied, modified, or distributed
+   except according to those terms.
+
+.. _running_an_experiment_tutorial:
+
+======================
+Running An Experiment
```
douglasjacobsen, line 12: Same comment about adding Gromacs

lib/ramble/docs/tutorials/running_an_experiment.rst
```diff
@@ -0,0 +1,270 @@
...
+
+.. code-block:: console
+
+    $ ramble attributes --all --tags
+
+^^^^^^^^^^^^^^^^^^^^^^^^^
+What's in an application?
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Knowing what applications are available is only part of how a user interacts
```
douglasjacobsen, line 69: I'll try to generally apply this comment, but I think we can change `user` to a more active `you` in text like this.

i.e. this is a tutorial, so we can directly tell people that they should do things.

lib/ramble/docs/tutorials/running_an_experiment.rst
```diff
@@ -0,0 +1,270 @@
...
+command can be used to get more information about a specific application
+definition. As an example:
+
+.. code-block:: console
+
+    $ ramble info gromacs
+
+Will print the workloads and variables the ``GROMACS`` application definition contains.
+
+--------------------
```
douglasjacobsen, line 81: (I will try to come back to this, after I verify) If we don't use modifiers, we should remove this section from this tutorial.

lib/ramble/docs/tutorials/running_an_experiment.rst
```diff
@@ -0,0 +1,270 @@
...
+                  variables:
+                    n_ranks: '1'
+                    n_threads: '1'
+                    size: '0003'
+                    type: 'rf'
+      spack:
+        concretized: true
+        packages:
+          gcc9:
+            spack_spec: gcc@9.3.0 target=x86_64
```
douglasjacobsen, line 199: We might want to call out that installing compilers that spack doesn't know about could take a while.

And give users a way to define their own compilers if they want to.

lib/ramble/docs/tutorials/running_an_experiment.rst
```diff
@@ -0,0 +1,270 @@
...
+.. code-block:: console
+
+    $ ramble workspace setup
+
+This command will create experiment directories, download and expand input files,
+and install the required software stack (and generate spack environments for
+each workload).
+
+It can take a bit to run depending on if you need to build new software and how
+long the input files take to download.
```
douglasjacobsen, line 234: Maybe add some info about how the logging works, to point people to the different log files.

douglasjacobsen (changes requested):

lib/ramble/docs/getting_started.rst
dapomeroy: Edited to 'some applications also require Spack...' Does that sound OK as a way to prompt folks to install Spack?

lib/ramble/docs/tutorials/running_an_experiment.rst
dapomeroy: Would it be worth adding a modifier to the tutorial yaml to demo how to use it?

lib/ramble/docs/tutorials/running_an_experiment.rst
dapomeroy: Is this ok verbiage? "Note that specifying compilers that Spack doesn't have installed may take a while.
To see available compilers, use ``spack compilers``."

lib/ramble/docs/tutorials/running_an_experiment.rst
douglasjacobsen: I think in a future tutorial. For now, lets keep the intro tutorial pretty simple.

Speaking of which, it might be good to number this tutorial (1) so people know it should go first.

lib/ramble/docs/tutorials/running_an_experiment.rst
douglasjacobsen: Yes, I might even suggest linking to: https://spack.readthedocs.io/en/latest/getting_started.html#spack-compilers

lib/ramble/docs/getting_started.rst
douglasjacobsen: Yeah, that sounds good.

douglasjacobsen (approved):

douglasjacobsen:
Thanks! I'm going to merge this because `namd` isn't an actual typo / spelling error.

PR #274, opened 2023-10-05
Add simple prototype of functionality to push to spack caches

rfbgo:
I had a simple idea to push caches over breakfast, so threw together a quick prototype 

This is 90% glue but I wanted to experiment with how hard it would be to push a ramble generated env to a spack cache for later use (turns out its super easy!) 

This PR is more of a PoC and I can tidy it up a bit if we decide it's a feature we are interested in making mainline


lib/ramble/ramble/application_types/spack.py
```diff
@@ -365,6 +365,21 @@ def _mirror_software(self, workspace):
             with self.logger.force_echo():
                 tty.die(e)
 
+    register_phase('push_to_spack_cache', pipeline='pushtocache', depends_on=[])
+
+    def _push_to_spack_cache(self, workspace):
+        try:
+            # TODO: make sure this is called as infrequently as possible
+            # TODO: this is a lot of repeated stuff, can we DRY with out pipelines?
```
douglasjacobsen, line 373: It is called a lot of times. Maybe there's a better way we can do it though.
I don't have any great suggestions on where this should go though. Maybe a wrapper within the runner?

You probably also want to make sure you deactivate the spack_runner.

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -51,6 +51,7 @@
     'concretize',
     'setup',
     'analyze',
+    'pushtocache',
```
douglasjacobsen, line 54: Can names have hyphens?

lib/ramble/ramble/pipeline.py
```diff
@@ -319,13 +319,33 @@ def _complete(self):
...
+
+    def _prepare(self):
+        super()._prepare()
+        self.workspace.spack_cache_path = self.spack_cache_path
+
+    def _complete(self):
+        tty.msg('Pushed envs to spack cache %s' % self.spack_cache_path)
+
+
+# TODO: should this use the classes `.name`?
```
douglasjacobsen, line 340: Yeah, I think there are several ways we can remove this. I think I only left it for error checking (i.e. trying to create a pipeline that doesn't exist yet)

lib/ramble/ramble/spack_runner.py
```diff
@@ -676,6 +676,36 @@ def mirror_environment(self, mirror_path):
...
+    def push_to_spack_cache(self, spack_cache_path):
+        """Push packages for a given env to the spack cache"""
+        self._check_active()
+
+        hash_list = self.get_env_hash_list()
+
+        args = [
+            "buildcache",
+            "push",
+            "-a",  # TODO: remove this, it's just here for testing
```
douglasjacobsen, line 698: I think we'll just need a way to pass arguments in to push. Maybe the same way we pass in args for install and concretize?

lib/ramble/ramble/application_types/spack.py
rfbgo: Good catch on the deactivate.

I added an `if already_pushed` to the spack_runner which will help some

lib/ramble/ramble/cmd/workspace.py
rfbgo: Right now, in short, no 

The reason is that we directly map this dict to a function named `workspace_{key}` which would mean we need a function with hypens in. Not impossible to remap if we do want it but does not work out of the box today

lib/ramble/ramble/pipeline.py
rfbgo: OK I will make a task to clean this up and remove the TODO here

lib/ramble/ramble/spack_runner.py
rfbgo: Good idea, done

lib/ramble/ramble/cmd/workspace.py
rfbgo: Updated this functionality in a seperate PR (#286) and added it here

douglasjacobsen (approved):

PR #281, opened 2023-10-10
simlink log files to allow for known latest path

rfbgo:
It will give:

```
ls -al ramble/var/ramble/workspaces/$WORKSPACE_NAME/logs/
...
544 Oct 10 11:09 analyze.2023-10-10_11.08.59
120 Oct 10 11:09 analyze.2023-10-10_11.08.59.out
86 Oct 10 11:08 analyze.latest -> $HOME/ramble/var/ramble/workspaces/WORKSPACE_NAME/logs/analyze.2023-10-10_11.08.59
90 Oct 10 11:08 analyze.latest.out -> $HOME/ramble/var/ramble/workspaces/WORKSPACE_NAME/logs/analyze.2023-10-10_11.08.59.out
```

### DIFF
--- a/lib/ramble/ramble/pipeline.py
+++ b/lib/ramble/ramble/pipeline.py
@@ -48,9 +48,18 @@ class Pipeline(object):
         log_file = f'{self.name}.{dt}.out'
         self.log_dir = os.path.join(self.workspace.log_dir,
                                     f'{self.name}.{dt}')
+        self.log_dir_latest = os.path.join(self.workspace.log_dir,
+                                           f'{self.name}.latest')
         self.log_path = os.path.join(self.workspace.log_dir,
                                      log_file)
+        self.log_path_latest = os.path.join(self.workspace.log_dir,
+                                            f'{self.name}.latest.out')
+
         fs.mkdirp(self.log_dir)
+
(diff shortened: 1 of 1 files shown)


lib/ramble/ramble/pipeline.py
```diff
@@ -159,6 +168,15 @@ def run(self):
         with self.workspace.logger(self.log_path, echo=self.workspace.dry_run):
             self._complete()
 
+    def simlink_log(self, base, link):
+        """
+        Create simlink of log file to give a known and predictable path
+        """
+        if os.path.islink(link):
+            os.unlink(link)
```
douglasjacobsen, line 176: I don't know if we do this elsewhere, but should we catch a case where `link` is created, and is not a link? Or should we just let the `symlink` call below fail?

rfbgo: What we decided last time was to let the simlink fail to avoid clobbering / silently giving confusing behavior

douglasjacobsen: OK cool. Thanks!

douglasjacobsen (approved):

PR #285, opened 2023-10-11
Clean out unneeded cachcing in spack runner

rfbgo:
Leave concretized as we can import an external env


lib/ramble/ramble/spack_runner.py
```diff
@@ -103,7 +103,6 @@ def __init__(self, shell='bash', dry_run=False):
         self.compilers = []
         self.includes = []
         self.dry_run = dry_run
-        self.concretized = False
```
douglasjacobsen, line 106: Is this removing the right line? I think we want to remove https://github.com/GoogleCloudPlatform/ramble/blob/907605e84dce79b327aa18c149bff46333f88930/lib/ramble/ramble/spack_runner.py#L99 instead, right?

douglasjacobsen: Oh, maybe this is a repeat. So, we probably want to remove both.

rfbgo: Yea we remove installed because it's dead code, and one of the conretzied because it's a dupe

douglasjacobsen (approved):

PR #298, opened 2023-10-25
Allow the setup pipeline to skip/fail workspace inventory generation

rfbgo:
Currently any call into setup hard requires an inventory, regardless of the phases selected. This causes issues as not all phases are able to safely generate the inventory as it depends on the software being installed

This PR allows the inventory to be optional and for the execution to proceed when non-software phases are selected (eg get_inputs)

@douglasjacobsen let me know what you think of this. It would be nice if we were more able to discriminately set `require_inventory = True` again when it makes sense to

### DIFF
--- a/lib/ramble/ramble/pipeline.py
+++ b/lib/ramble/ramble/pipeline.py
@@ -319,7 +319,7 @@ class SetupPipeline(Pipeline):
     def __init__(self, workspace, filters):
         super().__init__(workspace, filters)
         self.force_inventory = True
-        self.require_inventory = True
+        self.require_inventory = False
         self.action_string = 'Setting up'
 
     def _prepare(self):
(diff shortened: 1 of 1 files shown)


douglasjacobsen:
I think this looks good as a short term fix.

Maybe we can set `require_inventory` to true if the phases aren't limited as a start though.

rfbgo:
> I think this looks good as a short term fix.
> 
> Maybe we can set `require_inventory` to true if the phases aren't limited as a start though.

Yea I couldn't decide what the right answer was, and I didn't want to do something that only half addressed it that "looked complete" 

One idea I through around was to check if the `write_inventory` phase was active then switch that var to `True`

Happy to implement whatever subset you think is best

douglasjacobsen:
Yeah, that's not a bad idea. Given the dependencies are encoded now, you shouldn't be able to run `write_inventory` without `software_configure` (for spack applications)

douglasjacobsen:
Maybe go ahead and add a check that `write_inventory` is included. Then we can see if we should clean it up again in the future.

rfbgo:
Sounds good, done

douglasjacobsen (approved):

PR #299, opened 2023-10-25
update hashes on hmmer to match current release

rfbgo:
At some point we should move this over to a fixed version oh the inputs instead of being on latest, but this at least makes the package work until the next update

### DIFF
--- a/var/ramble/repos/builtin/applications/hmmer/application.py
+++ b/var/ramble/repos/builtin/applications/hmmer/application.py
@@ -35,6 +35,7 @@ class Hmmer(SpackApplication):
     software_spec('hmmer', spack_spec='hmmer@3.3.2', compiler='gcc9')
 
     # This would ideally not use the current_relase, as the package will need to be manually updated per release
+    # Here current_release == 'Pfam36.0'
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/hmmer/application.py
```diff
@@ -34,15 +34,16 @@ class Hmmer(SpackApplication):
 
     software_spec('hmmer', spack_spec='hmmer@3.3.2', compiler='gcc9')
 
+    # This would ideally not use the current_relase, as the package will need to be manually updated per release
```
douglasjacobsen, line 37: If possible, can you document what the current version of "current_release" is, so we don't have to figure this out in the future?

douglasjacobsen (changes requested):

var/ramble/repos/builtin/applications/hmmer/application.py
rfbgo: Done (its Pfam36.0)

douglasjacobsen (approved):

PR #302, opened 2023-10-26
Fix nested compiler install

douglasjacobsen:
This merge fixes compiler installation. Previously, if a compiler needed to be installed, and was required to be installed with a different specific compiler the installation specs were incorrect.

This merge extracts and installs all compilers, including the compilers they are dependent on.

### DIFF
--- a/lib/ramble/ramble/application_types/spack.py
+++ b/lib/ramble/ramble/application_types/spack.py
@@ -82,11 +82,11 @@ class SpackApplication(ApplicationBase):
 
     def _software_install_requested_compilers(self, workspace):
         """Install compilers an application uses"""
-
         # See if we cached this already, and if so return
         env_path = self.expander.env_path
         if not env_path:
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/application_types/spack.py
```diff
@@ -83,11 +83,11 @@ def _long_print(self):
 
     def _software_install_requested_compilers(self, workspace):
         """Install compilers an application uses"""
-
         # See if we cached this already, and if so return
         env_path = self.expander.env_path
         if not env_path:
             raise ApplicationError('Ramble env_path is set to None.')
+        tty.msg('Installing compilers')
```
rfbgo, line 90: Should we update this (and the other tty calls below) now the logger is in?

rfbgo (approved):
Small change, and need to fix conflicts

rfbgo (approved):

PR #311, opened 2023-11-08
Disable color characters in log files

douglasjacobsen:
This merge disables the printing of color strings when a log* file is active. Additionally, it lowers the precision of the elapsed time on the progress bar.

### DIFF
--- a/lib/ramble/ramble/util/logger.py
+++ b/lib/ramble/ramble/util/logger.py
@@ -8,6 +8,9 @@
 
 import llnl.util.tty as tty
 import llnl.util.tty.log
+import llnl.util.tty.color
+
+from contextlib import contextmanager
 
 
 class Logger(object):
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/util/logger.py
```diff
@@ -114,6 +117,14 @@ def _stream_kwargs(self, default_kwargs={}, index=None):
 
         return kwargs
 
+    @contextmanager
+    def configure_colors(self, **kwargs):
+        old_value = llnl.util.tty.color.get_color_when()
+        if 'stream' in kwargs:
+            llnl.util.tty.color.set_color_when('never')
+        yield
+        llnl.util.tty.color.set_color_when(old_value)
```
dapomeroy, line 126: Since the 'set_color_when(old_value)' is not nested in an 'else' and is not indented under 'yield', would this method always overwrite the 'if' and set it back to 'old_value' instead of 'never'?

douglasjacobsen: Yes. So, if we print without an active stream the logic here looks like:

```
get old value
(print stuff)
set value to old value
```

but the logic when a stream is active changes it to:
```
get old value
set value to never
(print stuff)
set value to old value
```

dapomeroy (approved):

PR #313, opened 2023-11-09
Add a hello world tutorial

douglasjacobsen:
This merge performs some general cleanup (mostly in support of the hello world tutorial), and adds a hello world tutorial to introduce users to ramble without adding a real application.

### DIFF
--- a/lib/ramble/docs/tutorials/hello_world.rst
+++ b/lib/ramble/docs/tutorials/hello_world.rst
@@ -1,0 +1,219 @@
+.. Copyright 2022-2023 Google LLC
+
+   Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+   https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+   <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+   option. This file may not be copied, modified, or distributed
+   except according to those terms.
+
+.. _hello_world_tutorial:
+
+=====================================================
+1) Getting Started Running A "Hello World" Experiment
+=====================================================
+
+This tutorial will provide a basic introduction to navigating Ramble and running
+experiments. In this tutorial, you will set up and run a basic experiment using
+``hostname``.
+
+Installation
+============
+
+To install Ramble, see the :doc:`../getting_started` guide.
+
+**NOTE**: This tutorial does not require ``spack`` to be installed or configured.
+
+Ramble Basics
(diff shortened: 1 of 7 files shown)


lib/ramble/docs/tutorials/hello_world.rst
```diff
@@ -0,0 +1,219 @@
...
+
+.. _hello_world_tutorial:
+
+=====================================================
+1) Getting Started Running A "Hello World" Experiment
+=====================================================
+
+This tutorial will provide a basic introduction to navigating Ramble and running
+experiments. In this tutorial, you will set up and run a basic experiment using
+``hostname``.
```
rfbgo, line 17: I think because hostname can mean lots of different things to different people, we might want to call out that it's a basic linux utilty that we're using as our application?

lib/ramble/docs/tutorials/hello_world.rst
```diff
@@ -0,0 +1,219 @@
...
+.. code-block:: console
+
+    $ ramble list
+
+The ``ramble list`` command also takes a query string, which can be used to
+filter available application definitions. For example:
+
+.. code-block:: console
+
+    $ ramble list h*
```
rfbgo, line 47: Does this definitely work? Does not work for me on OSX with either zsh or bash (but dropping the * does show any package with an h in, just not starting with)

lib/ramble/docs/tutorials/hello_world.rst
```diff
@@ -0,0 +1,219 @@
...
+
+Executing Experiments
+^^^^^^^^^^^^^^^^^^^^^^
+
+After the workspace is set up, its experiments can be executed. The two methods
+to run the experiments are:
+
+.. code-block:: console
+
+    $ ramble on
```
rfbgo, line 201: Shall we also "unhide" `ramble on`?

rfbgo (approved):
Amazing PR! 

Few small suggestions but LGTM

lib/ramble/docs/tutorials/hello_world.rst
douglasjacobsen: Yeah, probably. :)

lib/ramble/docs/tutorials/hello_world.rst
douglasjacobsen: Good catch. It's the same on my laptop.

PR #314, opened 2023-11-09
Add a tutorial about modifying existing experiments

douglasjacobsen:


lib/ramble/docs/tutorials/modifying_a_gromacs_experiment.rst
```diff
@@ -0,0 +1,221 @@
...
+3) Modifying A GROMACS Experiment
+=======================================
+
+In this tutorial, you will learn how to modify an existing workspace
+configuration that contains experiments using
+`GROMACS <https://www.gromacs.org/>`_, a free and open-source application
+for molecular dynamics.
+
+This tutorial builds off of concepts introduced in previous tutorials. Please
+make sure you review those before starting with this tutorial's content.
```
rfbgo, line 21: Do we need an explicit call out like "go setup the workspace from tutorial 2?"

lib/ramble/docs/tutorials/modifying_a_gromacs_experiment.rst
```diff
@@ -0,0 +1,221 @@
...
+
+.. code-block:: console
+
+    $ ramble workspace info -v
+
+The experiments section of this command's output might contain the following:
+
+.. code-block:: console
+
+    Experiments:
```
rfbgo, line 76: Why is this different from the output in tutorial 2?

lib/ramble/docs/tutorials/modifying_a_gromacs_experiment.rst
```diff
@@ -0,0 +1,221 @@
...
+            type:
+                Description: Workload type.
+                Default: pme
+                Suggested Values: ['pme', 'rf']
+            input_path:
+                Description: Input path for water bare hbonds
+                Default: {water_bare_hbonds}/{size}
+
+Using this information, change the size on each experiment based on the
+suggested values. **NOTE** The larger the value, the more expensive the
```
rfbgo, line 182: Do we want to link to a file with the edits done for the user, so they can click and compare they did it right at the end?

rfbgo: Also, bit of an annoying nit, but this tutorial about modify is >200 lines but this one line of "ok edit the thing!" is kind of subtle and hidden. Do we want to make it more obvious some how? I feel like people might skim over it and miss the point

rfbgo (approved):
Nice job. Few suggestions for possible improvements

lib/ramble/docs/tutorials/modifying_a_gromacs_experiment.rst
douglasjacobsen: One uses `ramble workspace info` (To show the experimens ) and the other uses `ramble info`  (to show the application info)

lib/ramble/docs/tutorials/modifying_a_gromacs_experiment.rst
douglasjacobsen: I think this should be there now (It's maybe hard to see when you look at the source, but there's an include line that pulls in the same "Create and configure the workspace" text as in tutorial 2.

lib/ramble/docs/tutorials/modifying_a_gromacs_experiment.rst
douglasjacobsen: I don't think I want to link to a file with the edits made. Mostly because I want the people going through the tutorial to pick a size and put it into the configuration file (which means it's harder to give a 'true' answer since there are like 9 different values they can use).

But yes, let me see if I can bring more emphasis to the editing part.

douglasjacobsen: OK, this is updated. PTAL

rfbgo (approved):
Thanks for the changes! One small question

lib/ramble/docs/tutorials/modifying_a_gromacs_experiment.rst
```diff
@@ -0,0 +1,248 @@
...
+-------------------
+
+Now that you know how to determine which values are possible, select one or
+more of the possible values for the ``size`` variable. You will modify the
+experiments in your workspace to use this (remember, you have 4 experiments
+defined currently). **NOTE** The larger the value, the more expensive the
+experiment will be. However you can give each experiment a unique value.
+
+To make editing the workspace easier, use the following command (assuming you
+have an ``EDITOR`` environment variable set):
```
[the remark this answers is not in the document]
douglasjacobsen: Should be updated now.

rfbgo (approved):
LGTM!

PR #321, opened 2023-11-20
Add a tutorial for creating experiments from scratch

douglasjacobsen:


lib/ramble/docs/tutorials/configuring_a_scaling_study.rst
```diff
@@ -0,0 +1,245 @@
...
+
+    ramble:
+      applications:
+        wrfv4:
+          workloads:
+            CONUS_12km:
+              experiments:
+                scaling_{n_nodes}:
+                  variables:
+                    n_nodes: [1, 2, 4]
```
rfbgo, line 107: I think in the context of the snippet below its not so bad, but do we want to also add ppn or n_ranks here? I know it complicates the example but this snippet also goes against are best (required) practices of specifying 2 of the 3

lib/ramble/docs/tutorials/configuring_a_scaling_study.rst
```diff
@@ -0,0 +1,245 @@
...
+This creates a ``results`` file in the root of the workspace that contains
+extracted figures of merit. If the experiments were successful, this file will
+show the following results:
+
+* Average Timestep Time: Time (in seconds) on average each timestep takes
+* Cumulative Timestep Time: Time (in seconds) spent executing all timesteps
+* Minimum Timestep Time: Minimum time (in seconds) spent on any one timestep
+* Maximum Timestep Time: Maximum time (in seconds) spent on any one timestep
+* Number of timesteps: Count of total timesteps performed
+* Avg. Max Ratio Time: Ratio of Average Timestep Time and Maximum Timestep Time
```
rfbgo, line 245:
```suggestion
* Avg. Max Ratio Time: Ratio of Average Timestep Time and Maximum Timestep Time

Ramble also supports uploading the analyzed data to online databases, using `ramble workspace analyze --upload`. We will not cover that there, but it is a very useful feature for production experiments
```

rfbgo: ie shall we add a plug for zep here? :)

rfbgo (approved):
Nice job!

lib/ramble/docs/tutorials/configuring_a_scaling_study.rst
douglasjacobsen: I can't decide. I was going to add `n_ranks` here, but that just adds the math and then is required to have `processes_per_node` defined as well. And I don't really want it to be defined in the experiment section.

Maybe it'd be good to inject the error that you would see if you tried to use this experiment directly, as a point of "Hey, you need to define these things now."

lib/ramble/docs/tutorials/configuring_a_scaling_study.rst
douglasjacobsen: Sure, I think we should also add (in the docs) what the supported formats for `--upload` are. I'll take care of adding this into the tutorial, and make a task to add docs around supported upload formats.

rfbgo (approved):
LGTM

PR #322, opened 2023-11-22
Stop using "shm:tcp" as it is no longer supported

rohitramu:
When running the Intel MPI microbenchmarks, this is the first line of the output:

```
MPI startup(): shm:tcp fabric is unknown or has been removed from the product, please use ofi or shm:ofi instead.
```

This PR updates the variable `I_MPI_FABRICS` to be `shm:ofi` (previously `shm:tcp`) and adds `FI_PROVIDER=tcp`.

### DIFF
--- a/var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
+++ b/var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
@@ -38,16 +38,11 @@ class IntelMpiBenchmarks(SpackApplication):
     required_package('intel-mpi-benchmarks')
 
     executable('pingpong',
-               '-genv FI_PROVIDER=tcp '
-               '-genv I_MPI_FABRICS=shm:ofi '
-               '-genv I_MPI_PIN_PROCESSOR_LIST=0 '
                '{install_path}/IMB-MPI1 {pingpong_type} -msglog {msglog_min}:{msglog_max} '
                '-iter {num_iterations} {additional_args}',
                use_mpi=True)
 
     executable('multi-pingpong',
-               '-genv FI_PROVIDER=tcp '
-               '-genv I_MPI_FABRICS=shm:ofi '
                '{install_path}/IMB-MPI1 Pingpong -msglog {msglog_min}:{msglog_max} '
                '-iter {num_iterations} -multi {multi_val} -map {map_args} {additional_args}',
                use_mpi=True)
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
```diff
@@ -38,14 +38,16 @@ class IntelMpiBenchmarks(SpackApplication):
     required_package('intel-mpi-benchmarks')
 
     executable('pingpong',
-               '-genv I_MPI_FABRICS=shm:tcp '
+               '-genv FI_PROVIDER=tcp '
```
douglasjacobsen, line 41: I think we don't want to specify `FI_PROVIDER`, and let modifiers set this for us.

The `tcp` provider is only available in some versions of MPI.

var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
```diff
@@ -38,14 +38,16 @@ class IntelMpiBenchmarks(SpackApplication):
...
+               '-genv I_MPI_FABRICS=shm:ofi '
```
douglasjacobsen, line 42: Maybe the `shm:ofi` portion should be a variable instead of hard-coded here. So we can parametrize it more easily. This would also let us swap out MPI versions more easily.

douglasjacobsen (changes requested):

douglasjacobsen:
I suggested some changed, but I want @rfbgo to review too, since he's the maintainer.

var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
rfbgo: I agree, I think we should probably remove all the `-genv` lines and let the user specify it at the ramble.yaml level 

Realistically there is a good chance these values are default on GCP, and don't make as much sense outside of GCP (or at least TCP networks)

rohitramu: Done!

var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
rohitramu: I've removed all `-genv` arguments as @rfbgo suggested - users will need to set the environment variables themselves.  Please let me know if you prefer a different solution!

douglasjacobsen (approved):

PR #323, opened 2023-11-22
Define zips as a config section

douglasjacobsen:
This merge allows zips to be defined anywhere variables can be defined, including in the scopes supported by any config section.

It also performs some removal of deprecated definitions in the workspace schema.


rfbgo (approved):
LGTM

PR #324, opened 2023-11-27
Tutorials for zips, matrices, variable expansion and indirection, and software stack parameterization

douglasjacobsen:


lib/ramble/docs/tutorials/using_zips_and_matrices.rst
```diff
@@ -0,0 +1,354 @@
...
+   except according to those terms.
+
+.. _zips_and_matrices_tutorial:
+
+==============================
+7) Zips and Matrices
+==============================
+
+In this tutorial, you will learn how to generate a more comprehensive set of
+experiments using zips and matrices for
```
rfbgo, line 16: Should we decouple the purpose of the tutorial (zips etc) from the app? 

```suggestion
experiments using zips. For this tutorial we will use
```

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
```diff
@@ -0,0 +1,354 @@
...
+        batch_submit: '{execute_experiment}'
+        mpi_command: mpirun -n {n_ranks}
+      applications:
+        wrfv4:
+          workloads:
+            CONUS_12km:
+              experiments:
+                scaling_{n_nodes}:
+                  variables:
+                    n_nodes: [1, 2, 4]
```
rfbgo, line 80: Should we add a comment here calling out this is the magic?

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
```diff
@@ -0,0 +1,354 @@
...
+   Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+   https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+   <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+   option. This file may not be copied, modified, or distributed
+   except according to those terms.
+
+.. _zips_and_matrices_tutorial:
+
+==============================
+7) Zips and Matrices
```
rfbgo, line 12: Is this techincally Zips, Vectors, and Matrices? or are you OK with the zip implies vector?

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
```diff
@@ -0,0 +1,354 @@
...
+              ~pnetcdf
+            compiler: gcc9
+        environments:
+          wrfv4:
+            packages:
+            - intel-mpi
+            - wrfv4
+
+The above configuration will execute 3 experiments, comprising a basic scaling
+study on three different sets of nodes.
```
rfbgo, line 100: Is this assuming too much of the user to realize the 3 comes from the vector? should we call it out?

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
```diff
@@ -0,0 +1,354 @@
...
+        - platform
+        - processes_per_node
+      applications:
+        wrfv4:
+          workloads:
+            CONUS_12km:
+              experiments:
+                scaling_{n_nodes}:
+                  variables:
+                    n_nodes: [1, 2, 4]
```
rfbgo, line 160: Does this work? Don't you need to matrix it in to avoid having vecs of len 2 and 3?

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
```diff
@@ -0,0 +1,354 @@
...
+            - intel-mpi
+            - wrfv4
+
+Define an Experiment Matrix
+---------------------------
+
+At this point, your workspace should have a single zip along with one other
+list variable, ``n_nodes``. This configuration will not work properly for two reasons.
+
+The first is that neither the zip, nor ``n_nodes`` are unconsumed. Unconsumed
```
rfbgo, line 185: Do we have a good definition of what consumed means somewhere?

lib/ramble/docs/tutorials/var_expansion_indirection_and_stack_parameterization.rst
```diff
@@ -0,0 +1,495 @@
...
+   except according to those terms.
+
+.. _variable_expansion_and_indirection_and_stack_parameterization_tutorial:
+
+=======================================================================
+8) Variable Expansion, Indirection, and Software Stack Parameterization
+=======================================================================
+
+In this tutorial, you will learn how to use variable expansion, indirection,
+and software stack parameterization when generating experiments for
```
rfbgo, line 16: Similar comment to above. Is this better as `<goal>. We will use <app>`

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
```diff
@@ -0,0 +1,354 @@
...
+
+Zips are explicit groupings of variables, that are combined into a larger
+variable set. A zip is defined by a list of variable definitions, where each
+individual variable is a list / vector variable and all variables are the same
+length. As an example, imagine we had the following variable / zip definitions:
+
+.. code-block:: YAML
+
+    variables:
+      platform: ['platform1', 'platform2']
```
rfbgo, line 121: Is there something other than platform we can have as our second dimension? I like that it is a nice concrete and relatable use case, but I don't like how easy it is for a user to go adapt this and get very different behavior depending on the target spack auto detects

rfbgo (changes requested):
Apologies in advance, there are some really annoying comments here from me 

Overall this is great, and I only have soft suggestions but let me know if you want to discuss anything further offline

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
douglasjacobsen: Vectors are already used in the basic scaling study tutorial. So, this is more specifically `Explicit Zips and Matrices`.

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
douglasjacobsen: No, this is a carry over from one of the previous tutorials. The vector was used to define a scaling study.

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
douglasjacobsen: No, I'm assuming they know this from a previous tutorial already. I can call it out if that's important, but that's not really the purpose of this tutorial.

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
douglasjacobsen: I kind of wanted to lead users into the error case without them expecting it. That way they are forced to see how it breaks and then fix it.

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
douglasjacobsen: It should actually do something right now. The platform is not overly interesting in this case but the platforms have different PPN (so n_ranks should change on the various platforms).

However, I'm open to changing the zip to be something else, I just don't have any great ideas about what it should be.

lib/ramble/docs/tutorials/var_expansion_indirection_and_stack_parameterization.rst
```diff
@@ -0,0 +1,495 @@
...
+            spack_spec: wrf@4.2 build_type=dm+sm compile_type=em_real nesting=basic ~chem
+              ~pnetcdf
+            compiler: gcc9
+        environments:
+          wrfv4:
+            packages:
+            - intel-mpi
+            - wrfv4
+
+The above configuration will execute 6 experiments, comprising a basic scaling
```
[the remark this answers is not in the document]
douglasjacobsen: I think it *will execute* as written (as long as you have 16 cores on the machine you're running on). It won't actually act like they are different platforms though.

lib/ramble/docs/tutorials/var_expansion_indirection_and_stack_parameterization.rst
```diff
@@ -0,0 +1,495 @@
...
+    ramble:
+      variables:
+        processes_per_node: 4
+        n_ranks: '{processes_per_node}*{n_nodes}'
+        batch_submit: '{execute_experiment}'
+        platform: ['platform1', 'platform2']
+        processes_per_node: ['2', '4']
+        mpi_name: ['intel-mpi', 'openmpi']
+        intel-mpi_args: '-n {n_ranks} -ppn {processes_per_node}'
+        openmpi_args: '--np {n_ranks} --map-by ppr:{processes_per_node}:node'
```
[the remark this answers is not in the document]
douglasjacobsen: Sure, I think we can come up with a nicer example, I'm happy to change it. I figured this was a good example of something that could be ugly to do as a zip and really just wanted to show people how they could use this.

lib/ramble/docs/tutorials/using_zips_and_matrices.rst
rfbgo: OK sounds good!

lib/ramble/docs/tutorials/var_expansion_indirection_and_stack_parameterization.rst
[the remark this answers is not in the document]
rfbgo: OK makes sense. As I said, I was a bit torn but at the end of the day I don't have a better example so lets go for this 

I just think this one invites some criticism, but maybe it's at least a good talking point :). If we get round to specializing ramble far enough that it is aware of different MPI implementations we will also have to revisit this, but it's NBD

rfbgo (approved):
LGTM!

PR #325, opened 2023-11-28
Convert GCP modifier to executable_modifier format and collect all ids

rfbgo:
This is techincally a breaking change, as it renames the fom from gid to gids (since it is now plural). This intentionally still does not collect all hostnames from every node

This PR also adds the rm of old files, and now uses `_prepare_analysis` to build a csv of all ids across all nodes

### DIFF
--- a/var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
@@ -24,23 +24,62 @@ class GcpMetadata(BasicModifier):
     mode('standard', description='Standard execution mode')
     default_mode('standard')
 
-    variable_modification('log', '{experiment_run_dir}/gcp-metadata.log', method='set', modes=['standard'])
-    archive_pattern('{log}')
+    executable_modifier('gcp_metadata_exec')
 
-    register_builtin('gcp_metadata_exec')
+    def gcp_metadata_exec(self, executable_name, executable, app_inst=None):
+        from ramble.util.executable import CommandExecutable
+        post_cmds = []
+        pre_cmds = []
 
-    def gcp_metadata_exec(self):
-        machine_type = 'curl -s -w "\\n" "http://metadata.google.internal/computeMetadata/v1/instance/machine-type" -H "Metadata-Flavor: Google" >> {log}'
+        payloads = [
+            # end point, use_mpi
+            ('machine-type', False),
+            ('image', False),
+            ('hostname', False),
+            ('id', True),  # True since we want the gid of every node
+        ]
 
-        image = 'curl -s -w "\\n" "http://metadata.google.internal/computeMetadata/v1/instance/image" -H "Metadata-Flavor: Google" >> {log}'
+        rm_template = []
+        for end_point, _ in payloads:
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
```diff
@@ -24,23 +24,62 @@ class GcpMetadata(BasicModifier):
...
+            ('id', True),  # True since we want the gid of every node
+        ]
 
-        image = 'curl -s -w "\\n" "http://metadata.google.internal/computeMetadata/v1/instance/image" -H "Metadata-Flavor: Google" >> {log}'
+        rm_template = []
+        for end_point, _ in payloads:
+            rm_template.append(f'rm "{{experiment_run_dir}}/gcp-metadata.{end_point}.log"')
 
-        gid = 'curl -s -w "=GID\\n" "http://metadata.google.internal/computeMetadata/v1/instance/id" -H "Metadata-Flavor: Google" >> {log}'
+        pre_cmds.append(
```
douglasjacobsen, line 46: Instead of explicitly removing the templates, could you just change the `output_capture` operator to `>`.

Either is fine, I was more just curious if there's a reason to explicitly `rm` it every time.

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
```diff
@@ -24,23 +24,62 @@ class GcpMetadata(BasicModifier):
     mode('standard', description='Standard execution mode')
     default_mode('standard')
 
-    variable_modification('log', '{experiment_run_dir}/gcp-metadata.log', method='set', modes=['standard'])
-    archive_pattern('{log}')
+    executable_modifier('gcp_metadata_exec')
 
-    register_builtin('gcp_metadata_exec')
+    def gcp_metadata_exec(self, executable_name, executable, app_inst=None):
```
douglasjacobsen, line 29: One odd gotcha with this is you're taking every executable in the experiment and prepending this to it. So, you'll be `rm`-ing and re-creating the same files many times. You might want to add and instance tracker (which is a bit gross) to make sure you only do it once.

Something like:

```
def gcp_metadata_exec(....):
  if not hasattr(self, '_apply_gcp_metadata_exec'):
    self._apply_gcp_metadata_exec = False
    ....
```

douglasjacobsen (changes requested):
Thanks, this looks good. Two small comments.

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
rfbgo: Great catch! I meant to add that, but got carried away in the details. Thanks, will ad it on

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
rfbgo: I agree, good change. Initially I was thinking of supporting other ways to request the data in parallel and was scared that would cause a data race.. but honestly I don't think that's possible. Will update

douglasjacobsen (approved):

PR #326, opened 2023-11-29
Add three tutorials and cleanup tutorial content

douglasjacobsen:
This merge adds three more tutorials, for using modifiers, internals, and success criteria.

Additionally, it changes the tutorial naming scheme to include the numerical order of tutorials, and extracts shared portions of the tutorials to reduce copy-paste text..

PR #327, opened 2023-11-30
Allow custom executables to be injected relative to other executables

douglasjacobsen:
This merge adds functionality, tests, and documentation for controlling executable order through injection rather than only the explicit order list that was previously supported.

This defines a new block in the `internals` section, named `exectuable_injection`. This block is a list of executables to inject into an experiment. The format of an executable in this list is:

```
- name: <executable_name>
  order: <before / after>
  [realtive_to: <relative_executable_name>
```

The default `order` is `after`. `relative_to` is an optional attribute which controls where the new executable will be injected in the list. If it is omitted, the `before` and `after` orders apply to the entire list.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -15,6 +15,7 @@ import textwrap
 import string
 import shutil
 import fnmatch
+import enum
 from typing import List
 
 import llnl.util.filesystem as fs
@@ -666,7 +667,7 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
         # Perform executable injection
         if namespace.executable_injection in self.internals:
 
-            supported_orders = set(['before', 'after'])
+            supported_orders = enum.Enum('supported_orders', ['before', 'after'])
 
             # Order can be 'before' or 'after.
             # If `relative_to` is not set, then before adds to be the beginning of the list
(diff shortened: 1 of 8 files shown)


douglasjacobsen:
This draft will be ready after #326 is merged.

lib/ramble/ramble/application.py
```diff
@@ -650,13 +656,84 @@ def _get_executables(self):
...
             for name, conf in self.internals[namespace.custom_executables].items():
                 self.executables[name] = ramble.util.executable.CommandExecutable(
                     name=name,
                     **conf
                 )
 
+        # Perform executable injection
+        if namespace.executable_injection in self.internals:
+
+            supported_orders = set(['before', 'after'])
```
rfbgo, line 669: Should we DRY this out into a keyword/enum? These magic strings are already in a few places and previous experience says they might only spread :P

lib/ramble/ramble/application.py
```diff
@@ -650,13 +656,84 @@ def _get_executables(self):
...
+                if relative is None:
+                    if injection_order == 'before':
+                        executables.insert(0, exec_name)
+                    elif injection_order == 'after':
+                        executables.append(exec_name)
+                else:
+
+                    found = False
+                    if injection_order == 'before':
+                        relative_index = 0
```
rfbgo, line 717: nit: all this relative index rejiggering doesn't feel very pythonic.. but I am struggling to come up with a definitely better solution

I'm kind of surprised you did not turn this into a more graph-style solution where you walk it and just append the pre/node/post and went right to the 1d splat of that.. unless you have some awesome solution lets leave this as-is for now?

rfbgo (approved):
LGTM! 

One soft suggestion/request, one ignorable passing comment :)

lib/ramble/ramble/application.py
douglasjacobsen: There are lots of places a graph would be nice for. Maybe we should add that to the task list and use it to clean things up in the future.

lib/ramble/ramble/application.py
douglasjacobsen: This is actually the only place these magic strings show up. For the chain we use `before` and `after`, but the strings are `before_root` `after_root` etc...

Most other places, we use append and prepend, but those are used to control position relative to an entire list rather than within the list.

rfbgo: For that comment specifically I was thinking about the local use here, eg: before/after in

```
 if relative is None:
                    if injection_order == 'before':
                        executables.insert(0, exec_name)
                    elif injection_order == 'after':
                        executables.append(exec_name)
                else:

                    found = False
                    if injection_order == 'before':
```

Long term I was indeed thinking about the stuff you mentioned, but agree its OK for now

rfbgo (approved):

lib/ramble/ramble/application.py
douglasjacobsen: OK, thanks. I pushed an update. Let me know if that looks better.

rfbgo (approved):

PR #338, opened 2023-12-07
Add `experiment_index` as a defined variable

douglasjacobsen:
This merge adds `experiment_index` to every rendered experiment. This can then be used within `--where` clauses to control which experiments will be actioned within pipelines.


rfbgo (approved):
Awesome and clean feature. This will be really nice for supporting things like GKE and batch

PR #349, opened 2023-12-13
Allow for empty spack env

rfbgo:
Follow on from https://github.com/GoogleCloudPlatform/ramble/pull/336 which I think did not do what was intended 

I want to be able to support something like:
```
 42     environments:
 43       app_name: {}
```

This gives a `len(self.env_contents) == 0` so is currently considered invalid.

This PR instead lets it continue and hash an empty dict

### DIFF
--- a/lib/ramble/ramble/experimental/uploader.py
+++ b/lib/ramble/ramble/experimental/uploader.py
@@ -129,11 +129,7 @@ def upload_results(results):
                 logger.die('No upload URI (config:upload:uri) in config.')
 
             logger.msg(f'Uploading Results to {uri}')
-
-            if len(formatted_data) == 0:
-                logger.warn('No data to upload')
-            else:
-                uploader.perform_upload(uri, formatted_data)
+            uploader.perform_upload(uri, formatted_data)
         else:
             raise ConfigError("Unknown config:upload:type value")
 
--- a/lib/ramble/ramble/experimental/uploader.py
+++ b/lib/ramble/ramble/experimental/uploader.py
@@ -129,7 +129,11 @@ def upload_results(results):
                 logger.die('No upload URI (config:upload:uri) in config.')
 
             logger.msg(f'Uploading Results to {uri}')
-            uploader.perform_upload(uri, formatted_data)
+
+            if len(formatted_data) == 0:
+                logger.warn('No data to upload')
+            else:
+                uploader.perform_upload(uri, formatted_data)
         else:
             raise ConfigError("Unknown config:upload:type value")
 
--- a/lib/ramble/ramble/spack_runner.py
+++ b/lib/ramble/ramble/spack_runner.py
@@ -602,20 +602,17 @@ class SpackRunner(object):
         contents_to_hash = None
 
         if not self.dry_run:
-            if not lock_exists and require_exist and len(self.env_contents) > 0:
-                logger.die(
-                    'spack.lock file does not exist in environment '
-                    f'{self.env_path}\n'
-                    'Make sure your workspace is fully setup with\n'
-                    '    ramble workspace setup'
-                )
-            elif len(self.env_contents) == 0:
-                contents_to_hash = {}
-            elif lock_exists:
+            if not lock_exists:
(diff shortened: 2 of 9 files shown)


lib/ramble/ramble/spack_runner.py
```diff
@@ -602,16 +602,20 @@ def inventory_hash(self, require_exist=False):
...
         if not self.dry_run:
-            if not lock_exists and (require_exist or len(self.env_contents) == 0):
+            if not lock_exists and require_exist and len(self.env_contents) > 0:
                 logger.die(
                     'spack.lock file does not exist in environment '
                     f'{self.env_path}\n'
                     'Make sure your workspace is fully setup with\n'
                     '    ramble workspace setup'
                 )
+            elif len(self.env_contents) == 0:
```
douglasjacobsen, line 612: Should this `elif` either:
1. Be nested under a `not lock_exists` test
or
2. Be below (but not nested under) the `elif lock_exists` test?

To make sure if a lock file exists, and no packages are defined we follow the logic for the existing lock file

lib/ramble/ramble/spack_runner.py
```diff
@@ -602,16 +602,20 @@ def inventory_hash(self, require_exist=False):
...
+                contents_to_hash = {}
             elif lock_exists:
                 with open(spack_file, 'r') as f:
                     contents_to_hash = sjson.load(f)
+            else:
```
douglasjacobsen, line 617: I don't think we want an error here. I can't remember exactly why, but I think we want to support falling through this in the case that the lock file doesn't exist, but we also don't require it.

Or, maybe this should just be a return in that case, instead of an error?

lib/ramble/ramble/spack_runner.py
rfbgo: Looking at this again, I think this is actually not needed and this situation can be handled by:

```
617         if self.dry_run or not lock_exists:
```

lib/ramble/ramble/spack_runner.py
rfbgo: Will update

lib/ramble/ramble/experimental/uploader.py
```diff
@@ -129,7 +129,11 @@ def upload_results(results):
                 logger.die('No upload URI (config:upload:uri) in config.')
 
             logger.msg(f'Uploading Results to {uri}')
-            uploader.perform_upload(uri, formatted_data)
+
+            if len(formatted_data) == 0:
+                logger.warn('No data to upload')
+            else:
+                uploader.perform_upload(uri, formatted_data)
```
douglasjacobsen, line 136: This seems like it shouldn't be part of this PR (and the rest of the PR could just be rebased if you want this change)

lib/ramble/ramble/spack_runner.py
```diff
@@ -602,20 +602,17 @@ def inventory_hash(self, require_exist=False):
...
-                    'spack.lock file does not exist in environment '
-                    f'{self.env_path}\n'
-                    'Make sure your workspace is fully setup with\n'
-                    '    ramble workspace setup'
-                )
-            elif len(self.env_contents) == 0:
-                contents_to_hash = {}
-            elif lock_exists:
+            if not lock_exists:
+                if require_exist and len(self.env_contents) > 0:
```
douglasjacobsen, line 606: This LGTM

douglasjacobsen (approved):

lib/ramble/ramble/experimental/uploader.py
rfbgo: Yea this got picked up accidentally, will remove (but it's actually a 0 delta from HEAD)

lib/ramble/ramble/spack_runner.py
rfbgo: As discussed offline I will also add a `status=SETUP` to have a more explicit semantic around if something is setup

douglasjacobsen (approved):

PR #355, opened 2023-12-18
Convert ramble root to existing variable

douglasjacobsen:
This merge changes the tutorials from `$ramble_root` to `$RAMBLE_ROOT` which is a variable that exists.

### DIFF
--- a/lib/ramble/docs/tutorials/1_hello_world.rst
+++ b/lib/ramble/docs/tutorials/1_hello_world.rst
@@ -112,7 +112,7 @@ This will create a named workspace for you in:
 
 .. code-block:: console
 
-    $ $ramble_root/var/ramble/workspaces/hello_world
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/expander.py
```diff
@@ -613,6 +613,24 @@ def _eval_function_call(self, node):
 
         if node.func.id == 'range':
             return list(range(*args, **kwargs))
+        elif node.func.id == 'pad_lhs':
```
rfbgo, line 616: Is this meant to be in this PR?

rfbgo (changes requested):
Looks good to me, but I think we picked up an unwanted change?

lib/ramble/ramble/expander.py
douglasjacobsen: Nope! Good catch.

rfbgo (approved):

Issue #361, opened 2023-12-19
Better error message for missing spack spec definitions


becker33:
### Summary

Currently, the error message for an undefined spack definition "handle" is a `KeyError` and does not describe which handles are missing. We would like an error message that clearly lists which handles are undefined.

### General information

- [X] I have run `ramble --version` and reported the version of Ramble
- [X] I have searched the issues of this repo and believe this is not a duplicate

becker33:
Closed by #362

PR #362, opened 2023-12-20
Detect undefined packages and print a useful error message

douglasjacobsen:
This merge updates the error message when undefined packages are used within software environments.

### DIFF
--- a/lib/ramble/ramble/software_environments.py
+++ b/lib/ramble/ramble/software_environments.py
@@ -206,11 +206,16 @@ class SoftwareEnvironments(object):
                             env_packages.append(expanded_pkg)
 
                         pkgs_with_compiler = []
+                        missing_pkgs = set()
                         for pkg_name in env_packages:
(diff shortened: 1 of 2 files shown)


douglasjacobsen:
Previously, this was a KeyError that didn't really tell you what to do to fix the issue.

Fixes #361

rfbgo (approved):

becker33:
Thank you! This will allow us to take a ramble.yaml that is designed to be system agnostic and use the error message to generate a list of "handles" that have to be defined in system-specific configs to be able to work with it.

```
$ ramble workspace create tmp /path/to/ramble.yaml
$ ramble workspace setup
< get error message > 
$ $EDITOR /path/to/system/config/spack.yaml
# write definitions for handles listed in error message
$ ramble config edit
# add an include for /path/to/system/config/spack.yaml
$ ramble workspace setup
# success
```

PR #373, opened 2024-01-16
Add `expandable` attribute to variable definitions

douglasjacobsen:
This merge adds the ability for variables to say they cannot be expanded. When variables are defined as `expandable=False` the expander replaces references to the variable with their exact value.

This is also used in the Gromacs application definition on the `size` variable.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -85,7 +85,8 @@ class ExpansionNode(object):
             self.children.append(children)
 
     def define_value(self, expansion_dict, allow_passthrough=True,
-                     expansion_func=str, evaluation_func=eval):
+                     expansion_func=str, evaluation_func=eval,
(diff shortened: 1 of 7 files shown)


rfbgo (approved):

lib/ramble/ramble/expander.py
```diff
@@ -130,6 +130,8 @@ def define_value(self, expansion_dict, allow_passthrough=True,
                 kw_parts = format_kw.split(':')
                 required_passthrough = False
 
+                print(kw_parts)
```
rfbgo, line 133: Remove?

rfbgo (changes requested):

lib/ramble/ramble/expander.py
douglasjacobsen: Good catch. Thanks

rfbgo (approved):

PR #375, opened 2024-01-16
Add repeated experiments with summary statistics

dapomeroy:
Adds the ability to specify 'n_repeats' for any or all experiments, which will cause those experiments to be repeated n times and generate summary statistics during analysis.


lib/ramble/ramble/application.py
```diff
@@ -325,6 +329,18 @@ def set_template(self, is_template):
         """Set if this instance is a template or not"""
         self.is_template = is_template
 
+    def set_repeats(self, repeats):
+        """Set if this instance will use repeats
+
+        repeats = (is_repeat_base, n_repeats)
+
+        is_repeat_base (Bool): True if this is the base experiment of a repeat set,
```
douglasjacobsen, line 337: This doesn't look like it's an argument anymore.

lib/ramble/docs/configuration_files.rst
```diff
@@ -561,3 +564,39 @@ format is as follows:
...
+
+Repeats can be defined at the Ramble config level, which applies to all
+experiments, or at the application, workload, or experiment levels. When repeats
+are enabled, each experiment will be replicated n times and individually executed.
+Summary statistics will be calculated for each set of repeats when experiment results
+are analyzed.
+
+By default, a set of repeats is only successful if all individual experiments succeed.
+If ``repeats_success_strict`` is set to ``False``, Ramble will ignore failed experiments
+and calculate statistics for successful experiments only.
```
douglasjacobsen, line 602: It also marks t he repeat set as a success in this case, right?

lib/ramble/ramble/experiment_set.py
```diff
@@ -83,11 +85,13 @@ def read_config_vars(self, workspace):
         global_context.context_name = self._contexts.global_conf.name
         global_context.variables = self.get_config_vars(workspace)
         global_context.env_variables = self.get_config_env_vars(workspace)
+        global_context.n_repeats = ramble.config.get('config:n_repeats')
         self._set_context(self._contexts.global_conf,
                           global_context)
 
     def get_config_vars(self, workspace):
         conf = ramble.config.config.get_config('config')
+        print(f'config conf = {conf}')
```
douglasjacobsen, line 94: Remove

lib/ramble/ramble/util/stats.py
```diff
@@ -0,0 +1,141 @@
...
+
+    max = 0
+    for val in list:
+        if decimal_places(val) > max:
+            max = decimal_places(val)
+    return max
+
+
+class StatsBase(object):
+    # def __init__(self, name):
```
douglasjacobsen, line 33: You can probably remove these comments in all of the places in this file.

lib/ramble/ramble/util/stats.py
```diff
@@ -0,0 +1,141 @@
...
+    """Returns the max decimal places of a list of values"""
+
+    max = 0
+    for val in list:
+        if decimal_places(val) > max:
+            max = decimal_places(val)
+    return max
+
+
+class StatsBase(object):
```
douglasjacobsen, line 32: Given the stats classes are so self contained, this PR should probably have some unit tests for them.

douglasjacobsen (changes requested):
This is great. Thanks for working on this!

lib/ramble/ramble/util/stats.py
dapomeroy: Done

lib/ramble/ramble/experiment_set.py
dapomeroy: Done. Thanks for catching this.

lib/ramble/docs/configuration_files.rst
dapomeroy: Yes, though I just double checked and found a mistake in the success criteria code, which is now fixed. Updated the verbiage here to be clear as well.

lib/ramble/ramble/application.py
dapomeroy: Can you explain what you mean?

lib/ramble/docs/configuration_files.rst
```diff
@@ -561,3 +564,39 @@ format is as follows:
...
+          workloads:
+            <workload name>:
+              n_repeats: 'int'
+              experiments:
+                <experiment name>:
+                  n_repeats: 'int'
+
+Repeats can be defined at the Ramble config level, which applies to all
+experiments, or at the application, workload, or experiment levels. When repeats
+are enabled, each experiment will be replicated n times and individually executed.
```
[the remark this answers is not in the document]
dapomeroy: Done

lib/ramble/docs/configuration_files.rst
```diff
@@ -561,3 +564,39 @@ format is as follows:
...
+The repeats config is used to define the number of times each experiment will be
+repeated. Its format is as follows:
+
+.. code-block:: yaml
+
+    ramble:
+      config:
+        n_repeats: 'int'
+        repeats_success_strict: [True/False]
+      applications:
```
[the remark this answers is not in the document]
dapomeroy: Removed this and simplified the section at the bottom to just cover config and link to the workspace config file.

lib/ramble/docs/configuration_files.rst
```diff
@@ -561,3 +564,39 @@ format is as follows:
...
+          n_repeats: 'int'
+          workloads:
+            <workload name>:
+              n_repeats: 'int'
+              experiments:
+                <experiment name>:
+                  n_repeats: 'int'
+
+Repeats can be defined at the Ramble config level, which applies to all
+experiments, or at the application, workload, or experiment levels. When repeats
```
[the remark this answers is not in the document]
dapomeroy: Done

lib/ramble/docs/configuration_files.rst
```diff
@@ -561,3 +564,39 @@ format is as follows:
...
 another fully qualified experiment (through the ``cross_ref_var`` syntax).
+
+
+.. _repeats-config:
+
+------------------
+Repeats Section:
+------------------
+
+The repeats config is used to define the number of times each experiment will be
```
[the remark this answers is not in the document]
dapomeroy: Done

lib/ramble/ramble/util/stats.py
dapomeroy: Added

lib/ramble/ramble/application.py
douglasjacobsen: Oh sorry, I misunderstood t his comment. You mean the input argument is a tuple, and the tuple has is_repate_base / n_repeats as the two values.

I thought this was describing the arguments to the method.

Can you update the doc string here to explain that a bit more (could just be with plain English) for the next reader?

dapomeroy: That's right, input is a tuple. I've reworded to make it more clear.

lib/ramble/docs/configuration_files.rst
```diff
@@ -561,3 +561,21 @@ format is as follows:
...
+Experiment Repeats:
+-------------------
+
+The repeats config options within the ``config`` section are used to define a number
+of times each experiment will be repeated. Summary statistics will be calculated for
+the set of repeats. Its format is as follows:
+
+.. code-block:: yaml
+
+    ramble:
```
douglasjacobsen, line 575: nit; Can you remove `ramble;` here? In the config section it's intended to start with `config:` while in the workspace config, you'd use `ramble:config` instead.

lib/ramble/docs/configuration_files.rst
```diff
@@ -561,3 +561,21 @@ format is as follows:
...
+      config:
+        n_repeats: 'int'
+        repeats_success_strict: [True/False]
+
+More information on experiment repeats is found in the :ref:`workspace configuration
```
douglasjacobsen, line 580: Maybe "More information on experiment repeats is found" --> "More information on using repeats within a workspace can be found"

lib/ramble/ramble/application.py
douglasjacobsen: nit; Maybe (to make this more consistent with other argument docs):

"""
repeats (tuple): consists of:
is_repeat_base( Bool): ...
num_repeats (Int): ...
"""

lib/ramble/ramble/test/util/stats.py
```diff
@@ -0,0 +1,51 @@
...
+# except according to those terms.
+
+import ramble.util.stats
+
+
+def test_stats_for_repeat_foms():
+    test_values = [-2, 0, 2, 5.5]
+    test_units = 's'
+
+    expected_output = [(-2, 's', 'summary::min'),
```
douglasjacobsen, line 16: I'd prefer if this used `pytest.mark.parametrize` over an internal loop.

lib/ramble/ramble/test/util/stats.py
```diff
@@ -0,0 +1,51 @@
...
+
+    for expected_calc in expected_output:
+        assert expected_calc in calcs
+
+
+def test_stats_for_single_fom_value():
+    test_values = [3]
+    test_units = 's'
+
+    expected_output = [(3, 's', 'summary::min'),
```
douglasjacobsen, line 37: Same parameterize stuff here too.

lib/ramble/ramble/util/stats.py
douglasjacobsen: Awesome. Thanks!

douglasjacobsen (changes requested):
Mostly nits now. :) I'll test this next.

lib/ramble/ramble/schema/applications.py
```diff
@@ -86,6 +86,11 @@
     'additionalProperties': False
 }
 
+repeats_def = {
+    'type': 'string',
```
douglasjacobsen, line 90: I think either:
1) this should use the string_or_num from `schema/types.py`
or;
2) This should be `'type': 'number'` with a default value of 0.

It's a bit awkward to require it to be `'2'` instead of `2`.

douglasjacobsen (changes requested):

lib/ramble/docs/configuration_files.rst
dapomeroy: Done

lib/ramble/ramble/test/util/stats.py
dapomeroy: I've rewritten this to test the stats classes directly with parametrize, and cut out the loops and the variations that aren't testing a unique result. Please let me know if this is what you were thinking and if you have any suggestions.

lib/ramble/ramble/schema/applications.py
dapomeroy: Thank you, fixed this to be string_or_num

lib/ramble/ramble/pipeline.py
```diff
@@ -76,7 +76,8 @@ def __init__(self, workspace, filters):
     def _construct_hash(self):
         """Hash all of the experiments, construct workspace inventory"""
         for exp, app_inst, _ in self._experiment_set.all_experiments():
-            if self.require_inventory:
+            if self.require_inventory and not (app_inst.is_template or app_inst.repeats[0]):
+                print(f'checking status of {exp}. Status is {app_inst.get_status()}')
```
douglasjacobsen, line 80: This should be removed.

lib/ramble/ramble/application.py
```diff
@@ -1419,9 +1443,168 @@ def format_context(context_match, context_format):
...
+                    exp_inst = self.experiment_set.chained_experiments[exp]
+                else:
+                    continue
+
+                # When strict success is off for repeats (loose success), skip failed exps
+                if (not workspace.repeat_success_strict
+                    and exp_inst.results['RAMBLE_STATUS'] == 'FAILED'):
+                    continue
+
+                for context in exp_inst.results['CONTEXTS']:
```
douglasjacobsen, line 1567: If a repeat experiment is missing the expected context (or if it doesn't have any results at all) this errors.

It should probably be guarded with something like `if 'CONTEXT' in exp_inst.results:` to make sure it has results to try and compute stats with.

And maybe the entire contents of this for loop could be guarded by this to prevent trying to iterate over results when they don't exist.

douglasjacobsen (changes requested):

lib/ramble/ramble/pipeline.py
dapomeroy: Done

lib/ramble/ramble/util/stats.py
```diff
@@ -0,0 +1,124 @@
...
+            new_units = units
+        except statistics.StatisticsError as e:
+            var = str(e)
+            new_units = ''
+
+        return (var, new_units, f'summary::{self.name}')
+
+
+class StatsCountValues(StatsBase):
+    name = 'n_repeats'
```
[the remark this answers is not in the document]
dapomeroy: Good suggestion, updated

douglasjacobsen (approved):

lib/ramble/ramble/application.py
```diff
@@ -1426,9 +1449,170 @@ def format_context(context_match, context_format):
...
+                exp_inst = self.experiment_set.experiments[exp]
+            elif exp in self.experiment_set.chained_experiments.keys():
+                exp_inst = self.experiment_set.chained_experiments[exp]
+            else:
+                continue
+
+            exp_success.append(exp_inst.get_status())
+
+        if workspace.repeat_success_strict:
+            if 'FAILED' in exp_success:
```
rfbgo, line 1531: Should this use `experiment_status` ?

lib/ramble/ramble/application.py
```diff
@@ -1426,9 +1449,170 @@ def format_context(context_match, context_format):
...
+            repeat_experiments[repeat_exp_namespace]['base_exp'] = base_exp_namespace
+            if n == 1:
+                first_repeat_exp = repeat_exp_namespace
+
+        # Create initial results dict since analysis pipeline was skipped for base exp
+        self.results = {'name': base_exp_namespace}
+        self.results['N_REPEATS'] = n_repeats
+        self.results['EXPERIMENT_CHAIN'] = self.chain_order.copy()
+
+        #
```
rfbgo, line 1512: Do we need these empty line comments around the blocks?

lib/ramble/ramble/application.py
```diff
@@ -1426,9 +1449,170 @@ def format_context(context_match, context_format):
                     fom_copy['name'] = fom_name
                     context_map['foms'].append(fom_copy)
 
-                results['CONTEXTS'].append(context_map)
+                self.results['CONTEXTS'].append(context_map)
```
rfbgo, line 1452: It would be nice to dry out the repeated use of the magic string `'CONTEXTS'`

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -355,6 +355,12 @@ def workspace_setup_setup_parser(subparser):
              'all scripts. Prints commands that would be executed ' +
              'for installation, and files that would be downloaded.')
 
+    subparser.add_argument(
+        '--n_repeats', dest='n_repeats',
+        nargs='+',
```
rfbgo, line 360: Why is this + not 1? What happens if we give multiple args here?

lib/ramble/ramble/pipeline.py
```diff
@@ -238,6 +238,12 @@ def _prepare(self):
         super()._prepare()
 
     def _complete(self):
+        # Calculate statistics for repeats and inject into base experiment results
+        for _, app_inst, _ in self._experiment_set.filtered_experiments(self.filters):
+
+            if app_inst.repeats[0] and app_inst.repeats[1] > 0:
```
rfbgo, line 244: Is there cleaner way to do this? I assume this semantic is essentially `if has_repeats` but it's not super obvious at a glance

lib/ramble/ramble/application.py
dapomeroy: It does. ``exp_success`` is a list of the ``experiment_status`` for every repeat in the set. It then applies the strict/loose criteria to determine the status of the repeat set. If strict, any failure causes the set to fail. If not strict, any success causes the set to succeed.

lib/ramble/ramble/cmd/workspace.py
dapomeroy: That's a mistake, it should only accept 1 arg.

lib/ramble/ramble/application.py
```diff
@@ -478,6 +494,9 @@ def run_phase(self, phase, workspace):
         if self.is_template:
             logger.debug(f'{self.name} is a template. Skipping phases')
             return
+        if self.repeats[0]:
```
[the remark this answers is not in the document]
dapomeroy: If we made it a class, where do you think that ought to live? New file, or would it make sense to put the class within context.py?

lib/ramble/ramble/application.py
dapomeroy: Some of the other strings here are also repeated in multiple modules (e.g., 'RAMBLE_STATUS', 'RAMBLE_VARIABLES'). Ok if I create a task to come back and address all rather than just doing contexts now?

rfbgo: Yep, great idea

lib/ramble/ramble/application.py
rfbgo: Sorry I mean more "it looks like you are using a magic string that I think we have in an enum. Are they the same and if so can we use the enum?"

lib/ramble/ramble/application.py
```diff
@@ -1426,9 +1449,170 @@ def format_context(context_match, context_format):
...
+
+        repeat_experiments = {}
+        repeat_foms = {}
+        first_repeat_exp = ''
+
+        # repeat_experiments dict {repeat experiment namespace: {dict}}
+        # repeat_foms dict {context: {(fom name, units, origin, origin_type): [list of values]}}
+        # origin_type is generated as 'summary::stat_name'
+
+        # Check if this is the base experiment of a set of repeats
```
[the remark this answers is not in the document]
dapomeroy: Done

lib/ramble/ramble/application.py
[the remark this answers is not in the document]
dapomeroy: Added a new class in a new file

lib/ramble/ramble/application.py
dapomeroy: Got it. Since the use of get_status() above creates a list of enum names as strings, I've updated it to check for the enum name. Does this look right?
```
if experiment_status.FAILED.name in exp_success:
```

lib/ramble/ramble/pipeline.py
dapomeroy: This is more readable now with the repeats class

lib/ramble/ramble/application.py
```diff
@@ -126,6 +129,7 @@ def copy(self):
         new_copy.set_variables(self.variables.copy(), self.experiment_set)
         new_copy.set_internals(self.internals.copy())
         new_copy.set_template(False)
+        new_copy.set_repeats((False, 0))
```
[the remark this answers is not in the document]
dapomeroy: This copy method is used for creating the copies of an experiment that get inserted as chained experiments, so repeats should be set to 'off'. I've changed this now to have it run 'repeats.set_repeats()' with no args, which uses the default values of (False, 0), so that it's not hardcoded in multiple places

lib/ramble/ramble/application.py
```diff
@@ -1419,9 +1443,168 @@ def format_context(context_match, context_format):
...
+        if not self.repeats[0]:
+            return
+        n_repeats = self.repeats[1]
+
+        base_exp_name = self.expander.experiment_name
+        base_exp_namespace = self.expander.experiment_namespace
+
+        # Create a list of all repeats by inserting repeat index
+        for n in range(1, n_repeats + 1):
+            if '.chain' in base_exp_name:
```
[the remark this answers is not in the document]
dapomeroy: Changed this to look at membership in the experiment_set.chained_experiments and experiment_set.experiments instead of parsing the name. If it's in the chained experiment list, and NOT in the experiment list, it will have the '.chain' in the name. For non-chained experiments, the ``base_exp_name`` is found in both lists (as part of the name of chained experiments)

rfbgo (approved):

lib/ramble/ramble/application.py
rfbgo: Yea, I think so

lib/ramble/ramble/application.py
```diff
@@ -359,6 +363,18 @@ def set_template(self, is_template):
         """Set if this instance is a template or not"""
         self.is_template = is_template
 
+    def set_repeats(self, repeats):
```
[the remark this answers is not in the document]
dapomeroy: Discussed offline. To recap, this was modeled off 'set_template()' which has a setter but no getter. I don't have any other reason for it besides keeping it simple.

lib/ramble/ramble/repeats.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __init__(self):
+        """Constructor for a Repeats object
+
+        Creates a Repeats object with default setting of 'no repeats'."""
+        self.n_repeats = 0
+        self.is_repeat_base = False
+        self.repeat_index = None
+
+    def set_repeats(self, is_repeat_base=False, n_repeats=0):
```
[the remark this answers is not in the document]
dapomeroy: Explicit sounds good. I've removed the default

rfbgo (approved):

PR #386, opened 2024-01-25
Allow Ramble Upload to track the calling user

rfbgo:
Note this is a SQL schema breaking change

### DIFF
--- a/lib/ramble/ramble/experimental/uploader.py
+++ b/lib/ramble/ramble/experimental/uploader.py
@@ -30,13 +30,15 @@ class Uploader():
                 f"{self.__class__} requires %{data} argument.")
         pass
 
+
 def get_user():
-    import getpass
-    user = getpass.getuser()
     config_user = ramble.config.get('config:user')
     if config_user:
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/experimental/uploader.py
```diff
@@ -30,6 +30,13 @@ def perform_upload(self, uri, data):
                 f"{self.__class__} requires %{data} argument.")
         pass
 
+def get_user():
```
douglasjacobsen, line 33: extreme nit;

can we only call getpass if config_user is not set, to avoid making unnecessary system calls?

douglasjacobsen (changes requested):
I don't have any concerns with this, but the docs should probably be updated around this config option as well.

lib/ramble/ramble/experimental/uploader.py
rfbgo: Done!

douglasjacobsen (approved):

PR #406, opened 2024-02-12
Suppress printing non-actionable experiments in pipeline output

douglasjacobsen:
This merge prevents experiments like tempaltes and repeat bases from being printed in the output of pipeline actions.


dapomeroy:
When repeats are enabled, the analyze pipeline is skipping the repeat base experiment and the summary statistics are not being generated

dapomeroy (approved):
LGTM

PR #419, opened 2024-02-24
Add support for attribute graphs, apply to phases and executables

douglasjacobsen:
This merge adds a new set of classes to help manage / manipulate graphs. This is based off of graphlib (so graphlib-backport is added to the requirements.txt).

This graph is used for ordering phases when they are registered, and command executables / builtins within experiments.

### DIFF
--- a/lib/ramble/ramble/graphs.py
+++ b/lib/ramble/ramble/graphs.py
@@ -1,0 +1,424 @@
+# Copyright 2022-2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import enum
+import graphlib
+
+import ramble.error
+import ramble.util.graph
+
+from ramble.util.logger import logger
+
+
+class AttributeGraph(object):
+
+    node_type = 'object'
+
+    def __init__(self, obj_inst):
+        self._obj_inst = obj_inst
+        self.node_definitions = {}
+        self.adj_list = {}
+        self._prepared = False
+        self._sorted = None
+
+    def _make_editable(self):
+        """Make this graph editable, and remove any defined ordering"""
+        if self._prepared:
+            del self._sorted
+            self._sorted = None
+            self._prepared = False
+
+    def update_graph(self, node, dep_nodes=[], internal_order=False):
+        """Update the graph with a new node and / or new dependencies.
+
+        Given a node, and list of dependencies, define new edges in the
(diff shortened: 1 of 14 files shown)


lib/ramble/ramble/graphs.py
```diff
@@ -0,0 +1,424 @@
...
+        self._obj_inst = obj_inst
+        self.node_definitions = {}
+        self.adj_list = {}
+        self._prepared = False
+        self._sorted = None
+
+    def _make_editable(self):
+        """Make this graph editable, and remove any defined ordering"""
+        if self._prepared:
+            del self._sorted
```
rfbgo: Do you need to do the `del` if you assign to `None`? I guess it forced the GC to happen sooner?

lib/ramble/ramble/graphs.py
```diff
@@ -0,0 +1,424 @@
...
+            if relative_node not in cur_exec_order:
+                relative_error = True
+
+            if relative_error:
+                logger.die('In experiment '
+                           f'"{exp_name}" '
+                           f'attempting to inject executable "{exec_name}" '
+                           f'relative to a non existing executable "{relative}".')
+
+        if relative is not None and relative not in self.node_definitions:
```
rfbgo: Any reason not to nest this in the above conditional? I'm OK either way, just surprised to see the same (half) if twice

lib/ramble/ramble/graphs.py
```diff
@@ -0,0 +1,424 @@
...
+            relative_node = self.node_definitions[relative]
+            if relative_node not in cur_exec_order:
+                logger.die('In experiment '
+                           f'"{exp_name}" '
+                           f'attempting to inject executable "{exec_name}" '
+                           f'relative to a non existing executable "{relative}".')
+
+        # If relative is none, determine head and tail nodes to inject properly
+        if relative is None:
```
rfbgo: Given the above `if is not None` this feels like it should be an else, but given this also has an else it makes me wonder if we can tidy this flow up now we have it all written

rfbgo (changes requested):
Few nits, otherwise looks great!

lib/ramble/ramble/graphs.py
douglasjacobsen: You don't need to, so I could definitely just set it to None and let it get cleaned automatically.

lib/ramble/ramble/graphs.py
douglasjacobsen: Ah sure, I can clean that up.

lib/ramble/ramble/graphs.py
douglasjacobsen: I actually think this is dead code. And looking at the assignment below it, it should always error since we key off of relative in node_definitions, but we already checked that it wasn't in node_definitions. (And this exact error case is already handled above).

So, I'll just remove it.

rfbgo (approved):

PR #431, opened 2024-03-06
Fix propagation of non expandable variables from modifiers

douglasjacobsen:
This merge fixes an issue where non-expandable variables defined within a modifier were not properly propagated into the expander objects.

A test is added to capture this in the future as well.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -287,6 +287,7 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
             for var, conf in self.workload_variables[workload_name].items():
                 if 'expandable' in conf and not conf['expandable']:
                     self.no_expand_vars.add(var)
+
         self.expander.set_no_expand_vars(self.no_expand_vars)
 
     def set_internals(self, internals):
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/application.py
```diff
@@ -702,6 +703,12 @@ def build_modifier_instances(self):
         # (note: the base ramble variables are checked earlier too)
         self.keywords.check_required_keys(self.variables)
 
+        # Ensure no expand vars are set correctly for modifiers
+        for mod_inst in self._modifier_instances:
+            for var in mod_inst.no_expand_vars():
+                self.expander.add_no_expand_var(var)
```
dapomeroy, line 709: Is L709 adding the modifier's no expand var to the application's list of no expand vars? Could this potentially cause an app var to fail to expand if it has the same name as a mod var?

douglasjacobsen: Modifier variable definitions are "inherited" by application instances.

So, in general the application should contain the union of its list of non-expandable variables, and every non-expandable variable from it's modifiers.

Each modifier should only include the union of its own non-expandable variables and the applications (but only the ones defined within the application explicitly, not all of the other modifier's)

dapomeroy (approved):
LGTM

PR #433, opened 2024-03-06
Convert environment generation logic to lazy / implcit instead of explicit

douglasjacobsen:
This merge converts the logic around generating software environments and packages away from being explicitly defined in the workspace configuration file.

Instead, packages and environments are now generated lazily when processing experiments. This avoids several issues related to vector / matrix logic causing problems rendering individual packages and environments.

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -542,6 +542,8 @@ def workspace_info(args):
     # Build experiment set
     experiment_set = ws.build_experiment_set()
 
+    software_environments = ramble.software_environments.SoftwareEnvironments(ws)
+
     if args.tags:
         color.cprint('')
         all_tags = experiment_set.all_experiment_tags()
@@ -587,6 +589,9 @@ def workspace_info(args):
 
                 for exp_name, _, _ in print_experiment_set.filtered_experiments(filters):
                     app_inst = experiment_set.get_experiment(exp_name)
+                    software_environments.render_environment(
+                        app_inst.expander.expand_var('{env_name}'), app_inst.expander
+                    )
 
                     if print_header:
                         color.cprint(rucolor.nested_1('  Application: ') +
@@ -650,8 +655,12 @@ def workspace_info(args):
     # Print software stack information
     if args.software:
         color.cprint('')
-        software_environments = ramble.software_environments.SoftwareEnvironments(ws)
-        software_environments.print_environments(verbosity=args.verbose)
+        #  software_environments.print_environments(verbosity=args.verbose)
+        color.cprint(rucolor.section_title('Software Stack:'))
+        color.cprint(software_environments.info(verbosity=args.verbose, indent=4, color_level=1))
+
+        #  print('All new envs:')
+        #  print(software_environments.new_envs)
 
 #
 # workspace list
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -589,9 +589,10 @@ def workspace_info(args):
 
                 for exp_name, _, _ in print_experiment_set.filtered_experiments(filters):
                     app_inst = experiment_set.get_experiment(exp_name)
-                    software_environments.render_environment(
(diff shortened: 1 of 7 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -650,8 +656,12 @@ def workspace_info(args):
     # Print software stack information
     if args.software:
         color.cprint('')
-        software_environments = ramble.software_environments.SoftwareEnvironments(ws)
-        software_environments.print_environments(verbosity=args.verbose)
+        #  software_environments.print_environments(verbosity=args.verbose)
+        color.cprint(rucolor.section_title('Software Stack:'))
+        color.cprint(software_environments.info(verbosity=args.verbose, indent=4, color_level=1))
+
+        #  print('All new envs:')
```
rfbgo: Shall we remove this?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -587,6 +589,10 @@ def workspace_info(args):
 
                 for exp_name, _, _ in print_experiment_set.filtered_experiments(filters):
                     app_inst = experiment_set.get_experiment(exp_name)
+                    if app_inst.uses_spack:
```
rfbgo, line 592: Not important for this PR, but I guess in the future we should abstract out the package manager and we more like:

```
if app_inst.package_manager == PACKAGE_MANAGERS.SPACK:
  ...
```

edit: Oh you even just added  a `package_managers` below in `software_environments` :)

lib/ramble/ramble/software_environments.py
```diff
@@ -20,402 +20,653 @@
...
+        """
 
-        conf_type = self._detect_conf_type()
+        self.name = name
+        self.spec = spec
+        self.compiler = compiler
+        self.compiler_spec = compiler_spec
+        self._package_type = 'Rendered'
+
+        if package_manager and hasattr(package_managers, package_manager):
```
rfbgo: Do you think this would ber cleaner with something more like 

```
self.package_manager = package_managers[package_manger]
```

In what situations are we expecting to get passed an invalid package manager and still want to continue?

lib/ramble/ramble/software_environments.py
```diff
@@ -20,402 +20,653 @@
...
+            indent (int): Number of spaces to inject as indentation
+            verbosity (int): Verbosity level
+
+        Returns:
+            (str): information of this environment
+        """
+
+        indentation = ''
+        for _ in range(0, indent):
+            indentation += ' '
```
rfbgo: nit: I feel like this is the C way to do this, and the python way is more like 
```
indentation = ' '*indent 
```

lib/ramble/ramble/software_environments.py
```diff
@@ -20,402 +20,653 @@
...
 
-    def _require_raw_package(self, pkg):
-        """Raise an error if the raw package is not defined"""
-        if pkg not in self._raw_packages.keys():
-            raise RambleSoftwareEnvironmentError(
-                f'Package {pkg} is not defined.'
-            )
+        if equal:
+            for self_pkg, other_pkg in zip(self._packages, other._packages):
+                equal = equal and self_pkg == other_pkg
```
rfbgo: nit: If find this a bit needlessly confusing. Any reason not to just do

```
if self_pkg != other_pkg:
  return False
```

rfbgo (dismissed):
Few comments, overall LGTM

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Yeah, I think we need to consider how to make the package manager stuff better though. So I didn't go too crazy into trying to design something that would work in a general sense.

lib/ramble/ramble/software_environments.py
douglasjacobsen: Yeah, that's a good point. I think we generally don't want to allow "unsupported" package managers. So it probably makes sense to just error here.

rfbgo (approved):

PR #434, opened 2024-03-07
Add a deployment command and publish pipeline

douglasjacobsen:
This merge adds a deployment command with publish and fetch as subcommands, and a publish deployment pipeline.

The deployment command can be used to publish (from) and fetch (into) deployments using a workspace definition.

### DIFF
--- a/lib/ramble/docs/workspace.rst
+++ b/lib/ramble/docs/workspace.rst
@@ -184,6 +184,70 @@ application definition files, one can use:
 
     $ ramble workspace concretize
 
+---------------------
+Workspace Deployments
+---------------------
+
+A deployment is one mechanism of transferring a configured workspace from one
+location to another. Ramble provides commands to handle creating (and pushing)
+a deployment from a local workspace to a remote location, or pulling a
+deployment from a remote location into a local workspace. 
+
+A deployment is a directory that contains the necessary artifacts required to
+recreate the experiments in the workspace on a separate machine. Deployments
+copy the workspace configuration file, along with creating an object
+repository, containing the application, modifier, and any package manager files
+needed for the experiments (that might not be upstreamed).  This section
+describes the commands that can be used to use deployments.
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Preparing a Workspace Deployment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Once a workspace is configured, it can be used to create a deployment.  To prepare a
+deployment, one can use:
+
+.. code-block:: console
+
+  $ ramble deployment push
+
+This will populate a directory named ``deployments``, where the default is the
+name of the workspace.
+
+The name of the created deployment can be controlled using:
+
+.. code-block:: console
+
+  $ ramble deployment push -d <deployment_name>
+
+Additionally, Ramble can create a tar of the deployment using:
+
+.. code-block:: console
+
+  $ ramble deployment push -t
+
+And upload the deployment to a remote URL using:
+
+.. code-block:: console
+
+  $ ramble deployment push -u <remote_url>
+
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Pulling a Workspace Deployment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To apply a deployment to an existing workspace, the ``pull`` sub-command can be used. For example:
+
+.. code-block:: console
+
+  ramble workspace pull -p file://path/to/deployment
+
+Will overwrite the contents of the currently active workspace with the contents
+from the deployment contained in ``file://path/to/deployment``.
+
+It is important to note that this command is destructive, and there is no way
+to revert a workspace back to its state prior to the pull action.
 
 .. _workspace-setup:
 
--- a/lib/ramble/ramble/cmd/deployment.py
+++ b/lib/ramble/ramble/cmd/deployment.py
@@ -1,0 +1,180 @@
+# Copyright 2022-2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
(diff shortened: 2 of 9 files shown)


lib/ramble/docs/workspace.rst
```diff
@@ -184,6 +184,43 @@ application definition files, one can use:
...
+
+.. code-block:: console
+
+  $ ramble deployment publish -t
+
+And upload the deployment to a remote URL using:
+
+.. code-block:: console
+
+  $ ramble deployment publish -u <remote_url>
```
rfbgo: Awesome stuff! Should we also document the "fetch" side?

rfbgo (approved):

lib/ramble/ramble/cmd/deployment.py
```diff
@@ -0,0 +1,178 @@
...
+        local_index_path = \
+            os.path.join(ws.root, publish_cls.index_filename)
+
+        fetch_file(remote_index_path, local_index_path)
+
+        with open(local_index_path, 'r') as f:
+            index_data = sjson.load(f)
+
+        for file in index_data[publish_cls.index_namespace]:
+            src = args.deployment_path + '/' + file
```
rfbgo: We use this pattern here, but we should probably rely on something like `os.join` since paths with too many `/` break in gcs buckets

rfbgo (approved):

lib/ramble/ramble/cmd/deployment.py
douglasjacobsen: My main concern was in URLs that might not want to use the OS separator (i.e. windows). But, this would also break local deployments at the OS level if you were publishing to a different directory.

Do you have any thoughts on how to manage this aspect of it?

rfbgo: So more like `urllib`'s `urljoin`?

lib/ramble/ramble/pipeline.py
```diff
@@ -521,9 +512,136 @@ def _execute(self):
...
+
+        aux_software_dir = os.path.join(configs_dir, ramble.workspace.auxiliary_software_dir_name)
+        fs.mkdirp(aux_software_dir)
+        aux_repo_conf = os.path.join(aux_software_dir, 'repos.yaml')
+        if os.path.exists(aux_repo_conf):
+            f = open(aux_repo_conf, 'a')
+        else:
+            f = open(aux_repo_conf, 'w+')
+            f.write('repos:\n')
+        f.write(f'- ../../{self.object_repo_name}\n')
```
rfbgo: There is some bug where if you do a `publish` for the same workspace we instead keep appending to files and end up with an invalid distribution. 

First publish:

```
./deployments/gke/configs/modifier_repos.yaml:- ../object_repo
./deployments/gke/configs/repos.yaml:- ../object_repo
```

After many publishes:
```
 $ grep -r "object_repo" .
./deployments/gke/configs/modifier_repos.yaml:- ../object_repo- ../object_repo- ../object_repo- ../object_repo- ../object_repo- ../object_repo- ../object_repo- ../object_repo
./deployments/gke/configs/repos.yaml:- ../object_repo- ../object_repo- ../object_repo- ../object_repo- ../object_repo- ../object_repo- ../object_repo- ../object_repo
./deployments/gke/configs/auxiliary_software_files/repos.yaml:- ../../object_repo
./deployments/gke/configs/auxiliary_software_files/repos.yaml:- ../../object_repo
./deployments/gke/configs/auxiliary_software_files/repos.yaml:- ../../object_repo
./deployments/gke/configs/auxiliary_software_files/repos.yaml:- ../../object_repo
./deployments/gke/configs/auxiliary_software_files/repos.yaml:- ../../object_repo
```
I think we need to add some logic around cleaning out old deployments

rfbgo (changes requested):

lib/ramble/ramble/pipeline.py
```diff
@@ -521,9 +512,136 @@ def _execute(self):
...
+
+        for conf_file in ['repos', 'modifier_repos']:
+            repos_config = os.path.join(configs_dir, conf_file + '.yaml')
+            if os.path.exists(repos_config):
+                f = open(repos_config, 'a')
+            else:
+                f = open(repos_config, 'w+')
+                f.write('repos:\n')
```
rfbgo: should this be `repos` or `f{conf_file}` or similar? 

I think the modifier repo needs the string `modifier_repos` here (https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/schema/modifier_repos.py#L18)

rfbgo (changes requested):

lib/ramble/docs/workspace.rst
douglasjacobsen: Yeah, and maybe (as we discussed offline) I'll change this to `push` and `pull`. To align with other tools that use these same commands.

lib/ramble/ramble/cmd/deployment.py
douglasjacobsen: Yes! But I also think it might be trickier than that... I wonder what urllib does if the root is `file://` Maybe I'll play with it some and see if that would be a better replacement than just injecting `/`.

lib/ramble/ramble/pipeline.py
douglasjacobsen: Ah that's fun. OK, I can look at that.

lib/ramble/ramble/pipeline.py
douglasjacobsen: Yeah, I think you're right. It should be a string instead of just `repos`.

douglasjacobsen:
@rfbgo  I think I addressed all of your comments. I'd like to squash these commits before this is merged, assuming you're happy with the changes though.

rfbgo:
> @rfbgo I think I addressed all of your comments. I'd like to squash these commits before this is merged, assuming you're happy with the changes though.

Looks good to me. Squash away and we can merge

douglasjacobsen:
@rfbgo All done!

rfbgo (approved):

PR #441, opened 2024-03-14
Fix env var setting in applications

rfbgo:
Previously, when an application.py sets environment variables (through the environment_variable directive) but a workspace doesn't define env_vars (as in `env_vars:set:OMP_NUM_THREADS:'{n_threads}'`) it fails to inject the environment variables into the rendered scripts properly.

This fixes it by ensuring a default format if the schema returns empty

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -803,6 +803,12 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
                 action = vals['action']
                 value = vals['value']
 
+                # Since the type coming from the schema can either be None, or
+                # a complex polymorphic type we need to ensure it has a
+                # sensible base structure when it is not given
+                if not self._env_variable_sets:
+                    self._env_variable_sets.append({'set': {}})
+
                 for env_var_set in self._env_variable_sets:
                     if action in env_var_set:
                         if name not in env_var_set[action].keys():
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/application.py
```diff
@@ -803,6 +803,12 @@ def _set_default_experiment_variables(self):
                 action = vals['action']
                 value = vals['value']
 
+                # Since the type coming from the schema can either be None, or
+                # a complex polymorphic type we need to ensure it has a
+                # sensible base structure when it is not given
+                if not self._env_variable_sets:
+                    self._env_variable_sets.append({'set': {}})
```
douglasjacobsen, line 810: Is there a good reason to not also define empty dicts for `append`, `prepend`, and `unset` other than those being unsupported methods of defining variables in application.py files?

douglasjacobsen (changes requested):
This LGTM for a fix.

Should we also add a test to make sure we don't encounter this issue in the future?

lib/ramble/ramble/application.py
rfbgo: Yea simply that we chose not to do it before, and it's not needed here

douglasjacobsen:
Can you incorporate the new test from this branch?https://github.com/douglasjacobsen/ramble/tree/fix_app_env_vars

rfbgo:
> Can you incorporate the new test from this branch?https://github.com/douglasjacobsen/ramble/tree/fix_app_env_vars

Done :) Good reminder and thanks for helping out with the tests

douglasjacobsen (approved):

PR #442, opened 2024-03-19
Deprecate Ramble 'concretized' flag

dapomeroy:
Deprecates 'concretized' flag and config variable and replaces them with 'force_concretize' option to import default software configurations. This simplifies use of Ramble without Spack and helps prepare for support of additional package managers.

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -341,7 +341,7 @@ def workspace_remove(args):
 def workspace_concretize_setup_parser(subparser):
     """Concretize a workspace"""
     subparser.add_argument(
-        '-f', '--force_concretize',
+        '-f', '--force-concretize',
         dest='force_concretize',
         action='store_true',
         help='Overwrite software environment configuration with defaults defined in application ' +
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -340,11 +340,20 @@ def workspace_remove(args):
 
 def workspace_concretize_setup_parser(subparser):
     """Concretize a workspace"""
-    pass
+    subparser.add_argument(
+        '-f', '--force_concretize',
+        dest='force_concretize',
+        action='store_true',
+        help='Overwrite software environment configuration with defaults defined in application ' +
+              'definition',
+        required=False)
 
 
 def workspace_concretize(args):
     ws = ramble.cmd.require_active_workspace(cmd_name='workspace concretize')
+
+    if args.force_concretize:
+        ws.force_concretize = True
 
     logger.debug('Concretizing workspace')
     ws.concretize()
--- a/lib/ramble/ramble/software_environments.py
+++ b/lib/ramble/ramble/software_environments.py
@@ -9,8 +9,8 @@
 from enum import Enum
 try:
     import deprecation
-except ImportError:
-    print('Please use pip to install the requirements.txt')
+except ModuleNotFoundError:
+    raise ModuleNotFoundError('Please use pip to install the requirements.txt')
 
 import ramble.repository
 import ramble.workspace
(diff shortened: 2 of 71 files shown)


lib/ramble/ramble/software_environments.py
```diff
@@ -413,7 +413,8 @@ class SoftwareEnvironments(object):
     """Class representing a group of software environments"""
 
     _deprecated_sections = [namespace.variables, namespace.zips,
-                            namespace.matrix, namespace.matrices, namespace.exclude]
+                            namespace.matrix, namespace.matrices,
+                            namespace.exclude, 'concretized']
```
rfbgo, line 421: Can we move this magic string into a variable like the others?

lib/ramble/ramble/software_environments.py
```diff
@@ -7,6 +7,10 @@
 # except according to those terms.
 
 from enum import Enum
+try:
+    import deprecation
+except ImportError:
```
rfbgo: Do you think we can figure out which of these deprecation imports will happen first, and only warn on one of them? 

Ideally we would stop ramble hiding the regular import fail error. I added this print as a bandaid since we only did the important in one place, but the more places the uglier it gets (but perhaps we can just have a follow up task to sort that)

rfbgo (changes requested):
Thanks for the changes!

One small nit, and one question otherwise LGTM

lib/ramble/ramble/software_environments.py
dapomeroy: I had asked Doug about this and he seemed to think it would be ok since it was in a 'deprecated' section and we'll remove it fairly soon.

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -340,12 +340,21 @@ def workspace_remove(args):
 
 def workspace_concretize_setup_parser(subparser):
     """Concretize a workspace"""
-    pass
+    subparser.add_argument(
+        '-f', '--force_concretize',
```
douglasjacobsen: Can you change this from `--force_concretize` to `--force-concretize` ? That's how we have the arguments in most of the other commands, and I think it'd prefer to keep the underscore / hyphen issues consistent.

lib/ramble/ramble/software_environments.py
douglasjacobsen: Yeah, I figured since it's going to be deleted soon, it wasn't necessarily worth adding into a variable. The other variables were already defined, and are still useful after `concretized` is removed.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -859,13 +859,17 @@ def external_spack_env(self, env_name):
...
         spack_dict = self.get_spack_dict()
 
-        if 'concretized' in spack_dict and spack_dict['concretized']:
-            raise RambleWorkspaceError('Cannot conretize an ' +
-                                       'already concretized ' +
-                                       'workspace')
+        if not self.force_concretize:
+            try:
+                if spack_dict[namespace.packages] or spack_dict[namespace.environments]:
+                    raise RambleWorkspaceError('Cannot concretize an already concretized '
```
douglasjacobsen, line 865: Can we make a task to track the work of making `concretize` additive as well (which would remove this message)?

douglasjacobsen (changes requested):
This looks great. One small change, and a request to make an internal task.

lib/ramble/ramble/software_environments.py
dapomeroy: When I uninstall 'deprecation' now, it seems like any command I try causes Ramble to error out on this one (SoftwareEnvironments.__init__()) rather than the others, so I'll leave this in and remove the warning from the rest

dapomeroy: Removed the try blocks per our chat discussion

lib/ramble/ramble/cmd/workspace.py
dapomeroy: Done

lib/ramble/ramble/workspace/workspace.py
dapomeroy: Added task

rfbgo (approved):

douglasjacobsen (approved):

PR #446, opened 2024-04-02
Enhance user experience around workspace archive

rfbgo:
- Add warning if -u is passed but not -t, since no action upload will be taken
- Add print to show the user the path the file was uploaded to


douglasjacobsen:
Looks good! Can you rebase onto develop so the tests pass?

douglasjacobsen (approved):

PR #454, opened 2024-04-04
Update context output format to reduce duplication

rfbgo:
This is a **breaking change** for anyone who consumes our output format wrt contexts

This PR makes it so that contexts only use the format string when printing, to reduce how repetitive and annoying to construct the format is.

The new output looks like:

txt
```
 10   Bytes: 0 (BW) figures of merit:
 11     Bandwidth = 0.00 Mbytes/sec
```

json
```
131         {
132           "name": "Bytes: 0 (BW)",
133           "foms": [
134             {
135               "value": "0.00",
136               "units": "Mbytes/sec",
137               "origin": "IntelMpiBenchmarks",
138               "origin_type": "application",
139               "name": "Bandwidth"
140             }
141           ]
142         },
```

This is also reflected the same way in the upload db

The old formats looked like:

txt
```
 10   Bytes: bytes (BW) = Bytes: 0 (BW) figures of merit:
 11     Bandwidth = 0.00 Mbytes/sec
```

json
```
  "name": "Bytes: bytes (BW) = Bytes: 0 (BW)",
133           "foms": [
134             {
135               "value": "0.00",
136               "units": "Mbytes/sec",
137               "origin": "IntelMpiBenchmarks",
138               "origin_type": "application",
139               "name": "Bandwidth"
140             }
141           ]
142         },
```

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -1427,8 +1427,7 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
                     if group[1]:
                         context_val[group[1]] = context_match[group[1]]
 
-            context_string = context_format.replace('{', '').replace('}', '') \
-                + ' = ' + context_format.format(**context_val)
+            context_string = context_format.format(**context_val)
             return context_string
 
         fom_values = {}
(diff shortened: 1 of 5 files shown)


douglasjacobsen (dismissed):
Looks good, once the CI passes

douglasjacobsen (approved):

PR #464, opened 2024-04-16
Update workload variables to use a dict

rfbgo:
This was causing undesirable behavior where applications that used inheritence could define could the same variable twice, since the list did not allow for collision detection on the name

This PR updates it to use a dict, so a given var name can only be used once with the child class taking precendence


rfbgo:
I want to also port modifier variables to use the current `WorkloadVariable` class as part of this, so marking as draft for now

douglasjacobsen (approved):

PR #467, opened 2024-04-18
Support globbing in directive list arguments

dapomeroy:
Adds support for wildcards in list arguments within directives, so that arbitrary workloads can be supported more easily.

### DIFF
--- a/lib/ramble/ramble/language/application_language.py
+++ b/lib/ramble/ramble/language/application_language.py
@@ -73,11 +73,14 @@ def workload(name, executables=None, executable=None, input=None,
     def _execute_workload(app):
         all_execs = ramble.language.language_helpers.require_definition(executable,
                                                                         executables,
+                                                                        app.executables,
                                                                         'executable',
                                                                         'executables',
                                                                         'workload')
 
-        all_inputs = ramble.language.language_helpers.merge_definitions(input, inputs)
+        all_inputs = ramble.language.language_helpers.merge_definitions(input,
+                                                                        inputs,
+                                                                        app.inputs)
 
         app.workloads[name] = ramble.workload.Workload(name, all_execs, all_inputs, tags)
 
@@ -164,6 +167,7 @@ def workload_variable(name, default, description, values=None, workload=None,
     def _execute_workload_variable(app):
         all_workloads = ramble.language.language_helpers.require_definition(workload,
                                                                             workloads,
+                                                                            app.workloads,
                                                                             'workload',
                                                                             'workloads',
                                                                             'workload_variable')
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/test/end_to_end/globbing_patterns.py
dapomeroy, line 1: What's the right amount of rigor to add to the tests here? This one checks workload(executables), and all the rest use the same helper method for expansion. Seems to run quite fast. Think it's worthwhile to add tests for variables, env_vars, mods, etc.?

douglasjacobsen: Yeah, I think we should try to use a kind of ridiculous amount of globbing to help make sure it all works.

If it's in an end-to-end test, you can just test all of it in a single test though.

lib/ramble/ramble/language/application_language.py
```diff
@@ -71,13 +71,22 @@ def workload(name, executables=None, executable=None, input=None,
     """
 
     def _execute_workload(app):
+        exp_executables = executables
+        if isinstance(exp_executables, list):
+            exp_executables = ramble.language.language_helpers.expand_patterns(app.executables,
```
douglasjacobsen: Could this logic be pushed down into `require_definition` or `merge_definition`? (i.e. the list check that then expands and merges with the singular definition)

lib/ramble/ramble/language/modifier_language.py
```diff
@@ -81,8 +81,13 @@ def _execute_variable_modification(mod):
             raise DirectiveError('variable_modification directive given an invalid method.\n'
                                  f'  Valid methods are {str(supported_methods)}')
 
+        exp_modes = modes
+        if isinstance(exp_modes, list):
```
douglasjacobsen: Same comment about pushing the logic down.

douglasjacobsen (changes requested):
One request. Otherwise LGTM! Thanks!

lib/ramble/ramble/language/application_language.py
dapomeroy: Definitely. I've updated this to be in `merge_definition` and it's much cleaner now.

lib/ramble/ramble/language/modifier_language.py
dapomeroy: Done

lib/ramble/ramble/test/end_to_end/globbing_patterns.py
dapomeroy: OK, I've added a lot to the test and it still runs fast.

douglasjacobsen (approved):

PR #468, opened 2024-04-19
Add support for typed expansion, and yaml type passthrough

douglasjacobsen:
This merge changes some of the logic around implicitly identifying variable usage through rendering, and removing unused variables. No variable definitions are removed anymore after this merge.

Zips and matrices convey usage information to the rendered now, and are used to generate experiments from multiple lists.

Variables in the ramble.yaml are now allowed to have any type (i.e. lists or dicts).

Additionally, the `expander.expand_var` and `expander.expand_var_name` have a `typed=False` attribute (false by default) which will force the expander to return a python typed version of the expanded variable.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -13,6 +13,8 @@ import operator
 import math
 import random
 
+from typing import Dict
+
 import ramble.error
 import ramble.keywords
 from ramble.util.logger import logger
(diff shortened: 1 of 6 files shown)


rfbgo (dismissed):

rfbgo (approved):
Perhaps unsurprising, but I really love the changes we discussed! Awesome work

PR #476, opened 2024-05-03
required package setting breaks benchpark

jdomke:
this is related to this issue https://github.com/LLNL/benchpark/issues/225
(unsure if removing those lines does not break anything else, but at least it fixes benchpark)


douglasjacobsen:
Fixed in https://github.com/GoogleCloudPlatform/ramble/pull/477

PR #483, opened 2024-05-09
Redirect stderr to stdout for bash and sh

douglasjacobsen:
This merge adds redirection of stderr to stdout for the bash and sh shells.


linsword13:
Change LGTM. There's a test failure (test_custom_executables) though.

linsword13 (approved):

PR #484, opened 2024-05-10
Add keywords as top level attributes to base repeat experiments

douglasjacobsen:
Previously, base experiments in a repeat set didn't have their top level attributes set correctly (i.e. application_name) which caused uploading to fail.

This merge adds the top level attributes to base experiments allowing them to be uploaded.


dapomeroy (approved):
LGTM

PR #485, opened 2024-05-13
Ensure chained experiment definitions are copies of base

douglasjacobsen:
This merge fixes an issue where chained experiment definitions were all references to the base experiment definition. As a result, when an experiment generated the commands for its chained experiments the commands were never allowed to render again. This caused repeat experiments (which clone a base experiment) to have the incorrect execution command (which only referred to the base experiment).


dapomeroy (approved):
LGTM

PR #487, opened 2024-05-15
Change the display of null context

linsword13:
The intent is to make it less confusing by indicating the "null" here refers to the null context.

Also add in debug printing for the analysis results, just for convenience.


linsword13:
Definitely open for suggestions for how to make the display clearer.

douglasjacobsen (approved):

PR #489, opened 2024-05-16
Presubmit verify lic header for changed files only

linsword13:
The full scan will still happen in the pr-style check.

This makes it a bit faster for local commits.

Tested:

```sh
$ sed -i '1d' lib/ramble/ramble/cmd/license.py

$ git commit -a
...
Flake8 checks were clean.
Modified files:
 ['lib/ramble/ramble/cmd/license.py', 'share/ramble/qa/run-flake8-tests']
/usr/local/google/home/linsword/hpc-dev/ramble/lib/ramble/ramble/cmd/license.py: the license does not match the expected format
==> Error: 1 improperly licensed files
  files with wrong SPDX-License-Identifier:   0
  files with old license header:              0
  files not containing expected license:      1
```

### DIFF
--- a/lib/ramble/ramble/cmd/license.py
+++ b/lib/ramble/ramble/cmd/license.py
@@ -65,20 +65,30 @@ lgpl_exceptions = [
 ]
 
 
-def _all_ramble_files(root=ramble.paths.prefix):
+def _get_modified_files(root):
+    """Get a list of modified files in the current repository."""
+    diff_args = ['-C', root, 'diff', 'HEAD', '--name-only']
+    files = git(*diff_args, output=str).split()
+    return files
+
+
+def _all_ramble_files(root=ramble.paths.prefix, modified_only=False):
     """Generates root-relative paths of all files in the ramble repository."""
-    visited = set()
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/cmd/license.py
```diff
@@ -65,20 +65,31 @@
 ]
 
 
-def _all_ramble_files(root=ramble.paths.prefix):
+def _get_modified_files(root):
+    """Get a list of modified files in the current repository."""
+    diff_args = ['-C', root, 'diff', 'HEAD', '--name-only']
+    files = git(*diff_args, output=str).split()
+    print('Modified files:\n', files)
```
douglasjacobsen: Do we need the print here?

douglasjacobsen (dismissed):
Looks good! One question, otherwise it's fine to merge.

lib/ramble/ramble/cmd/license.py
linsword13: Good catch. It was a remnant from local debugging. Removed.

douglasjacobsen (approved):

PR #495, opened 2024-05-18
Configure spack install dir based on `env_name`

linsword13:
Configure spack install dir based on `env_name`
    
This helps with saving redundant spack concretization, when experiments
under different workloads are actually using the same software environment.

Some more contexts (courtesy of Bob and Doug):

* Historically the spack env was rendered during workspace render time,
  that's why the existing namespacing.

* That's changed in https://github.com/GoogleCloudPlatform/ramble/pull/433/.

* And it's safe to assume same `env_name` always 1-1 maps to a spack
  env. There's a check to guarantee that: https://github.com/GoogleCloudPlatform/ramble/blob/f4e0e06c7f4ab0efd5d05000cb7ebec7b3ad33b2/lib/ramble/ramble/software_environments.py#L371-L374.
    
Tested: tested with the following:

* A new created workspace with gromacs ran fine.

* (backward-compt) An existing workspace created from the current
  develop head ran fine after running `ramble workspace setup` under the
  current branch.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -318,7 +318,6 @@ class Expander(object):
         self._application_namespace = None
         self._workload_namespace = None
         self._experiment_namespace = None
-        self._env_namespace = None
         self._env_path = None
 
         self._application_input_dir = None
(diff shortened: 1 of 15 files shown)


lib/ramble/ramble/expander.py
```diff
@@ -388,15 +387,6 @@ def experiment_namespace(self):
 
         return self._experiment_namespace
 
-    @property
-    def env_namespace(self):
```
linsword13, line 392: This is removed as it's not currently in use.

douglasjacobsen (approved):

PR #499, opened 2024-05-21
Create instance variables for directives

douglasjacobsen:
This merge changes the type of attributes created from directives from class attributes to instance attributes. This allows modification of an instance's attributes without affecting all of other instances.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -45,6 +45,7 @@ import ramble.util.env
 import ramble.util.directives
 import ramble.util.stats
 import ramble.util.graph
+import ramble.util.class_attributes
 from ramble.util.logger import logger
 
 from ramble.workspace import namespace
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/application.py
```diff
@@ -89,6 +89,12 @@ class ApplicationBase(object, metaclass=ApplicationMeta):
     def __init__(self, file_path):
         super().__init__()
 
+        # Convert metaclass directives from class attributes to instance attributes
```
linsword13: Pretty nifty! Can this be made into a util function and shared with the modifier code?

Also curious if there's an example showing the usage of this change, or maybe a bug that's caused by the current behavior? Perhaps a test case would be informative.

douglasjacobsen: Done and done. :)

linsword13 (approved):

PR #502, opened 2024-05-22
Proof of concept for workload_groups

rfbgo:
This change add a new directive: `workload_groups`

It is intended to give the user a way to group workloads together, and unlocks a key CUJ where inherited applications want to apply variables defined in parent classes

### DIFF
--- a/lib/ramble/ramble/language/application_language.py
+++ b/lib/ramble/ramble/language/application_language.py
@@ -75,19 +75,16 @@ def workload(
             executable, executables, app.executables, "executable", "executables", "workload"
         )
 
-        all_inputs = ramble.language.language_helpers.merge_definitions(input,
-                                                                        inputs,
-                                                                        app.inputs,
-                                                                        'input',
-                                                                        'inputs',
-                                                                        'workload')
+        all_inputs = ramble.language.language_helpers.merge_definitions(
+            input, inputs, app.inputs, "input", "inputs", "workload"
+        )
 
         app.workloads[name] = ramble.workload.Workload(name, all_execs, all_inputs, tags)
 
     return _execute_workload
 
 
-@application_directive('workload_groups')
+@application_directive("workload_groups")
 def workload_group(name, workloads=[], mode=None, **kwargs):
     """Adds a workload group to this application
 
@@ -100,7 +97,7 @@ def workload_group(name, workloads=[], mode=None, **kwargs):
     """
 
     def _execute_workload_groups(app):
-        if mode == 'append':
+        if mode == "append":
             app.workload_groups[name].update(set(workloads))
         else:
             app.workload_groups[name] = set(workloads)
@@ -114,7 +111,7 @@ def workload_group(name, workloads=[], mode=None, **kwargs):
     return _execute_workload_groups
 
 
-@application_directive('executables')
+@application_directive("executables")
 def executable(name, template, **kwargs):
     """Adds an executable to this application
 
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/language/application_language.py
```diff
@@ -86,6 +86,34 @@ def _execute_workload(app):
...
+    """
+
+    def _execute_workload(app):
+        # TODO where is it best to validate the workloads
+        if mode == 'append':  # TODO: dry magic enum string
+            app.workload_groups[name] += workloads
+        else:
+            app.workload_groups[name] = workloads
+
+        # Apply any existing variables in the group to the workload
```
douglasjacobsen, line 105: It seems like this has to be added later regardless of if it's added now. Could this be removed if that's the case?

lib/ramble/ramble/language/application_language.py
```diff
@@ -165,20 +194,45 @@ def workload_variable(name, default, description, values=None, workload=None,
...
-                                                                            'workload',
-                                                                            'workloads',
-                                                                            'workload_variable')
-
-        for wl_name in all_workloads:
-            app.workloads[wl_name].add_variable(
-                ramble.workload.WorkloadVariable(
-                    name, default=default, description=description,
-                    values=values, expandable=expandable
+        if (workload is not None or workloads is not None) and workload_group is None:
```
douglasjacobsen: It seems like this could just be `if workload_group is None` --> Require definition.

lib/ramble/ramble/language/application_language.py
```diff
@@ -165,20 +194,45 @@ def workload_variable(name, default, description, values=None, workload=None,
...
+
+            for wl_name in all_workloads:
+                app.workloads[wl_name].add_variable(
+                    ramble.workload.WorkloadVariable(
+                        name, default=default, description=description,
+                        values=values, expandable=expandable
+                    )
                 )
-            )
+        elif workload is None and workloads is None and workload_group is not None:
```
douglasjacobsen: I think here you could use the `merge_definition` instead of `require_definition` from above, to get a merged workload + workloads. And then merge that with workload groups either in a helper or manually.

lib/ramble/ramble/language/application_language.py
```diff
@@ -165,20 +194,45 @@ def workload_variable(name, default, description, values=None, workload=None,
...
+            workload_group_list = app.workload_groups[workload_group]
+
+            # TODO: better place to init this?
+            if workload_group not in app.workload_group_vars:
+                app.workload_group_vars[workload_group] = []
+
+            # Track which vars we add to, to allow us to re-apply during inheritance
+
+            # TODO: dry this add with above?
+            workload_var = ramble.workload.WorkloadVariable(
```
douglasjacobsen: Maybe we should make a copy method (or clone?) on WorkloadVariable, so we can create one initially, and copy it into the different workloads.

lib/ramble/ramble/language/application_language.py
rfbgo: Agreed. Change made in upcoming commit

lib/ramble/ramble/language/application_language.py
rfbgo: Addressed in the commit I just pushed, ptal :)

douglasjacobsen:
LG! I think there are conflicts due to the style changes. Can you fix those and then I can merge?

rfbgo:
Rebased and updated :)

douglasjacobsen (approved):

PR #506, opened 2024-05-23
Update spack/util/debug.py to fix io.UnsupportedOperation paralell tests

rfbgo:
Without this we get:

```
ERROR lib/ramble/ramble/test/workspace_hashing/workspace_setup_creates_inventory.py - io.UnsupportedOperation: redirected stdin is pseudofile, has no fileno()
```

(We previously fixed something like this)

### DIFF
--- a/lib/ramble/spack/util/debug.py
+++ b/lib/ramble/spack/util/debug.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2024 Lawrence Livermore National Security, LLC and other
+# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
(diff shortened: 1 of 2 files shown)


lib/ramble/spack/util/debug.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Copyright 2013-2024 Lawrence Livermore National Security, LLC and other
```
douglasjacobsen: Is it OK to update the copyright for this? IMO, it should stay the same as when we vendored spack in, but I could be wrong about that.

douglasjacobsen (dismissed):

lib/ramble/spack/util/debug.py
rfbgo: I left it like that since I took the updated it directly from the current spack. I think it's a good question. To me keeping it like this was better because:

1. It's true to the file I bought in
2. It's a nice marker to us ramble folks that it's already updated relative to the rest of `./spack`

I'm happy to do whatever you think is best, I just felt weird about downgrading the header such that it mismatches the contents of the file

rfbgo: By luck there was an old version with the file contents we want and matching header. Going back to that to avoid any accidental headaches

douglasjacobsen (approved):

PR #508, opened 2024-05-23
Merge release v0.5.0 to develop

linsword13:
Changes included:

* Update version number to 0.5.0
* Deprecate `default_compiler`
* Deprecate `_deprecated_warnings`

### DIFF
--- a/lib/ramble/docs/dev_guides/application_dev_guide.rst
+++ b/lib/ramble/docs/dev_guides/application_dev_guide.rst
@@ -265,6 +265,7 @@ directives that are intended to be package manager specific. As an example,
 there are directives for Spack defined by:
 
 * :meth:`ramble.language.shared_language.software_spec`
+* :meth:`ramble.language.shared_language.define_compiler`
 * :meth:`ramble.language.shared_language.required_package`
 
 These provide Ramble with information about how Spack could install and require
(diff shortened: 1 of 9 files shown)


lib/ramble/docs/dev_guides/application_dev_guide.rst
```diff
@@ -265,7 +265,6 @@ directives that are intended to be package manager specific. As an example,
 there are directives for Spack defined by:
 
 * :meth:`ramble.language.shared_language.software_spec`
-* :meth:`ramble.language.shared_language.default_compiler`
```
douglasjacobsen, line 268: I think instead of just removing default compiler, we should replace it with `define_compiler`

lib/ramble/ramble/software_environments.py
```diff
@@ -416,10 +412,6 @@ def all_packages(self):
 class SoftwareEnvironments(object):
     """Class representing a group of software environments"""
 
-    _deprecated_sections = [namespace.variables, namespace.zips,
```
douglasjacobsen, line 419: In addition to removing them from here, these should be removed from the schema too: https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/schema/spack.py#L44-L48

douglasjacobsen (changes requested):
Looks good. Just missing the schema changes and it should be good to go!

lib/ramble/docs/dev_guides/application_dev_guide.rst
linsword13: Good catch, fixed.

lib/ramble/ramble/software_environments.py
linsword13: Ah right, sorry for the omission. Removed.

douglasjacobsen (approved):

PR #513, opened 2024-05-24
Only compute var and stddev with valid inputs

linsword13:
Also change the output to be "NA" when less than 2 inputs are given.


dapomeroy (approved):
LGTM

PR #515, opened 2024-05-30
Add `black` code formatter

linsword13:
Changes included:

* Add in a `ramble style` cmd. Currently it supports running `flake8` and `black`
* This new `style` cmd will replace `ramble flake8` in the future

Next step is to run `ramble style -a -t black -f` to apply auto-format to all the in-scope files.

### DIFF
--- a/lib/ramble/ramble/cmd/style.py
+++ b/lib/ramble/ramble/cmd/style.py
@@ -25,8 +25,8 @@ section = "developer"
 level = "long"
 
 
-def is_application(f):
-    """Whether flake8 should consider a file as a core file or a application.
+def is_object(f):
+    """Whether flake8 should consider a file as a core file or an object (application or modifier).
 
     We run flake8 with different exceptions for the core and for
     applications, since we allow `from ramble import *` and poking globals
@@ -341,8 +341,8 @@ def run_flake8(flake8_cmd, file_list, args):
         print_tool_header("flake8", file_list)
 
         # run flake8 on the temporary tree, once for core, once for apps
-        application_file_list = [f for f in file_list if is_application(f)]
-        primary_file_list = [f for f in file_list if not is_application(f)]
+        application_file_list = [f for f in file_list if is_object(f)]
+        primary_file_list = [f for f in file_list if not is_object(f)]
 
         # filter files into a temporary directory with exemptions added.
         # TODO: DRY this duplication
--- a/lib/ramble/ramble/cmd/style.py
+++ b/lib/ramble/ramble/cmd/style.py
@@ -89,16 +89,13 @@ pattern_exemptions = {
 pattern_exemptions = dict(
     (
         re.compile(file_pattern),
-        dict(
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/cmd/style.py
```diff
@@ -0,0 +1,488 @@
...
+import ramble.paths
+from spack.util.executable import which, ProcessError
+
+
+description = "runs source code style checks on Ramble."
+section = "developer"
+level = "long"
+
+
+def is_application(f):
```
douglasjacobsen: What about other things? Should we differentiate between modifiers and applications?

It looks like this code just says "Are you in a repo" rather than "Are you an application", if that's the case can you change this to "is_object" instead?

lib/ramble/ramble/cmd/style.py
```diff
@@ -0,0 +1,488 @@
...
+#: This is a dict that maps:
+#:  filename pattern ->
+#:     flake8 exemption code ->
+#:        list of patterns, for which matching lines should have codes applied.
+#:
+#: For each file, if the filename pattern matches, we'll add per-line
+#: exemptions if any patterns in the sub-dict match.
+pattern_exemptions = {
+    # exemptions applied only to application.py files.
+    r"application.py$": {
```
douglasjacobsen, line 77: There should probably be a pattern exemption for other types in the repo. Maybe we could auto inject these by iterating over the object types in the repoistory module?

douglasjacobsen (changes requested):
This is awesome, thanks for working on this!

lib/ramble/ramble/cmd/style.py
linsword13: Ack and good point, renamed to `is_object`. I copied this code from the existing flake cmd impl, so I am more inclined to keep it as-is for now, and see if we need to differentitate between app and mod down the line. Definitely open for suggestions though.

lib/ramble/ramble/cmd/style.py
linsword13: Good point, added exemptions for modifier. I didn't directly use the ObjectTypes enum, given that the import pattern is different between app and mod (appkit vs modkit,) so there need to have some hard-coded knowledge on that anyway.

lib/ramble/ramble/cmd/style.py
douglasjacobsen: SG. I think it's probably the same reading the actual code. I think it's more like "Is this an object in the repo" than it is "Is this actually an application".

Spack (as currently implemented) only has a concept of applications in repos, and we extended that which is why some of the stuff still refers to applications.

Thanks for the change!

lib/ramble/ramble/cmd/style.py
douglasjacobsen: OK cool. We'll need to update this with the package manager code as well. I'll add a todo for us to do that.

douglasjacobsen (approved):

PR #520, opened 2024-06-03
Update openfoam application definitions

douglasjacobsen:
This merge adds a definition for openfoam.org (called openfoam-org), and updated the previous openfoam application definition to inherit from the openfoam-org definition.

### DIFF
--- a/var/ramble/repos/builtin/applications/openfoam-org/application.py
+++ b/var/ramble/repos/builtin/applications/openfoam-org/application.py
@@ -1,0 +1,436 @@
+# Copyright 2022-2024 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import os
+from ramble.appkit import *
+from ramble.expander import Expander
+
+
+class OpenfoamOrg(SpackApplication):
+    """Define the Openfoam application"""
+
+    name = "openfoam-org"
+
+    maintainers("douglasjacobsen")
+
+    tags("cfd", "fluid", "dynamics")
+
+    define_compiler("gcc9", spack_spec="gcc@9.3.0")
+
+    software_spec(
+        "ompi412",
+        spack_spec="openmpi@4.1.2 +legacylaunchers +cxx",
+        compiler="gcc9",
+    )
+
+    software_spec(
+        "openfoam-org", spack_spec="openfoam-org@10", compiler="gcc9"
+    )
+
+    required_package("openfoam-org")
+
+    workload(
+        "motorbike",
+        executables=[
+            "clean",
+            "get_inputs",
+            "configure_mesh",
+            "surfaceFeatures",
+            "blockMesh",
+            "decomposePar1",
+            "snappyHexMesh",
+            "configure_simplefoam",
+            "redistributePar",
+            "decomposePar2",
+            "patchSummary",
+            "potentialFoam",
+            "checkMesh",
+            "simpleFoam",
+        ],
+    )
+
+    workload(
+        "motorbike_20m",
+        executables=[
+            "clean",
+            "get_inputs",
+            "configure_mesh",
+            "surfaceFeatures",
+            "blockMesh",
+            "decomposePar1",
+            "snappyHexMesh",
+            "configure_simplefoam",
+            "redistributePar",
+            "decomposePar2",
+            "patchSummary",
+            "potentialFoam",
+            "checkMesh",
+            "simpleFoam",
+        ],
+    )
+
+    workload(
+        "motorbike_42m",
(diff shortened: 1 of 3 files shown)


var/ramble/repos/builtin/applications/openfoam-org/application.py
```diff
@@ -0,0 +1,247 @@
...
+                    fom_regex=r'\s*ExecutionTime = (?P<foam_time>[0-9]+\.?[0-9]*).*',
+                    group_name='foam_time', units='s')
+
+    success_criteria('snappyHexMesh_completed', mode='string', match='Finalising parallel run',
+                     file='{experiment_run_dir}/log.snappyHexMesh')
+
+    success_criteria('simpleFoam_completed', mode='string', match='Finalising parallel run',
+                     file='{experiment_run_dir}/log.simpleFoam')
+
+    def _define_commands(self, exec_graph):
```
rfbgo, line 466: We agreed offline that this pattern is not something we'd ideally want in an application definition and instead would make more sense as a modifier or done through some other mechanism.

We will accept this for now to get unblocked and come back and tidy it up when possible

var/ramble/repos/builtin/applications/openfoam/application.py
```diff
@@ -6,221 +6,57 @@
...
+                              'FOAM_LIBBIN,FOAM_MPI,FOAM_RUN,FOAM_SITE_APPBIN,FOAM_SITE_LIBBIN,' +
+                              'FOAM_SOLVERS,FOAM_SRC,FOAM_TUTORIALS,FOAM_USER_APPBIN,' +
+                              'FOAM_USER_LIBBIN,FOAM_UTILITIES,LD_LIBRARY_PATH,PATH,' +
+                              'WM_ARCH,WM_COMPILER,WM_COMPILER_LIB_ARCH,WM_COMPILER_TYPE,' +
+                              'WM_COMPILE_OPTION,WM_DIR,WM_LABEL_OPTION,WM_LABEL_SIZE,' +
+                              'WM_MPLIB,WM_OPTIONS,WM_PRECISION_OPTION,WM_PROJECT,WM_PROJECT_DIR,' +
+                              'WM_PROJECT_USER_DIR,WM_PROJECT_VERSION,WM_THIRD_PARTY_DIR',
+                      workloads=['*'])
+
+    # TODO: Remove when base classes exist
```
rfbgo, line 74: Expanding on TODO:

Offline we agreed that this should be a base class in the future. We will soon have a formal "application" type so we will defer until then 

It will let us remove this, per the TODO

var/ramble/repos/builtin/applications/openfoam-org/application.py
```diff
@@ -0,0 +1,247 @@
...
+
+    figure_of_merit('Number of cells', log_file=(log_prefix + 'snappyHexMesh'),
+                    fom_regex=r'Layer mesh\s+:\s+cells:(?P<ncells>[0-9]+)\s+.*',
+                    group_name='ncells', units='')
+
+    figure_of_merit('snappyHexMesh Time ({n_ranks_hex} ranks)', log_file=(log_prefix + 'snappyHexMesh'),
+                    fom_regex=r'Finished meshing in = (?P<mesh_time>[0-9]+\.?[0-9]*).*',
+                    group_name='mesh_time', units='s')
+
+    figure_of_merit('simpleFoam Time ({n_ranks} ranks)', log_file=(log_prefix + 'simpleFoam'),
```
rfbgo: Are we sure we want a dynamic fom name based on `n_ranks`? Stuffing values in the keys/LHS like this makes comparing across runs much harder

I think it should probably be 2 FOMs:

1. Ranks used for each bit
2. Time for each bit 

That way we still support `SELECT * from FOMs where fom_name == 'simpleFoam Time'`. That is much harder when it's dynamic

rfbgo (dismissed):
Few notes, mostly already discussed but check out the comment on the FOMs

Happy to merge once tests past (and likely foms addressed)

var/ramble/repos/builtin/applications/openfoam-org/application.py
douglasjacobsen: OK sure, so you want `simpleFoam ranks` as a separate FOM, right? That's easy to add. I really just wanted to make it clear in the FOMs that the number of ranks between the parts are allowed to be different.

rfbgo: Yea I think it's kind of tricky. I agree it's important that we clearly mark that, but I think having the key/LHS change is kind of dangerous

In my mind different runs change overall `n_ranks` and we don't mark that in the FOM (so we can better compare) so it's better to follow that pattern here

I think a FOM tracking how many of the ranks used for the decomposition makes most sense, but I'm open to other ideas

douglasjacobsen: OK. I can add this. I think we might want to make an internal task for converting a variable into a FOM directly, as that isn't possible right now. So, something gross has to happen (like just writing the variable out into a file so we can then extract it again as a FOM). I just pushed an update though. Let me know what you think.

rfbgo (approved):

PR #522, opened 2024-06-05
Clean up unit-test warnings

linsword13:
* Fix up the suppression of "invalid decimal" in chained tests. In 3.11/3.12 it changed from being a `DeprecationWarning` to a `SyntaxWarning`.
* Suppress "invalid escape sequence" in known_apps test. This aligns with ignoring W503 in flake8_applications.
* Suppress a couple new deprecation warnings introduced in 3.12
  - Deprecation on AST node types (future TODO)
  - Warning on mix of thread and fork

Tested:

```sh
$ python3 --version
Python 3.12.3

$ ramble unit-test
...
==== 1050 passed in 496.54s (0:08:16) ====

$ python3 --version
Python 3.10.14

$ ramble unit-test
...
==== 1050 passed in 515.27s (0:08:35) ====
```

### DIFF
--- a/lib/ramble/ramble/test/cmd/mods.py
+++ b/lib/ramble/ramble/test/cmd/mods.py
@@ -6,8 +6,6 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
-import pytest
-
 from ramble.main import RambleCommand
 
 mods = RambleCommand("mods")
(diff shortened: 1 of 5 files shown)


rfbgo:
Do you have any ideas where the warning is actually originating from? I think there is a chance it's a real/valid warning (but also pretty hard to track down..)

linsword13:
> Do you have any ideas where the warning is actually originating from? I think there is a chance it's a real/valid warning (but also pretty hard to track down..)

Yeah the three newly added suppressed warnings (all from 3.12) are pretty well-understood. The ast node needs to be updated once we drop 3.6 and 3.7 support (I added a TODO there.) The fork warning seems harmless as that's caused by issuing multiple commands in the same test. For the "invalid decimal", that came form `fnmatch` and previously Doug investigated that in https://github.com/GoogleCloudPlatform/ramble/pull/139.

douglasjacobsen (approved):

PR #524, opened 2024-06-07
Manually wrap pip errors to help remind users to install requirements

rfbgo:


lib/ramble/ramble/graphs.py
```diff
@@ -103,7 +103,11 @@ def walk(self):
             node (GraphNode): Each node in the graph
         """
         if not self._prepared:
-            sorter = graphlib.TopologicalSorter(self.adj_list)
+            try:
+                sorter = graphlib.TopologicalSorter(self.adj_list)
+            except AttributeError as e:
```
linsword13: nit: do we want to include in the original error `e`? If not maybe we just remove `as e`?

linsword13 (dismissed):

lib/ramble/ramble/graphs.py
rfbgo: Ha yea, thanks! I actually meant to remove the `as e` and `ramble style` caught it too 😎

linsword13 (approved):

douglasjacobsen (approved):

PR #528, opened 2024-06-12
Add a util for ramble app to get file path

linsword13:
This serves as a thin wrapper to facilitate mocking out such paths during unit-tests.

### DIFF
--- a/lib/ramble/ramble/spack_runner.py
+++ b/lib/ramble/ramble/spack_runner.py
@@ -617,8 +617,8 @@ class SpackRunner(object):
             self._dry_run_print(self.spack, name_args)
             self._dry_run_print(self.spack, loc_args)
 
-            name = os.path.join(shlex.split(package_spec)[0])
-            location = os.path.join("dry-run", "path", "to", shlex.split(package_spec)[0])
+            name = shlex.split(package_spec)[0].split("@")[0]
+            location = os.path.join("dry-run", "path", "to", name)
             return (name, location)
 
     def mirror_environment(self, mirror_path):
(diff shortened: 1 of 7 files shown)


lib/ramble/ramble/spack_runner.py
```diff
@@ -617,8 +617,8 @@ def get_package_path(self, package_spec):
             self._dry_run_print(self.spack, name_args)
             self._dry_run_print(self.spack, loc_args)
 
-            name = os.path.join(shlex.split(package_spec)[0])
-            location = os.path.join("dry-run", "path", "to", shlex.split(package_spec)[0])
+            name = shlex.split(package_spec)[0].split("@")[0]
```
linsword13, line 620: This change is independent of the main change of this PR. I noticed that currently the dry-run would insert this path in the variables as `{"maxtext@main": "/path/to..."}` such that referencing `{maxtext}` in `ramble.yaml` doesn't work as expected under dry-run.

lib/ramble/ramble/test/get_file_path.py
```diff
@@ -0,0 +1,66 @@
...
+)
+
+workspace = RambleCommand("workspace")
+
+
+def test_get_file_path(mock_applications, mock_file_auto_create):
+    test_config = """
+ramble:
+  config:
+    disable_progress_bar: true
```
linsword13, line 32: I had to disable `tqdm` here otherwise this unit-test fails when running with python 3.7, not sure why though. But later version of python (or tqdm) does not have this issue.

lib/ramble/ramble/spack_runner.py
douglasjacobsen: I feel like I recently fixed this as well, but it might be in the package manager stuff. So we might get a conflict with this when we merge that in but it should be easy to correct.

douglasjacobsen (approved):

PR #529, opened 2024-06-12
 Update gid parsing to only accept number/ids

rfbgo:


var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
```diff
@@ -71,7 +71,9 @@ def _prepare_analysis(self, workspace):
         if os.path.isfile(file_name):
             with open(file_name, "r") as f:
                 for cur_id in f.readlines():
-                    ids.add(cur_id.strip())
+                    cur_id = cur_id.strip()
+                    if cur_id.isnumeric():
```
linsword13, line 75: Do we want to put a placeholder for the erroneous ones? I may get confused if I have `n_nodes = 4` but I only get a list of 2 vm_ids?

rfbgo: I think thats a good idea, but the intent of the change here is to filter things out like ssh warnings (ie we read more lines than we expect). In the known failure cases this will still give us 4 vm id's unless the requests fully fail

I think its hard to know which ones are the erroneous ones, but am very open to ideas. I'm not sure the modifier even knows how many results to expect..

linsword13: Oh I see what you mean, so the ids would still be included (assuming the output is something such that the id is still included, just that extra stuff got included in the log file, like "gibberish\n1234\n".) LGTM then.

linsword13 (approved):

douglasjacobsen (approved):

PR #531, opened 2024-06-14
consider `$RAMBLE_PYTHON` environment variable in `ramble` script to pick `python` command to use

boegel:
On some systems multiple `python*` commands are available, like `python3.9` on a RHEL8 system next to the standard system Python `python3` which is equivalent to `python3.6`.

If `ramble` picks up on an environment variable like `$RAMBLE_PYTHON`, the user can control which `python*` command is used to run Ramble...

### DIFF
--- a/bin/ramble
+++ b/bin/ramble
@@ -14,8 +14,8 @@
 # Following line is a shell no-op, and starts a multi-line Python comment.
 # See https://stackoverflow.com/a/47886254
 """:"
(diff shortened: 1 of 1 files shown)


rfbgo:
Closing in favor of discussion of #537, since a few more decisions need to be made before this is mergable 

Thanks for the nice idea, and feel free to continue the discussion there

PR #535, opened 2024-06-20
Package manager abstraction

douglasjacobsen:
This merge is a major enhancement to ramble workspaces. Primarily this adds a `variants` configuration section with the initial option for `package_manager`.

Package manager objects are added to ramble object repositories. These encapsulate the behaviors that package managers should use to modify experiments.

This also adds a `software` configuration section which will replace the (now deprecated) `spack` configuration section.


lib/ramble/ramble/config.py
```diff
@@ -84,9 +87,12 @@
     "modifier_repos": ramble.schema.modifier_repos.schema,
     "modifiers": ramble.schema.modifiers.schema,
     "spack": ramble.schema.spack.schema,
+    "software": ramble.schema.software.schema,
     "success_criteria": ramble.schema.success_criteria.schema,
+    "package_manager_repos": ramble.schema.package_manager_repos.schema,
     "applications": ramble.schema.applications.schema,
     "variables": ramble.schema.variables.schema,
+    "variants": ramble.schema.variants.schema,
```
linsword13, line 95: Would it make sense to make `ramble:variants:package_manager` to default to `spack` instead of none, given that that was the de facto existing behavior?

douglasjacobsen: I think given we have some apps that strictly don't work with spack, I'm hesitant to make it the default.

lib/ramble/llnl/util/tty/__init__.py
```diff
@@ -324,6 +324,7 @@ def get_yes_or_no(prompt, **kwargs):
     result = None
     while result is None:
         msg(prompt, newline=False)
+        ans = input().lower()
```
rfbgo: It feels a bit odd to me to partially update this file when compared to upstream. Is this us selectively picking up a bug fix, or are we moving to some intermediate hash?

I don't think I have a better suggestion, and I know long term we want to depend fully on the spack version.. so we can probably ignore this but I was suprised to see it

lib/ramble/ramble/schema/merged.py
```diff
@@ -17,9 +17,11 @@
 import ramble.schema.config
 import ramble.schema.formatted_executables
 import ramble.schema.repos
-import ramble.schema.spack
+import ramble.schema.spack  # DEPRECATED: Remove when spack is removed
```
rfbgo, line 21: nit: do we need this comment? Is black smart enough to remind us to do this? 

(but in general I would rather have too many times than not enough, so happy to leave it more looking forward / curious about the pattern in general)

lib/ramble/ramble/config.py
linsword13: I see, makes sense.

lib/ramble/ramble/repository.py
```diff
@@ -117,9 +125,28 @@ def _mods(repo_dirs=None):
     return path
 
 
+def _package_managers(repo_dirs=None):
+    """Get the singleton RepoPath instance for Ramble.
+
+    Create a RepoPath, add it to sys.meta_path, and return it.
+
+    TODO: consider not making this a singleton.
```
rfbgo, line 133: Do we still need this? I assume this means we decided against it

rfbgo: Ah I see this is from the other variants, so it's fine. Maybe we don't need it in all places though

lib/ramble/ramble/schema/spack.py
```diff
@@ -69,3 +77,37 @@
...
     "properties": properties,
 }
+
+
+def update(data):
+
+    changed = False
+
+    """
+    This entire section is deprecated.
```
rfbgo: What does this mean? Can we more it more clearly somehow?

lib/ramble/ramble/test/cmd/config.py
```diff
@@ -148,43 +148,45 @@ def test_merged_env_vars_section(mock_low_high_config):
     )
 
 
-def test_merged_spack_section(mock_low_high_config):
+# DEPRECATED: Remove `spack` when removed
```
rfbgo, line 151: Would `@deprecation.fail_if_not_removed` be appropriate here?

rfbgo (dismissed):
Few comments/question, but generally LGTM

lib/ramble/llnl/util/tty/__init__.py
douglasjacobsen: I took this from the latest spack (https://github.com/spack/spack/blob/develop/lib/spack/llnl/util/tty/__init__.py#L267) and think it somehow was missing (previously, `ans` wasn't defined in our tty, so getting an answer was just broken).

lib/ramble/ramble/schema/merged.py
douglasjacobsen: I'm not sure if black would remind us. But I think we'd get a python import error if we remove the schema and this stays.

lib/ramble/ramble/repository.py
douglasjacobsen: Yeah, at a minimum the comment tells you the parts you would updated if you wanted to remove the singleton stuff.

lib/ramble/llnl/util/tty/__init__.py
linsword13: And I had this PR last week: https://github.com/GoogleCloudPlatform/ramble/pull/542, so this should already be in develop.

lib/ramble/ramble/schema/spack.py
douglasjacobsen: Ah I meant to remove this. Originally I had hoped I could make the update method convert the spack section to a software section, but config sections don't behave like that. So I'll just remove this comment instead.

lib/ramble/ramble/test/cmd/config.py
douglasjacobsen: No because the deprecation warning isn't on the schema itself, it's on the usage of the schema (so this test wouldn't fail if spack wasn't removed when the version was updated)

rfbgo (approved):

PR #541, opened 2024-06-27
Add coefficient of variance to summary stats

linsword13:
This dimensionless stat can be useful for assessing the relative variation of individual FOMs.

### DIFF
--- a/lib/ramble/ramble/util/stats.py
+++ b/lib/ramble/ramble/util/stats.py
@@ -92,6 +92,27 @@ class StatsStdev(StatsBase):
         return round(statistics.stdev(values), max_decimal_places(values))
 
 
+class StatsCoefficientOfVariation(StatsBase):
+    name = "cv"
+    min_count = 2
+
+    def compute(self, values):
+        mean = statistics.mean(values)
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/util/stats.py
```diff
@@ -110,4 +131,5 @@ def get_unit(self, unit):
     StatsVar(),
     StatsStdev(),
     StatsCountValues(),
+    StatsCoefficientOfVariation(),
```
dapomeroy, line 133: We iterate over all_stats in generating the results, so putting it at the bottom here will put it last in the list of summary FOMs. Do you think it would make sense to move it up one line so it is grouped with variance and stdev, or is better to be separated?

dapomeroy (dismissed):
Looks great! Just one question about ordering of the FOM in results.

lib/ramble/ramble/util/stats.py
linsword13: Ah good call, updated!

dapomeroy (approved):
LGTM

PR #549, opened 2024-07-02
Allow staging inputs in variable defined directories

douglasjacobsen:
This merge is a re-introduction of https://github.com/GoogleCloudPlatform/ramble/pull/534 with a fix to preserve the previous input location as the default.

### DIFF
--- a/lib/ramble/ramble/language/application_language.py
+++ b/lib/ramble/ramble/language/application_language.py
@@ -6,6 +6,7 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
+import os
 import ramble.workload
 import ramble.language.language_base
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/language/application_language.py
```diff
@@ -149,7 +150,7 @@ def input_file(
     name,
     url,
     description,
-    target_dir="{input_name}",
+    target_dir="{workload_input_dir}" + os.sep + "{input_name}",
```
linsword13: Can this use `os.path.join` instead?

douglasjacobsen: Done. :)

linsword13 (approved):

PR #551, opened 2024-07-03
Add the EESSI package manager

douglasjacobsen:
This merge adds a package manager definition for EESSI.

### DIFF
--- a/var/ramble/repos/builtin/package_managers/eessi/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/eessi/package_manager.py
@@ -1,0 +1,45 @@
+# Copyright 2022-2024 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.pkgmankit import *  # noqa: F403
+
+from ramble.pkg_man.builtin.environment_modules import EnvironmentModules
+
+
+class Eessi(EnvironmentModules):
+    """Package manager definition for EESSI
+
+    This package manager uses the European Environment for Scientific Software
+    Installations to stream binaries from an EESSI mirror.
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/package_managers/eessi/package_manager.py
```diff
@@ -0,0 +1,37 @@
...
+    To control the target architecture used, add EESSI_SOFTWARE_SUBDIR_OVERRIDE
+    to the workspace's environment variable definitions.
+    """
+
+    name = "eessi"
+
+    register_builtin("eessi_source")
+
+    def eessi_source(self):
+        return [". /cvmfs/software.eessi.io/versions/2023.06/init/bash"]
```
boegel: Maybe you should parameterize `eessi_source` to accept a `version`, because we'll soon(ish) introduce another EESSI version next to 2023.06?

I would also use `source` insted of `.`, just for readability sake, and I would name it `eessi_init` instead of `eessi_source`.

douglasjacobsen: Parameterizing `eessi_source` is a good idea. How are you planning on passing in the version, would it just change the path, or is it an argument to the init script?

My main issue with changing from `.` to `source` is that `source` isn't compatible with `sh` but `.` is compatible with both sh and bash. I can change it to `eessi_init` though.

boegel: For now, a different EESSI version would be a different path to the `init/bash` script (`2023.06` is the current and only version).

We are working on an `eessi` CLI tool though (see https://github.com/EESSI/eessi-cli). The first version of that is dumb as a doornail, but the longer term idea there is to support something like `eessi init --version 2023.06`. Not sure that's useful in this context though, all `eessi init` does is source the `init/bash` script for you, so you don't need to remember where it's located, and so it can produce useful error messages when the EESSI repository is not mounted yet.

douglasjacobsen: OK, I think this should look better now. Thanks for the suggestions.

In the future I think we could change it to be an `eessi init --version 2023.06`

rfbgo (approved):

PR #562, opened 2024-07-11
Update modifiers which use mpirun to use pdsh

douglasjacobsen:
Previously, the gcp-metadata modifier used mpirun to get per-node metadata. This merge updates the modifier to use pdsh instead.

### DIFF
--- a/lib/ramble/ramble/test/end_to_end/missing_mpi_cmd.py
+++ b/lib/ramble/ramble/test/end_to_end/missing_mpi_cmd.py
@@ -32,10 +32,11 @@ ramble:
     mpi_command: ''
     batch_submit: 'batch_submit {execute_experiment}'
     processes_per_node: '1'
+    hostlist: 'foo'
   applications:
-    hostname:
+    gromacs:
       workloads:
-        local:
+        water_bare:
           experiments:
             multi-node-test:
               variables:
@@ -43,8 +44,6 @@ ramble:
             single-node-test:
               variables:
                 n_nodes: '1'
-  modifiers:
-  - name: gcp-metadata
   spack:
     packages: {}
     environments: {}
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/test/end_to_end/missing_mpi_cmd.py
linsword13, line 1: I would also remove the `gcp-metadata` modifier config, as previously that was used to introduce the `use_mpi=True` part.

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
```diff
@@ -24,6 +27,8 @@ class GcpMetadata(BasicModifier):
     mode("standard", description="Standard execution mode")
     default_mode("standard")
 
+    required_variable("hostlist")
```
linsword13, line 32: The change itself LGTM. My concern is that this is a breaking change, and I am a bit hesitant to merge it before having a good story of showing/supporting how people can provide this variable. (Probably at least update the `examples/slurm_execute_experiment.tpl` and the various example ramble configs.)

lib/ramble/ramble/test/end_to_end/missing_mpi_cmd.py
douglasjacobsen: Good point. I can remove that now.

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
douglasjacobsen: Yeah, this is tricky. Mostly because there is no standard way to create a hostlist, but pdsh wants a hostlist as input.

Slurm happens to set an env-var that can be used as a hostlist, but if you setup a beowulf cluster, there is no standard variable you can refer to.

This required variable wouldn't be defined in a `.tpl` file, and instead needs to be defined in a `ramble.yaml`, but putting it into the examples (which don't use this modifier) assumes a specific workflow (i.e. slurm) which I don't think it's safe to assume.

We can discuss this more before this gets merged though.

linsword13: I definitely agree this is a great change to get in. May not be worth it, but what about having a gcp-metadata-v2 that has this change, while keeping the "legacy" gcp-metadata modifier around for a bit?

Separate question: how much can we assume the presence of `pdsh`?

douglasjacobsen: I think I'd prefer it just be a mode in this modifier rather than a separate modifier entirely. That should be pretty easy to do, however it makes use a bit less straightforward (since the default would probably be the old mode, and we'd still need to figure out what to do about removing the mpirun mode which is technically broken).

I think whether or not we can assume `pdsh` exists is a great question, however the same can be true for `mpirun`. I think we have a few options for this currently:
1. Assume pdsh exists, and let it fail ungracefully if it doesn't (Maybe ensure the modifier docstring says it needs pdsh)
2. Add a required_package for pdsh, and let the package manager install pdsh (might result in two pdsh's)
3. Design / add something to detect pdsh exists at setup time, however this similarly doesn't guarantee it exists on compute nodes / at execution time
4. Add a check into the experiment (at run time) with a FOM to capture if pdsh was found or not.

linsword13 (approved):

PR #563, opened 2024-07-15
Update upload semantics to allow the upload of failed experiments

rfbgo:
Also removes the "always print" flag and makes it on by default

This changes the upload schema, so may be a breaking change for some users


lib/ramble/ramble/config.py
```diff
@@ -141,6 +141,7 @@
         "pip": {"install": {"flags": []}},
         "input_cache": "$ramble/var/ramble/cache",
         "workspace_dirs": "$ramble/var/ramble/workspaces",
+        "upload": {"push_failed": True},
```
douglasjacobsen, line 150: Do we need this? It seems like this allows you to cause failed experiments to not be uploaded, right?

Also, do you have thoughts on how https://github.com/GoogleCloudPlatform/ramble/blob/9eac85ae95408bcde8f8f53a1524f6d8e968f67f/lib/ramble/ramble/config.py#L131 changes (if it does at all? it might not...)

douglasjacobsen (changes requested):
Looks good so far!

I had one question, but I think that's easy to answer. As an aside, it seems like we should have a test that maybe mocks the uploader (and hijacks the output) to make sure the upload logic works.

lib/ramble/ramble/config.py
rfbgo: Yes I think this is needed. I feel we should give users a way to retain the behavior they have today incase they rely on it

douglasjacobsen: Thanks. Do you think the default for `repeat_success_strict` needs to change?

rfbgo: I personally think the default should be `repeat_success_strict=False`, but I think its logically separate from this PR 

Happy to change it if you like though

douglasjacobsen: I feel like it's somewhat intertwined with this PR, just because the semantic of "We're not printing stuff because it failed" is different now.

`False` for this option means we don't print any stats, while `True` means we print whatever stats we can print.

I can't decide how I feel about it, but it seems like `True` is closer to the behavior we're changing to in this PR.

rfbgo: Just to be super clear.. `properties["config"]["repeat_success_strict"] = {"type": "boolean", "default": True}` => you want no change?

rfbgo: We discussed this offline. I will make changes that:

- Leave the default as true
- Change the repeats code to always calculate stats regardless of status
- Honor the value `push_failed` even for sub-experiments of a repeat parent 

In short this means `repeat_success_strict` will now have a single purpose of controlling how status is reported

douglasjacobsen (dismissed):

douglasjacobsen (approved):

PR #571, opened 2024-07-18
Remove `time` from hostname application

douglasjacobsen:
This merge removes the use of `time` within the hostname application definition.

### DIFF
--- a/var/ramble/repos/builtin/applications/hostname/application.py
+++ b/var/ramble/repos/builtin/applications/hostname/application.py
@@ -6,9 +6,7 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
-import os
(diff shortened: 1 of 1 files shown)


rfbgo (commented):
This both breaks mac, and defies are normal redirection semantics. Whats the reason a normal output capture is not sufficient?

douglasjacobsen:
The normal output capture breaks the success criteria / fom capture for this workload.

Honestly, the app definition is broken for time, because time is non-standard across distros. Maybe it's better to just remove time and only capture hostnames rather than a time based FOM though.

douglasjacobsen:
I just changed this to remove the use of `time` and make it a standard "just hostname" application definition.

rfbgo (approved):

PR #572, opened 2024-07-18
Add caching to the expander

douglasjacobsen:
This merge adds a cache to the expander to reduce repeatedly expanding the same string multiple times.


douglasjacobsen:
I don't know that we need this yet, so I'm going to close this. We'll revisit if we want it later.

PR #573, opened 2024-07-18
Add a `run_in_background` option for executables

linsword13:


lib/ramble/ramble/application.py
```diff
@@ -1060,8 +1060,13 @@ def _define_commands(
                         if ramble.config.get("config:shell") in ["bash", "sh"]:
                             redirect += " 2>&1"
 
+                    if cmd_conf.run_in_background:
+                        bg_cmd = " &"
```
douglasjacobsen, line 1064: Do you know if this is the same method of running a process in the background for other shells? Should this be at least guarded with a check that we're on sh / bash / csh?

douglasjacobsen (changes requested):
One small question.

As an aside, is it useful to capture the PID of the backgrounded process, so we can do something like sleep until it's finished?

lib/ramble/ramble/application.py
linsword13: Good question. It seems this ampersand actually has pretty good support across shells ((t)csh, fish both support it, even powershell starting from version 6 has it: https://screenshot.googleplex.com/33qLeDJjhwFvVQr. The only one that's missing seems to be bat, but not sure how much we care about that ;)

linsword13:
> One small question.
> 
> As an aside, is it useful to capture the PID of the backgrounded process, so we can do something like sleep until it's finished?

Yeah good call. I was thinking to have a modifier to achieve that (using an `executable_modifier` to capture the PID,) and then do a wait-for-all at the end. I feel that might be more flexible. Will add such a modifier in this PR.

linsword13:
> > One small question.
> > As an aside, is it useful to capture the PID of the backgrounded process, so we can do something like sleep until it's finished?
> 
> Yeah good call. I was thinking to have a modifier to achieve that (using an `executable_modifier` to capture the PID,) and then do a wait-for-all at the end. I feel that might be more flexible. Will add such a modifier in this PR.

Added a modifier doing the wait-for-completion for background jobs.

var/ramble/repos/builtin/modifiers/wait-for-bg-jobs/modifier.py
```diff
@@ -0,0 +1,58 @@
...
+from ramble.modkit import *
+import ramble.config
+from ramble.util.executable import CommandExecutable
+from ramble.util.shell_vars import last_pid_var
+
+
+class WaitForBgJobs(BasicModifier):
+    """Define a modifier for waiting for background jobs.
+
+    This only waits for background jobs started from the experiment.
```
douglasjacobsen, line 19: This only works for background jobs in bash, right? If so, I think that's fine, but we should add a comment in the docstring that this only works for bash just so people know. We could also make the builtins / modifications conditional on the shell being bash (since I don't think sh has arrays, but I could be wrong on that)

douglasjacobsen (changes requested):
Thanks for making the changes! This is awesome.

One small comment, otherwise it LGTM. Feel free to choose between adding the conditional check or adding a comment in the doc string.

var/ramble/repos/builtin/modifiers/wait-for-bg-jobs/modifier.py
linsword13: That's a good catch, indeed this seems to be bash only. (Other newer shells like fish has array support, but the syntax seems different.) Added the check (instead of skipping it will error out when other shells are detected.)

douglasjacobsen (approved):

PR #576, opened 2024-07-20
Deploy base classes that are in an object repo

linsword13:
Deploy base classes that are in an object repo
Also add in support for deploying package_manager, as well as all the
base_* objects.

Here's an example [deployment](https://pantheon.corp.google.com/storage/browser/ramble-deployment-lin/test-deploy1).

### DIFF
--- a/lib/ramble/ramble/pipeline.py
+++ b/lib/ramble/ramble/pipeline.py
@@ -575,7 +575,14 @@ class PushDeploymentPipeline(Pipeline):
         fs.mkdirp(aux_software_dir)
         aux_repo_conf = os.path.join(aux_software_dir, "repos.yaml")
 
-        repo_conf_defs = [("repos", "repos.yaml"), ("modifier_repos", "modifier_repos.yaml")]
+        repo_conf_defs = [
+            ("repos", "repos.yaml"),
+            ("modifier_repos", "modifier_repos.yaml"),
+            ("package_manager_repos", "package_manager_repos.yaml"),
+            ("base_application_repos", "base_application_repos.yaml"),
+            ("base_modifier_repos", "base_modifier_repos.yaml"),
+            ("base_package_manager_repos", "base_package_manager_repos.yaml"),
+        ]
 
         for repo_conf in repo_conf_defs:
             aux_repo_conf = os.path.join(aux_software_dir, repo_conf[1])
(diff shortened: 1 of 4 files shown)


douglasjacobsen:
This is a great idea. We might want to wait until https://github.com/GoogleCloudPlatform/ramble/pull/575 is merged, and then deploy the base objects too.

And maybe we should update the deployment command to capture all base objects it can find. But that can be a later task.

linsword13:
> This is a great idea. We might want to wait until #575 is merged, and then deploy the base objects too.
> 
> And maybe we should update the deployment command to capture all base objects it can find. But that can be a later task.

Thanks. Looks like that PR just got merged. Will pull that in.

As I am at it, I will also add the deployment of all base objects it can find.

linsword13:
> > This is a great idea. We might want to wait until #575 is merged, and then deploy the base objects too.
> > And maybe we should update the deployment command to capture all base objects it can find. But that can be a later task.
> 
> Thanks. Looks like that PR just got merged. Will pull that in.
> 
> As I am at it, I will also add the deployment of all base objects it can find.

Added in the deployment of inherited classes.

lib/ramble/ramble/pipeline.py
```diff
@@ -575,7 +575,14 @@ def _execute(self):
         fs.mkdirp(aux_software_dir)
         aux_repo_conf = os.path.join(aux_software_dir, "repos.yaml")
 
-        repo_conf_defs = [("repos", "repos.yaml"), ("modifier_repos", "modifier_repos.yaml")]
+        repo_conf_defs = [
```
douglasjacobsen, line 579: Can we read these from repository.py instead of replicating them? (i.e. https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/repository.py#L82 + `'.yaml'`

lib/ramble/ramble/pipeline.py
```diff
@@ -597,11 +604,26 @@ def _execute(self):
                 f.write(syaml.dump_config(repo_data))
 
         repo_path = os.path.join(self.workspace.named_deployment, self.object_repo_name)
-        object_types = ["applications", "modifiers", "packages"]
+        object_types = [
```
douglasjacobsen: Similar comment as before, but for this: https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/repository.py#L59

lib/ramble/ramble/pipeline.py
```diff
@@ -597,11 +604,26 @@ def _execute(self):
...
+            "package_managers",
+            "base_applications",
+            "base_modifiers",
+            "base_package_managers",
+        ]
         for object_type in object_types:
             fs.mkdirp(os.path.join(repo_path, object_type))
 
-        for conf_file in ["repo.yaml", "modifier_repo.yaml"]:
+        for conf_file in [
```
douglasjacobsen: Same comment about the config sections.

lib/ramble/ramble/pipeline.py
```diff
@@ -597,11 +604,26 @@ def _execute(self):
...
+            "repo.yaml",
+            "modifier_repo.yaml",
+            "package_manager_repo.yaml",
+            "base_application_repo.yaml",
+            "base_modifier_repo.yaml",
+            "base_package_manager_repo.yaml",
+        ]:
             with open(os.path.join(repo_path, conf_file), "w+") as f:
                 f.write("repo:\n")
```
douglasjacobsen: If we are writing different config files for each (we don't need to) then the top level attribute should be the same as the config section name (i.e. filename sans `'.yaml'`)

douglasjacobsen (changes requested):

lib/ramble/ramble/pipeline.py
linsword13: Good call, done.

lib/ramble/ramble/pipeline.py
linsword13: Good point, I updated to use the `dir_name` from the type_definitions.

lib/ramble/ramble/pipeline.py
linsword13: I see, changed to only write out the unified config.

douglasjacobsen (approved):

PR #578, opened 2024-07-23
Update output capture semantics to not always capture stderr

rfbgo:
This PR makes the semantics around output capture more consistent, moves the logic to a single place, and allows ramble to not capture stdout/stderr when requested (which currently does not work as expected)

This is a more complex version of #577

### DIFF
--- a/lib/ramble/ramble/schema/types.py
+++ b/lib/ramble/ramble/schema/types.py
@@ -8,12 +8,31 @@
 
 """Base types for building schema files from"""
 
+from enum import Enum
 
-class OUTPUT_CAPTURE:
-    STDERR = "2>"
-    STDOUT = ">>"
-    ALL = "&>"
-    DEFAULT = STDOUT
+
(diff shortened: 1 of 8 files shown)


lib/ramble/ramble/schema/types.py
```diff
@@ -8,12 +8,31 @@
...
 
+from enum import Enum
 
-class OUTPUT_CAPTURE:
-    STDERR = "2>"
-    STDOUT = ">>"
-    ALL = "&>"
-    DEFAULT = STDOUT
+
+class OUTPUT_CAPTURE(Enum):
```
douglasjacobsen: I know this came from `schema.types` but would it still work if this was moved into `ramble.util` ?

rfbgo: I can't see why not, let me try

rfbgo: Done

douglasjacobsen (approved):

PR #587, opened 2024-07-30
Centralize shell support utils

linsword13:
* Put a few shell utils into a common util file
* Add a `target_shells` directive for ramble objects, to declare explicit supported set of shells

Also a unrelated change to ensure license file checks don't complain on removed files.


lib/ramble/ramble/application.py
```diff
@@ -72,6 +72,23 @@ def _get_context_display_name(context):
     )
 
 
+def _check_shell_support(app_inst):
```
douglasjacobsen, line 75: Should this also apply to modifiers and package managers?

Either in their base classes, or in the application class where the instances for the other objects are created?

lib/ramble/ramble/language/shared_language.py
```diff
@@ -415,3 +415,21 @@ def _execute_tag(obj):
...
+    If not specified, i.e., not directly specified or inherited from the base,
+    then it assumes all shells are supported.
+
+    Args:
+        shell_support_pattern: The glob pattern that is used to match with the configured shell
+    """
+
+    def _execute_target_shells(obj):
+        if shell_support_pattern is not None:
+            obj.shell_support_pattern = shell_support_pattern
```
douglasjacobsen, line 433: Odd question, but does this just create an instance attribute instead of a class attribute?

douglasjacobsen (changes requested):

lib/ramble/ramble/application.py
douglasjacobsen: Ah, never mind, sorry. I see that this handles checking the other objects too and that they aren't responsible for making sure they are valid

douglasjacobsen:
Look good! There is a conflict, and I had one question.

lib/ramble/ramble/language/shared_language.py
linsword13: Right now it's set as a class attribute, as it does not use the `dicts` argument to add into `_directive_names`. I think that's what I want as this shouldn't have per-instance modification, but maybe I miss something?

linsword13:
> Look good! There is a conflict, and I had one question.

Thanks. Resolved the conflict. I will also hold off submitting this and wait for your info PR to get in. Then I will add support for this new attribute in this PR.

linsword13:
> > Look good! There is a conflict, and I had one question.
> 
> Thanks. Resolved the conflict. I will also hold off submitting this and wait for your info PR to get in. Then I will add support for this new attribute in this PR.

Updated with the newest develop.

douglasjacobsen (approved):

PR #589, opened 2024-07-30
Update the info command to be more flexible

douglasjacobsen:
Introduces filtering, verbose mode, and formats for the object info command.

### DIFF
--- a/lib/ramble/ramble/cmd/common/info.py
+++ b/lib/ramble/ramble/cmd/common/info.py
@@ -102,8 +102,10 @@ def setup_info_parser(subparser):
     )
     attributes.add_argument(
         "--attributes",
+        "--attrs",
+        dest="attributes",
         default="",
-        help="comma delimited list of attributes to inspect for object",
+        help="comma separated list of attributes to inspect for object",
     )
 
 
--- a/lib/ramble/ramble/cmd/common/info.py
+++ b/lib/ramble/ramble/cmd/common/info.py
@@ -6,33 +6,305 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
+import fnmatch
 
 import llnl.util.tty.color as color
+from llnl.util.tty.colify import colified
 
+import ramble.util.colors
 import ramble.cmd.common.arguments as arguments
 import ramble.repository
 
-header_color = "@*b"
-plain_format = "@."
+from ramble.util.logger import logger
+
+import enum
+
+supported_formats = enum.Enum("formats", ["text", "lists"])
+
+obj_attribute_map = {
+    "maintainers": None,
+    "tags": None,
+    "pipelines": "_pipelines",
+    "figure_of_merit_contexts": None,
+    "figures_of_merit": None,
+    "builtins": None,
+    "package_manager_configs": None,
+    "required_packages": None,
+    "compilers": None,
+    "software_specs": None,
+    "archive_patterns": None,
+    "success_criteria": None,
+    # Application specific:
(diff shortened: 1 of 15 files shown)


lib/ramble/ramble/cmd/common/info.py
```diff
@@ -6,34 +6,306 @@
...
 
-header_color = "@*b"
-plain_format = "@."
+from ramble.util.logger import logger
+
+import enum
+
+supported_formats = enum.Enum("formats", ["text", "lists"])
+
+obj_attribute_map = {
```
linsword13, line 24: I think this is totally fine, just a random thought (maybe for future):

Would it be better if these are something provided by the object itself? Can also streamline it a bit for instance by doing auto registration via the directive decorator, etc. Otherwise it might be getting out-of-sync when new things are added on the object side.

lib/ramble/ramble/cmd/common/info.py
```diff
@@ -6,34 +6,306 @@
...
+    level.add_argument(
+        "--verbose", "-v", action="store_true", help="print full information about attributes"
+    )
+
+    attributes = subparser.add_mutually_exclusive_group()
+    attributes.add_argument(
+        "--all", action="store_true", default=True, help="print all object attributes"
+    )
+    attributes.add_argument(
+        "--attributes",
```
linsword13, line 104: Haha, being lazy, how about also having a shortcut like `--attrs`?

linsword13 (commented):
Great stuff!

linsword13 (commented):
Probably also need to update some info outputs of the tutorials?

lib/ramble/ramble/cmd/common/info.py
douglasjacobsen: I actually thought about doing that, but decided against it for the time being. Mostly because some of the attributes (like tags or default_mode) don't really know their actual attribute name via the directive.

This can be seen primarily with the default mode directive, where the attribute we modify is actually `_default_usage_mode` instead of `default_mode`. (Also using `default_mode` as the attribute name breaks the function map of directive names).

I think this is definitely a place that's ripe for cleanup, but I'm not sure what the best way to do this is currently.

lib/ramble/ramble/cmd/common/info.py
douglasjacobsen: Added both `--attrs` and `-a`. :)

douglasjacobsen:
@linsword13 Good call. I just pushed an update to the tutorials.

lib/ramble/ramble/cmd/common/info.py
linsword13: Makes sense, I guess one possible approach is to enhance the directive to allow passing in the attr name (a bit overlap with the current `dict` argument.) But yeah the current way is totally fine.

lib/ramble/ramble/cmd/common/info.py
linsword13: I thought about recommending `-a`, but then that seems to somewhat conflate with `--all`, and also that doesn't express the "plural" nature. WDYT?

linsword13 (dismissed):

lib/ramble/ramble/cmd/common/info.py
douglasjacobsen: Ah sure, those are good points. I can pull `-a` back out.

linsword13 (dismissed):

linsword13 (approved):

PR #595, opened 2024-08-05
Add mirror support check

linsword13:
With this change, package managers that don't implement the `mirror_software` phase will error out when execute `workspace mirror`

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -1224,6 +1224,21 @@ class ApplicationBase(metaclass=ApplicationMeta):
                 }
         self._input_fetchers = inputs
 
+    register_phase("check_mirror_support", pipeline="mirror", run_before=["mirror_inputs"])
+
+    def _check_mirror_support(self, workspace, app_inst=None):
+        """Check if the given package manager supports mirroring"""
+        pkgman = self.package_manager
+        if pkgman is not None:
+            # Use the presence of a `mirror_software` phase as an indicator.
+            # This is not very robust, a better approach may be to claim support
+            # from individual package managers.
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/application.py
```diff
@@ -1224,6 +1224,21 @@ def _inputs_and_fetchers(self, workload=None):
...
+    def _check_mirror_support(self, workspace, app_inst=None):
+        """Check if the given package manager supports mirroring"""
+        pkgman = self.package_manager
+        if pkgman is not None:
+            # Use the presence of a `mirror_software` phase as an indicator.
+            # This is not very robust, a better approach may be to claim support
+            # from individual package managers.
+            mirror_phases = self.get_pipeline_phases("mirror")
+            if "mirror_software" not in mirror_phases:
+                logger.die(
```
douglasjacobsen, line 1241: Does this break mirroring inputs only? If so, I think we should not die here, and maybe change to only skip trying to mirror software.

(i.e. it's fine if we only mirror inputs during the mirror pipeline)

douglasjacobsen (commented):
One question, just to make sure I understand.

lib/ramble/ramble/application.py
linsword13: Oh very good catch, initially I thought since this check is itself a phase, an invocation of `--phases mirror-inputs` would just work since this check phase doesn't get selected, but then indeed I forgot the `--include-phase-dependencies` option, which would bring in this check phase if specified.

Updated to not error out in such cases. (The code does seem to be somewhat more convoluted now, so definitely open for better ideas.)

PR #601, opened 2024-08-08
Fix spacing for info output

douglasjacobsen:
Previously, the output for info on modifiers had incorrect spacing around the `default_mode` attribute. This merge fixes the issue.


douglasjacobsen:
To test, you can look at the output before and after this PR with this command: `ramble info --type modifiers install-spack`

rfbgo (approved):

PR #607, opened 2024-08-10
Remove `dry-run` option for workspace analyze

linsword13:
Also add in a `PrintOnly` uploader as a way to test out uploader functionality.

Samle invocation:

```
ramble -c "config:upload:type:PrintOnly" -c "config:upload:uri:fake-db.all_data" workspace analyze --upload
```

### DIFF
--- a/lib/ramble/ramble/experimental/uploader.py
+++ b/lib/ramble/ramble/experimental/uploader.py
@@ -9,6 +9,7 @@
 import json
 import sys
 import math
+from enum import Enum
 
 import ramble.config
 from ramble.util.logger import logger
@@ -16,6 +17,8 @@ from ramble.config import ConfigError
 
 
 default_node_type_val = "Not Specified"
+
+uploader_types = Enum("uploader_types", ["BigQuery", "PrintOnly"])
 
 
 class Uploader:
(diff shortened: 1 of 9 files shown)


lib/ramble/ramble/experimental/uploader.py
```diff
@@ -304,3 +312,14 @@ def get_experiment_id(experiment):
...
+
+
+class PrintOnlyUploader(Uploader):
+    """An uploader that only prints out formatted data without actually uploading."""
+
+    def perform_upload(self, uri, results):
+        super().perform_upload(uri, results)
+        exp_table_id, exps_to_insert, fom_table_id, foms_to_insert = _prepare_data(results, uri)
+        logger.info("NOTE: The PrintOnly uploader only logs, but does not upload any data.")
+        logger.info(f"Experiments {exps_to_insert} would be uploaded to table {exp_table_id}")
```
douglasjacobsen: This would print the whole list of experiments and FOMs right? Is it possible to make it easier to consume this info (like maybe print the length here or at the end, and then print the whole list as one entry per line) ?

I don't know if that's the right choice for this, but it seems like the list being printed in the string would make it hard to connect the different parts.

douglasjacobsen (commented):
This is great. One comment, which could be a request for a change. But if you don't think it's worth the effort that's fine too.

lib/ramble/ramble/experimental/uploader.py
linsword13: Great call, updated.

douglasjacobsen (approved):

PR #612, opened 2024-08-15
Allow gromacs to support additional args, and use workload groups

rfbgo:


linsword13 (dismissed):
LGTM (may need to do `ramble style -f` though)

douglasjacobsen (approved):

PR #621, opened 2024-08-22
Add a command for generating new workspace configuration files

douglasjacobsen:
The primary purpose of this merge is to define a new front-end command (`ramble workspace generate-config`) which can be used to generate new workspace config files (or add new experiments to existing workspace config files).

The known_applications test is updated to use this to generate the test configs rather than through string manipulation.

Additionally, this merge updates some of the semantics around which variables are required, reserved, and optional.

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -966,7 +966,7 @@ def workspace_generate_config_setup_parser(subparser):
         "--wf",
         dest="workload_filters",
         action="append",
-        help="filter to use when selecting workloads in the application. "
+        help="glob filter to use when selecting workloads in the application. "
         + "Workload is kept if it matches any filter.",
     )
 
@@ -975,7 +975,7 @@ def workspace_generate_config_setup_parser(subparser):
         "--vf",
         dest="variable_filters",
         action="append",
-        help="filter to use when selecting variables in the workloads. "
+        help="glob filter to use when selecting variables in the workloads. "
         + "Variable is kept if it matches any filter.",
     )
 
@@ -1011,6 +1011,13 @@ def workspace_generate_config_setup_parser(subparser):
         action="store_true",
         help="perform a dry run. Print resulting config to screen and not "
         + "to the workspace configuration file",
+    )
+
+    subparser.add_argument(
+        "--overwrite",
+        dest="overwrite",
+        action="store_true",
+        help="overwrite existing definitions with newly generated definitions",
     )
 
     variable_control = subparser.add_mutually_exclusive_group()
@@ -1091,6 +1098,7 @@ def workspace_generate_config(args):
         args.package_manager,
         zips,
         matrix,
+        args.overwrite,
     )
 
 
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -1006,6 +1006,7 @@ def workspace_generate_config_setup_parser(subparser):
 
     subparser.add_argument(
         "--dry-run",
+        "--print",
         dest="dry_run",
         action="store_true",
         help="perform a dry run. Print resulting config to screen and not "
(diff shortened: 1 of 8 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -955,6 +956,134 @@ def workspace_mirror(args):
...
+    subparser.add_argument(
+        "--package-manager",
+        "-p",
+        dest="package_manager",
+        default=None,
+        help="name of (optional) package to define within the experiment scope",
+    )
+
+    subparser.add_argument(
+        "--dry-run",
```
rfbgo, line 1008: I really like the `dry-run`, and how its consistent with other commands, but should we alias alias this to `print` or something?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -955,6 +956,134 @@ def workspace_mirror(args):
     pipeline.run()
 
 
+def workspace_generate_config_setup_parser(subparser):
```
rfbgo, line 959: I think workspace is the right place for this, but do we want to support running this "outside" of a workspace? It somewhat does not make sense.. but I would also use this (as dry-run) locally to generate a config to paste to a remote machine

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -955,6 +956,134 @@ def workspace_mirror(args):
...
+    """generate current workspace config"""
+
+    arguments.add_common_arguments(subparser, ["application"])
+
+    subparser.add_argument(
+        "--workload-filter",
+        "--wf",
+        dest="workload_filters",
+        action="append",
+        help="filter to use when selecting workloads in the application. "
```
rfbgo: Is this sufficiently obvious that this is a regex not some sort of auto-glob? I can't decide if "filter" has a strict enough definition.

Based on reading only this I half expected that `--wf bare` might match `waterbare` but from the impl it's fairly clear it needs `*bare` or similar

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -955,6 +956,134 @@ def workspace_mirror(args):
...
+    subparser.add_argument(
+        "--experiment-name",
+        "-e",
+        dest="experiment_name",
+        default="generated",
+        help="name of generated experiment",
+    )
+
+    subparser.add_argument(
+        "--package-manager",
```
rfbgo, line 1000: I got into trouble with this in testing where I wanted to update the package manager for a config I had just generated. I think we should either (or maybe both):

1. Warn when a generated experiment would collide so we refuse to emit it; and
2. Perhaps allow a `--force` option that has a semantic like "I know this will collide and overwrite config, but I'm sure I want it"

rfbgo (changes requested):
Awesome feature, I'm using it already :)

Few questions and suggestions around the CUJs

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Sure, let me add that.

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: In this case it is just a glob. I added a bit to the help string though. LMK if that looks better.

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Done!

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Done!

lib/ramble/ramble/cmd/workspace.py
rfbgo: Looks good, thanks!

rfbgo (approved):

PR #627, opened 2024-08-29
Refactor app result into a class

linsword13:
For now when pass the result into workspace, it still uses the old dict format so that encoders and uploaders work without needing to change. That also ensures existing serialized results do not break.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -53,6 +53,7 @@ from ramble.util.shell_utils import source_str
 from ramble.util.naming import NS_SEPARATOR
 
 from ramble.workspace import namespace
+from ramble.experiment_result import ExperimentResult
 
 from ramble.language.application_language import ApplicationMeta
 from ramble.language.shared_language import SharedMeta, register_builtin, register_phase
@@ -144,7 +145,7 @@ class ApplicationBase(metaclass=ApplicationMeta):
         self.experiment_tags = []
         self._modifier_instances = []
         self._input_fetchers = None
-        self.results = {}
+        self.result = None
         self._phase_times = {}
         self._pipeline_graphs = None
         self.package_manager = None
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/application.py
```diff
@@ -1866,11 +1851,11 @@ def is_numeric(value):
                 continue
 
             # When strict success is off for repeats (loose success), skip failed exps
-            if exp_inst.results["RAMBLE_STATUS"] == "FAILED":
+            if exp_inst.result.status == "FAILED":
```
douglasjacobsen: (This is unrelated to this PR directly) Can we change "FAILED" to `experiment_status.FAILED` ?

lib/ramble/ramble/experiment_result.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+
+_DICT_MAPPING = {
+    "name": "name",
```
douglasjacobsen, line 13: Is it possible to use the namespace attributes where they make sense here?

douglasjacobsen (commented):
Two small changes, but otherwise LGTM

lib/ramble/ramble/application.py
linsword13: Good call, done

lib/ramble/ramble/experiment_result.py
linsword13: Definitely, that's a great suggestion, done.

douglasjacobsen (approved):

PR #634, opened 2024-09-06
Add support for arbitrary commands in formatted executables

douglasjacobsen:
This merge allows formatted_executable definitions to have a `commands` attribute, which can define the lines that should be in the formatted output.

The default is to format all of the commands for the experiment, but this can be override to format arbitrary strings / variables. Item in the `commands` list will be expanded as a variable definition, and then split over new-lines before formatting based on the formatted executable definition.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -1138,6 +1138,8 @@ class ApplicationBase(metaclass=ApplicationMeta):
         based on the formatting requested.
         """
 
+        self.variables[self.keywords.unformatted_command] = "\n".join(self._command_list)
+
         for var_name, formatted_conf in self._formatted_executables.items():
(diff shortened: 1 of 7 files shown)


lib/ramble/ramble/application.py
```diff
@@ -1138,6 +1138,14 @@ def _define_formatted_executables(self):
         based on the formatting requested.
         """
 
+        unformatted_str = ""
+        for cmd in self._command_list:
+            if unformatted_str:
```
linsword13: How about accumulating `cmd` in a list, and then `"\n".join(cmds)` at the end? (similarly for the formatted cmd below)

douglasjacobsen: Done! :) Thanks

linsword13 (approved):

PR #635, opened 2024-09-09
Allow specifying `spack:concretizer` settings in Ramble config

linsword13:
This is useful for enabling cross-compilation when using Ramble on a per-ramble-workspace basis.

Example Ramble config snip:

```
ramble:
  variants:
    package_manager: spack
  config:
    spack:
      concretizer:
        targets:
          host_compatible: false
  ...
```

### DIFF
--- a/lib/ramble/ramble/schema/config.py
+++ b/lib/ramble/ramble/schema/config.py
@@ -13,7 +13,9 @@
 """
 
 import spack.schema.config
-import spack.schema.concretizer
+
+import ramble.schema.util.import_util as import_util
+
 
 #: Properties for inclusion in other schemas
 properties = {
@@ -22,6 +24,9 @@ properties = {
 
 properties["config"]["shell"] = {"type": "string", "enum": ["sh", "bash", "csh", "tcsh", "fish"]}
 
+# Get the concretizer schema from the external spack instead of the one included in ramble src.
+# This aligns since the concretizer settings are fed into the external spack instance.
+spack_concretizer = import_util.import_external_spack_schema("spack.schema.concretizer")
 properties["config"]["spack"] = {
     "type": "object",
     "default": {"install": {"flags": "--reuse"}, "concretize": {"flags": "--reuse"}},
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/schema/config.py
```diff
@@ -103,6 +104,7 @@
             },
             "additionalProperties": False,
         },
+        "concretizer": spack.schema.concretizer.properties["concretizer"],
```
douglasjacobsen: I think in theory this is good, but in practice it's more complicated.

This implies that the schema we should use is the `ramble internal` schema, which is not the same as the spack schema that we would actually be running the spack commands with.

We should probably think about this some more.

douglasjacobsen (commented):
This is a cool idea! I'm worried about the schema use, so we should see if there's a way to get the schema from external spack instead of internal spack.

lib/ramble/ramble/schema/config.py
linsword13: Ah that's a very good point, ack.

It feels challenging to pull in the external spack schema (we can maybe do it in setup-env.sh somehow, but that still doesn't prevent it from changing via some other means.)

I guess one option is to define this concretizer schema very loosely ourselves (like only define `unify`, but allow additional properties.) That way we open this up and let user specify pretty much whatever they want.

Another tighter option may be to provide a higher-level Ramble config (say `enable-cross-compile`,) and then individual package manager decides on how the enablement works respectively.

douglasjacobsen: Well, I think we could potentially pull it in through some loader magic, and the path relative to the spack binary (which we know).

The main question is, is it worth it? It helps us catch errors earlier, so we can validate what we're writing / using against the correct schema. And it should allow it to change ad-hoc as well.

Is there a good reason to not do it that way, or try to get the actual schema file out?

linsword13: Great advice on the import suggestion. I added a simplistic version of loading this particular concretizer schema.

The auxiliary software approach indeed works, so I don't technically need this PR anymore. I think the only remaining benefit is that the setting lives inside the ramble.yaml file, so the config is less spread-out. Happy to close this PR though if the approach taken feels too kludgy.

douglasjacobsen: Awesome, thanks for the update.

I think given we don't need this anymore, and it would only support spack, we should try to design a more robust solution that can help with other package managers that need additional configuration as well.

So, lets close this for now and make a task for tracking the effort internally.

PR #636, opened 2024-09-09
Update include path support

douglasjacobsen:
This merge accomplishes 3 main goals:
- Remove support for `application_directory` within workspace configuration files
- Enhance support for `include` files / paths in workspace configuration files
- Update documentation and tests to ensure includes work as expected.

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -590,26 +590,6 @@ class Workspace:
                 template_name = workspace_execution_template[0:-ext_len]
                 self._read_template(template_name, template_execute_script)
 
-            self._read_all_application_configs()
-
-    def _read_all_application_configs(self):
-        path_replacements = {"workspace": self.root, "workspace_config": self.config_dir}
-        ramble_dict = self._get_workspace_dict()[namespace.ramble]
-        if namespace.application_dir in ramble_dict:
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1639,6 +1619,24 @@ def all_auxiliary_software_files(self):
         """Iterator over each file in $workspace/configs/auxiliary_software_files"""
         yield from self._auxiliary_software_files.items()
 
+    def workspace_paths(self):
+        """Dictionary of path replacements for workspace"""
+        if not hasattr(self, "_workspace_path_replacements"):
+            self._workspace_path_replacements = {
```
linsword13, line 1625: Do we want to document these keywords?

linsword13 (commented):
Great stuff! Want to confirm as you mentioned you updated tests, but I don't see changes there?

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: Thanks, I did mean to do that but forgot to.

douglasjacobsen (commented):
Ah! Good catch. I didn't push the last commit which added the test.

linsword13 (approved):

PR #642, opened 2024-09-11
Fix '--where' experiment index filter on info

dapomeroy:
Fixes a bug where filtering using '--where' on the experiment index does not work correctly

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -662,6 +662,18 @@ def workspace_info(args):
                 print_experiment_set.set_experiment_context(experiment_context)
                 print_experiment_set.build_experiment_chains()
 
+                # Reindex the experiments in the print set to match the overall set
+                for exp_name, _, _ in print_experiment_set.all_experiments():
+                    app_inst = experiment_set.get_experiment(exp_name)
+                    experiment_index = app_inst.expander.expand_var_name(
+                        app_inst.keywords.experiment_index
+                    )
+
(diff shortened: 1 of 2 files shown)


linsword13 (commented):
Good stuff! Wonder if we can add a test for it?

linsword13 (approved):

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -662,6 +662,18 @@ def workspace_info(args):
...
                 print_experiment_set.build_experiment_chains()
 
+                # Reindex the experiments in the print set to match the overall set
+                for exp_name, _, _ in print_experiment_set.all_experiments():
+                    app_inst = experiment_set.get_experiment(exp_name)
+                    experiment_index = app_inst.expander.expand_var_name(
+                        app_inst.keywords.experiment_index
+                    )
+
+                    print_app_inst = print_experiment_set.get_experiment(exp_name)
```
douglasjacobsen, line 672: The second argument in the loop is the same as the print_app_inst. So, I would suggest we change the loop to:

`for exp_name, print_app_inst, _ ...`

and then removing this line.

douglasjacobsen (changes requested):
LG One small request, but otherwise this is great!

lib/ramble/ramble/cmd/workspace.py
dapomeroy: Created https://github.com/GoogleCloudPlatform/ramble/pull/644 to fix this

PR #645, opened 2024-09-13
Update repos used in deployments

douglasjacobsen:
This merge updates how repositories are used within deployments.

Previously, when creating a deployment, Ramble would create configuration files to add the deployment into the resulting workspace.

This merge changes this behavior so that the repository is created, but it is added at deployment pull time instead of deployment push time. This allows the path of the deployment respository to be absolute on the remote machine rather than a relative path that hopefully will work on the remote machine.

### DIFF
--- a/lib/ramble/ramble/cmd/deployment.py
+++ b/lib/ramble/ramble/cmd/deployment.py
@@ -24,6 +24,7 @@ import ramble.workspace
 import ramble.workspace.shell
 import ramble.pipeline
 import ramble.filters
+from ramble.main import RambleCommand
 
 
 description = "(experimental) manage workspace deployments"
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/cmd/deployment.py
```diff
@@ -135,6 +136,13 @@ def pull_file(src, dest):
 
             pull_file(src, dest)
 
+        obj_repo_path = os.path.join(
+            ws.root, ramble.pipeline.PushDeploymentPipeline.object_repo_name
+        )
+        if os.path.exists(obj_repo_path):
+            repo_cmd = RambleCommand("repo")
```
linsword13, line 143: In `RambleCommand`'s doc-string there's a mention that that is for testing, I don't see test-only stuff in its impl, but want to confirm. I do think it's cleaner to invoke the command instead of doing the lower-level yaml writing.

linsword13 (commented):
This is a very helpful change, I recalled wanting it when doing the gke PoC!

One minor question in the review.

lib/ramble/ramble/cmd/deployment.py
douglasjacobsen: Yeah, I think the intent / motivation for `RambleCommand` was for testing, but it seemed nicer to use here than trying to re-do what it does already.

linsword13 (approved):

PR #649, opened 2024-09-17
Move tuned-adm to builtin repo

linsword13:
Also update the active profile query to check all nodes.

### DIFF
--- a/var/ramble/repos/builtin/modifiers/tuned-adm/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/tuned-adm/modifier.py
@@ -1,0 +1,76 @@
+# Copyright 2022-2024 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import os
+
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/modifiers/tuned-adm/modifier.py
```diff
@@ -0,0 +1,76 @@
...
+                if "active profile:" in line:
+                    profiles.add(line.split(":")[-1].strip())
+            profiles_str = ",".join(profiles)
+
+        if profiles:
+            with open(profile_path, "a") as f:
+                profiles_str = ",".join(profiles)
+                f.write(f"Applied profiles: {profiles_str}")
+
+    figure_of_merit(
```
douglasjacobsen, line 75: Odd request, so feel free to push back if it's a bad idea.

Do you think we could have the count of different profiles tracked as a FOM as well?

linsword13: No I think this is a good idea, in fact I was thinking maybe it'd be even more direct to have a `success_criteria` that checks there's a single profile and that matches with the given `tuning-profile` value?

douglasjacobsen: Yeah, that sounds like a good idea too.

linsword13: Done.

douglasjacobsen (approved):

PR #650, opened 2024-09-18
Allow using modifier_variables inside prepare_analysis

linsword13:


var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
```diff
@@ -156,7 +156,7 @@ def _process_physical_hosts(self):
                 f.write(f"Level 2 groups = {len(level2_groups)}\n")
                 f.write(f'All hosts = {",".join(all_hosts)}\n')
 
-    def _prepare_analysis(self, workspace):
+    def _prepare_analysis(self, workspace, app_inst=None):
```
douglasjacobsen: Instead of passing the `app_inst` would it be better / more backwards compatible to have the modifier instance have a reference to it's application instance? Maybe set in `inherit_from_application` ?

I'm worried about changing the interface to phases in a way that could break external apps / modifiers. But we could also just look at the modifiers in benchpark to see if they use this method much.

linsword13: Discussed offline. Resolving this comment.

douglasjacobsen (approved):

PR #652, opened 2024-09-18
Remove `_write_inventory` as a phase

rfbgo:
Currently we have a bug where we write the inventory before it is calculated, as that is often done outside of the phase system (eg in complete for setup).

The two fixes are:

1. Promote calls to caculating inventory into the phase system (not done here); or
2. Couple the inventory writing to the calculation, outside of the phase system (example here)

#2 might not actually be the right fix, but putting this up to start the discussion :)

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -1522,8 +1522,6 @@ class ApplicationBase(metaclass=ApplicationMeta):
 
         self.experiment_hash = ramble.util.hashing.hash_json(self.hash_inventory)
 
-    register_phase("write_inventory", pipeline="setup", run_after=["make_experiments"])
-
     def _write_inventory(self, workspace, app_inst=None):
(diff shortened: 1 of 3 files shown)


douglasjacobsen:
I kind of feel like having hash construction and inventory writing as a phase would make the semantics a bit more clear than they are right now.

Also, I don't think this PR is technically correct, because you're updating `construct_hash` to always write an inventory. And on pipelines like `analyze` we only want to read the inventory (And maybe diff it against what the experiment says the hash should be) but we don't ever want to write an inventory in that pipeline.

Looking at the way the code was before this PR, it seems like the problem is that in the setup pipeline, construct hash happens in the `complete` portion, while write inventory is a phase (thus happening in `execute`).

It seems like if construct_hash was a phase, that could help, but I also think the semantics don't really make that nice (based on how it happens at different points, and we want some error checking and stuff based on the pipeline it's used in).

PR #655, opened 2024-09-19
Do not abort when metadata log is missing

linsword13:


var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
```diff
@@ -123,16 +125,20 @@ def _process_id_list(self):
                 f.write(", ".join(sorted(ids)))
 
     def _process_physical_hosts(self):
+        run_dir = self.expander.expand_var("{experiment_run_dir}")
+        log_path = os.path.join(
```
douglasjacobsen: Should we use the `get_file_path` util here?

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
```diff
@@ -145,9 +151,7 @@ def _process_physical_hosts(self):
                     level2_groups.add(levels[2])
 
         with open(
-            self.expander.expand_var(
-                "{experiment_run_dir}/gcp-metadata.topology_summary.log"
-            ),
+            os.path.join(run_dir, "gcp-metadata.topology_summary.log"),
```
douglasjacobsen, line 157: Same comment here about the `get_file_path` util.

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
linsword13: Good call, updated.

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
linsword13: Here I think it's fine to not use it, as it opens this file with `w+` mode (and only write to it.)

douglasjacobsen (approved):

PR #657, opened 2024-09-23
Update intel-hpl app definition

linsword13:
This update includes:
    
* Ensure intel-hpl is invoked with correct numa control. Previously it
   directly invokes the `xhpl_intel64_dynamic` binary without setting the
   `HPL_HOST_NODE` env var. Now it's changed to call `runme_intel64_prv`
   runner script first which sets up the env var.
    
* Accommodate new mkl 2024 directory structure for locating the
   benchmark files.

### DIFF
--- a/var/ramble/repos/builtin/applications/intel-hpl/application.py
+++ b/var/ramble/repos/builtin/applications/intel-hpl/application.py
@@ -42,7 +42,7 @@ class IntelHpl(ExecutableApplication):
     required_package("intel-oneapi-mkl", package_manager="spack*")
 
     # This step does a few things:
-    # - Set up the env variables expected by runme_intel64_prv
+    # - Prepare calling for the script runme_intel64_prv
     #   (We call this runner script instead of the underlying xhpl_intel64_dynamic
     #    since it sets up derived env var HPL_HOST_NODE for numa placement control.)
     # - Link in the xhpl_intel64_dynamic binary to the running dir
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/intel-hpl/application.py
```diff
@@ -26,27 +26,66 @@ class IntelHpl(ExecutableApplication):
...
+    #   (This is needed due to runme_intel64_prv invoking it using "./")
+    # - Account for newer directory layout from mkl 2024
+    executable(
+        "prepare",
+        template=[
+            r"""
+export MPI_PROC_NUM="{n_ranks}"
+export MPI_PER_NODE="{processes_per_node}"
+export NUMA_PER_MPI="{numa_per_mpi}"
+export HPL_EXE=xhpl_intel64_dynamic
```
douglasjacobsen: For these environment variables (that are explicitly exported), do you think we could use this:
https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/language/application_language.py#L245

Instead if exporting them directly within the executable?

douglasjacobsen (changes requested):

var/ramble/repos/builtin/applications/intel-hpl/application.py
linsword13: Good call, updated.

douglasjacobsen (approved):

PR #658, opened 2024-09-23
fix/update tutorial issues

akiki-liang0:
- In tut 9, remove section with ramble workspace analyze --always-print-foms command as flag is deprecated
- In tut 10, ramble mods info command is deprecated, change to updated command
- In tut 11, update spack builtin reference in example code
- In mirrors tut, add spack package manager to fix required key 'wrf_path' not defined warning
- In mirrors tut, update mirror add commands for ramble and spack

### DIFF
--- a/lib/ramble/docs/tutorials/mirrors.rst
+++ b/lib/ramble/docs/tutorials/mirrors.rst
@@ -518,9 +518,9 @@ For example, using the  mirror directories we created above,
 
 .. code-block:: console
 
-    $ ramble mirror add --scope=[site,user] $HOME/wrfv4_mirror/inputs
+    $ ramble mirror add --scope={site,user} ramble_mirror $HOME/wrfv4_mirror/inputs
 
-    $ spack mirror add $HOME/wrfv4_mirror/software
+    $ spack mirror add spack_mirror $HOME/wrfv4_mirror/software
 
 To validate that the mirrors were installed correctly, try something like the following,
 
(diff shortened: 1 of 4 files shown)


lib/ramble/docs/tutorials/mirrors.rst
```diff
@@ -514,9 +518,9 @@ For example, using the  mirror directories we created above,
 
 .. code-block:: console
 
-    $ ramble mirror add --scope=[site,user] $HOME/wrfv4_mirror/inputs
+    $ ramble mirror add --scope={site,user} ramble_mirror $HOME/wrfv4_mirror/inputs
```
linsword13: I was under the impression that `--scope` allows for a single value only, but maybe I was mistaken. Does this actually work (by adding it to both site and user configs?)

linsword13 (commented):
Thanks for raising this PR! One question but else LGTM.

lib/ramble/docs/tutorials/mirrors.rst
douglasjacobsen: Yeah, this actually works. If i test it, it only places it into my user config though, So it doesn't appear to add it to both scopes.

I think the confusion is that in the original version, the `[site,user]` was more meant as an optional "Choose one of these" not "Add to both scopes at the same time"

douglasjacobsen (approved):

lib/ramble/docs/tutorials/mirrors.rst
akiki-liang0: After checking it only adds to users and not site. I see that a fix has already been made to the tut, so looks like that's resolved.

PR #659, opened 2024-09-24
Add warning when no valid phases are selected

rfbgo:


lib/ramble/ramble/pipeline.py
```diff
@@ -146,6 +146,11 @@ def _execute(self):
 
             phase_list = app_inst.get_pipeline_phases(self.name, self.filters.phases)
 
+            if not phase_list:
+                logger.warn("Not valid phases selected, please verify requested phases")
```
douglasjacobsen, line 150: I don't think we should warn at this level.

I think instead we should count how many experiments were executed without phases, and warn after all experiments are completed, only if no phase was executed across *all* experiments.

douglasjacobsen (changes requested):

lib/ramble/ramble/pipeline.py
rfbgo: Sure, that makes sense

PR #663, opened 2024-09-26
Add filtering so warning is only given for non-default-selection

rfbgo:
Previously this would also work for pipelines that don't have phases (eg ramble on). This adds a simple check to only warn if the user requested a change to the phases away from the default path

### DIFF
--- a/lib/ramble/ramble/filters.py
+++ b/lib/ramble/ramble/filters.py
@@ -8,13 +8,15 @@
 
 import itertools
 
+ALL_PHASES = ["*"]
+
 
 class Filters:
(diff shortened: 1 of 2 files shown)


douglasjacobsen (dismissed):
I think this is good. One alternative (I don't really have a strong preference) would be to make whether the pipeline uses phases an attribute of each individual pipeline, and then guard the warning by that.

That would also help catch an (arguably very weird) edge case where an application purposefully removes all of its phases but the phase filter is still `["*"]`.

douglasjacobsen (approved):

PR #670, opened 2024-10-01
Ramble reports

dapomeroy:


lib/ramble/ramble/cmd/results.py
```diff
@@ -36,25 +133,57 @@ def results_upload(args):
...
     """
-    Import Ramble experiment results from a JSON file.
+    Import Ramble experiment results from a JSON or YAML file.
+
+    Returns a results dictionary.
     """
     logger.debug("File to import:")
     logger.debug(filename)
 
+    # TODO: this isn't very safe checking
```
rfbgo: Can we wrap this in a `with` or (`try`) block please?

lib/ramble/ramble/language/shared_language.py
```diff
@@ -77,8 +79,57 @@ def _execute_figure_of_merit_context(obj):
...
+    INDETERMINATE = 3  # requires interpretation or FOM type not defined
+    INAPPLICABLE = 4  # non-numerical or no direction is 'better', like strings or categories
+
+    def get_direction_str(self):
+        if self == BetterDirection.HIGHER:
+            return " (Higher is Better)"
+        if self == BetterDirection.LOWER:
+            return " (Lower is Better)"
+        else:
+            return ""
```
rfbgo: Should this be `INDETERMINATE`?

lib/ramble/ramble/language/shared_language.py
```diff
@@ -77,8 +79,57 @@ def _execute_figure_of_merit_context(obj):
...
+            FomType.THROUGHPUT: BetterDirection.HIGHER,
+            FomType.MEASURE: BetterDirection.INDETERMINATE,
+            FomType.CATEGORY: BetterDirection.INAPPLICABLE,
+            FomType.INFO: BetterDirection.INAPPLICABLE,
+            FomType.UNDEFINED: BetterDirection.INDETERMINATE,
+        }
+
+        return direction[self]
+
+    def copy(self):
```
rfbgo: Whats the motivation behind doing this instead of expecting a user to do something like a deepcopy?

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,959 @@
...
+# except according to those terms.
+
+import copy
+import datetime
+import os
+import re
+
+import pandas as pd
+
+# import numpy as np
```
rfbgo: remove?

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,959 @@
...
+            logger.die(
+                "No JSON or YAML results file was found. Please run "
+                "'ramble workspace analyze -f json'."
+            )
+    return results_dict
+
+
+def is_repeat_child(experiment):
+    try:
+        if int(experiment["RAMBLE_VARIABLES"]["repeat_index"]) > 0:
```
rfbgo: Can we update this to use the recently added/propogated `repeat_index`? 

(see https://github.com/GoogleCloudPlatform/ramble/pull/633)

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,959 @@
...
+                    else:
+                        skip_exps.append(exp_name + f".{n}")
+
+            for context in exp["CONTEXTS"]:
+                for fom in context["foms"]:
+                    # Expand to one row/FOM/context w/ a copy of the experiment vars and metadata
+                    exp_copy = copy.deepcopy(exp)
+
+                    # Remove context dict and add the current FOM values
+                    # TODO: Can we clean this partial copy with a class?
```
rfbgo: Any reason this isn't something more like a copy?

lib/ramble/ramble/test/data/reports/results.test.json
```diff
@@ -0,0 +1 @@
+/usr/local/google/home/dpomeroy/VSCode/ramble/var/ramble/workspaces/basic_gromacs/results.2024-06-28_20.55.29.json
```
rfbgo: Not a big deal, but do you want such a "real" (and personal) path here?

pytest.ini
```diff
@@ -3,7 +3,7 @@
 addopts = --durations=30 -ra
 testpaths =
   lib/ramble/ramble/test
-  var/ramble/repos/*/*/*/test
+  #var/ramble/repos/*/*/*/test/*.py
```
rfbgo: Just double checking, is this intended?

rfbgo (changes requested):
Some quick feedback and suggestions, since I know this is not yet in it's final form 

Thanks for working on this awesome feature!

lib/ramble/ramble/cmd/results.py
dapomeroy: Done

lib/ramble/ramble/language/shared_language.py
dapomeroy: This method returns a string that we can append on a chart title to convey the 'better direction' in the chart text. For example, with Gromacs you might get "Core Time by workload name (Lower is Better)" as the chart title. But if better direction is indeterminate I don't think we should append anything on the title, hence the empty string.

lib/ramble/ramble/language/shared_language.py
dapomeroy: I was trying to avoid making a special case out of FomType in this loop within ApplicationBase._analysis_dicts() (L2102) so that it can copy the FomType object:
```
            for fom, fom_def in mod.figures_of_merit.items():
                fom_definitions[fom] = {"origin": f"{mod}", "origin_type": "modifier"}
                for attr in fom_def.keys():
                    if isinstance(fom_def[attr], (list, FomType)):
                        fom_definitions[fom][attr] = fom_def[attr].copy()
                    else:
                        fom_definitions[fom][attr] = self.expander.expand_var(
                            fom_def[attr], mod_vars
                        )
```

I saw several examples of this kind of copy method in the code already, like WorkloadVariable. Do you know of a better way to do it?

lib/ramble/ramble/reports.py
dapomeroy: Let's discuss. I can see some drawbacks here but not quite clear on how to improve it.

lib/ramble/ramble/test/data/reports/results.test.json
dapomeroy: Oh, this is weird. I have a full results file that's been de-identified in my branch, but it's showing up here as just one line. I must have mixed something up in Git. I'll fix it - please disregard for the moment. Thanks!

pytest.ini
dapomeroy: I don't know. You changed it in this commit: https://github.com/GoogleCloudPlatform/ramble/commit/7f8ab379ff063abf783b04f2d86531e333a53738 
Do you remember what it was about? Can I just assume it was to test something out and revert it?

lib/ramble/ramble/test/data/reports/results.test.json
dapomeroy: Figured it out. Somehow I ended up with a symlink instead of the file inside my workspace. Now fixed

lib/ramble/ramble/reports.py
dapomeroy: Done

lib/ramble/ramble/language/shared_language.py
dapomeroy: Per discussion, moved this out of the class and into reports.py

lib/ramble/ramble/language/shared_language.py
dapomeroy: Updated to deepcopy to match the other class methods

lib/ramble/ramble/reports.py
dapomeroy: Updated with a mapping dict to remove some of the magic strings.

pytest.ini
dapomeroy: Reverted

lib/ramble/ramble/reports.py
dapomeroy: Added enums to remove most of the rest of magic strings

lib/ramble/ramble/language/shared_language.py
```diff
@@ -77,8 +81,66 @@ def _execute_figure_of_merit_context(obj):
     return _execute_figure_of_merit_context
 
 
+# For a FOM, the direction that is 'better' e.g., faster is better
+class BetterDirection(Enum):
```
douglasjacobsen: This is super nit-picky, but I would kind of prefer if these (BetterDirection and FomType) lived in a separate module that is more clear it relates to FOMs.

Maybe in a util module or something?

IMO, I would almost never think to look in the shared_language module for enums about how FOMs behave.

lib/ramble/ramble/application.py
```diff
@@ -1824,11 +1825,17 @@ def calculate_statistics(self, workspace):
         namespace.
         """
 
-        def is_numeric(value):
-            """Returns true if a fom value is numeric"""
+        def is_numeric_fom(fom):
```
douglasjacobsen, line 1830: Could it make sense to move this into the FomType class? I'm not sure if it makes the most sense yet...

Maybe at a minimum put a TODO to help us remember to move it once we create a Fom class.

lib/ramble/ramble/cmd/results.py
```diff
@@ -6,25 +6,122 @@
...
+    report_parser.add_argument(
+        "--multi-line",
+        dest="multi_line",
+        nargs="+",
+        action="append",
+        help="generate a scaling report, requires two args: [performance metric] [scaling metric]"
+        "[optional: group by]",
+        required=False,
+    )
+    report_parser.add_argument(
```
douglasjacobsen, line 69: This might be answered as I get through more of the code, but if this is used for filtering how some experiments...
You can use the where common argument (in the common module) for this.

Also, you might want to add  the `--execlude-where` argument to keep the interfaces symmetric.

lib/ramble/ramble/cmd/results.py
```diff
@@ -36,25 +133,56 @@ def results_upload(args):
...
+    Import Ramble experiment results from a JSON or YAML file.
+
+    Returns a results dictionary.
     """
     logger.debug("File to import:")
     logger.debug(filename)
 
-    imported_file = open(filename)
+    with open(filename) as imported_file:
+        logger.msg(f"Importing {filename}")
```
douglasjacobsen: Maybe update to `Importing results file: {filename}`

requirements.txt
```diff
@@ -8,3 +8,5 @@ pyarrow==3.0.0;python_version<="3.6"
 google-cloud-bigquery
 tqdm
 deprecation
+pandas
```
douglasjacobsen, line 11: We discussed this internally, but each of these should have a guarded import in the code, where if the import fails we print something about installing the requirements.txt file.

dapomeroy: This is done

lib/ramble/ramble/application.py
dapomeroy: At first glance I don't find it intuitive, because the FomType class is an enum that doesn't have any information about the FOM inside of it. I think a FOM class seems like a better fit. Added todo.

lib/ramble/ramble/cmd/results.py
dapomeroy: Bob added this and I think it works differently from the common 'where'. It passes in a 'where' string to pandas to downselect the dataframe. It's more like a SQL 'where' statement. I haven't looked into this closely to see if the common 'where' might be useful, but making the pandas query format available adds a lot of control over what shows up in the chart.

lib/ramble/ramble/cmd/results.py
dapomeroy: Done

lib/ramble/ramble/language/shared_language.py
dapomeroy: OK, moved out to a new foms.py util module

lib/ramble/ramble/cmd/results.py
rfbgo: Will rename to `pandas-where`

PR #672, opened 2024-10-02
Disable explicit testing of ramble directives

douglasjacobsen:
This merge turns off the testing of ramble directives in the application_language tests. These are prone to leakage into the base classes, which breaks subsequent tests.


douglasjacobsen:
We also have good coverage of the directives using the end-to-end tests. But ultimately should update this test to do something else.

rfbgo (approved):

linsword13 (approved):

PR #673, opened 2024-10-02
Add a command runner utility class

douglasjacobsen:
This merge adds a utility class for creating command runners, which are external commands to Ramble that have behaviors that change whether the workspace is executing under dry-run or not.

The command runner is also applied to the spack package managers.

### DIFF
--- a/lib/ramble/ramble/util/command_runner.py
+++ b/lib/ramble/ramble/util/command_runner.py
@@ -1,0 +1,173 @@
+# Copyright 2022-2024 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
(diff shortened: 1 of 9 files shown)


lib/ramble/ramble/util/command_runner.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def _dry_run_print(self, executable, args, return_output=False):
+        """Print the command that would be executed if dry-run was false.
+
+        Args match the execute method.
+        """
+        logger.msg(f"DRY-RUN: would run {executable}")
+        logger.msg(f"         with args: {args}")
+        if return_output:
+            return None
```
linsword13: Given the implicit None return, this feels a bit redundant?

linsword13 (dismissed):

lib/ramble/ramble/util/command_runner.py
douglasjacobsen: Good point. Fixed.

linsword13 (approved):

PR #674, opened 2024-10-02
Update used variable tracking

douglasjacobsen:
This merge changes the used variable tracking logic to only track through expanding other variable definitions. This prevents modifier and package manager variables from being marked as used even if they aren't.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -545,7 +545,7 @@ class ApplicationBase(metaclass=ApplicationMeta):
         if self.package_manager is not None:
             # Add package manager variable defaults as placeholder
             for var in self.package_manager.package_manager_variables.values():
-                self.expander._variables[var.name] = var.default
+                self.variables[var.name] = var.default
 
         ##########################################
         # Expand used variables to track all usage
@@ -582,6 +582,9 @@ class ApplicationBase(metaclass=ApplicationMeta):
         ############################
         # Reset variable definitions
         ############################
+        for var in self.variables:
+            if var not in backup_variables:
+                del self.variables[var]
 
         for var, val in backup_variables.items():
             self.variables[var] = val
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/application.py
```diff
@@ -530,22 +530,37 @@ def build_used_variables(self, workspace):
...
+        # Add modifier mode variables defaults as a placeholder:
+        for mod_inst in self._modifier_instances:
+            for var in mod_inst.mode_variables().values():
+                if var.name not in self.variables:
+                    self.variables[var.name] = var.default
+
+        if self.package_manager is not None:
+            # Add package manager variable defaults as placeholder
+            for var in self.package_manager.package_manager_variables.values():
+                self.expander._variables[var.name] = var.default
```
linsword13: Curious why pkgman vars are added to expander, while modifiers are added to `self.variables`?

lib/ramble/ramble/application.py
```diff
@@ -557,12 +572,19 @@ def build_used_variables(self, workspace):
...
             self.expander._used_variables.add(template_name)
             self.expander.expand_var(template_conf["contents"])
 
-        if self.package_manager is not None:
-            self.package_manager.build_used_variables(workspace)
+        ############################
+        # Reset variable definitions
+        ############################
+
+        for var, val in backup_variables.items():
```
linsword13, line 589: Does that mean we are OK with keeping new variables added?

lib/ramble/ramble/application.py
douglasjacobsen: Silly typo. They are both references to the same thing. So, which one they are added to is not super important. I'll fix this though.

lib/ramble/ramble/application.py
douglasjacobsen: Good point. I meant to delete the difference, so let me fix that up

linsword13 (approved):

PR #681, opened 2024-10-03
Enable variable modifications to layer changes on the same variable

douglasjacobsen:
This merge allows modifiers that modify the same variable to layer their changes onto the modified variable.

Previously, only the last modifier to modify a variable would have it's modification stick.

### DIFF
--- a/lib/ramble/ramble/modifier.py
+++ b/lib/ramble/ramble/modifier.py
@@ -252,11 +252,12 @@ class ModifierBase(metaclass=ModifierMeta):
 
         for var, var_mod in self.variable_modifications[self._usage_mode].items():
             if var_mod["method"] in ["append", "prepend"]:
-                # var_str = app.expander.expansion_str(var)
-                # prev_val = app.expander.expand_var(var_str)
-                prev_val = app.variables[var]
                 if var in extra_vars:
                     prev_val = extra_vars[var]
+                elif var in app.variables:
+                    prev_val = app.variables[var]
+                else:
+                    prev_val = ""
 
                 if var_mod["method"] == "append":
                     mods[var] = f'{prev_val}{var_mod["modification"]}'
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/modifier.py
```diff
@@ -255,10 +255,13 @@ def modded_variables(self, app):
                 # var_str = app.expander.expansion_str(var)
                 # prev_val = app.expander.expand_var(var_str)
                 prev_val = app.variables[var]
+                if var in extra_vars:
```
linsword13, line 255: nit: maybe using the single-assignment-form is clearer?

```
if var in extra_vars:
    prev_val = extra_vars[var]
else:
    prev_val = app.variables[var]
```

linsword13 (commented):
LGTM. Do you think it'd be good to add a test for it?

lib/ramble/ramble/modifier.py
douglasjacobsen: Yeah, good point. Fixed.

douglasjacobsen:
I added a test, and updated existing modifiers to include the space separator as part of their variable modifications (given I removed the auto-injected space)

linsword13:
> I added a test, and updated existing modifiers to include the space separator as part of their variable modifications (given I removed the auto-injected space)

Do you feel having the separator auto-injected would be less error-prone than relying on modifier author to remember adding the space?

Perhaps a more general approach is to have the directive to accept a `template_func` (`(prev_value, new_mod) => new_value`)? (More just for a discussion, no need to add to the current PR though.)

douglasjacobsen:
Yes, I actually have an internal task tracking this which I think is a high priority.

Injecting a space manually is definitely a hack for the time being, and we should have a nicer interface for this.

linsword13 (approved):

PR #688, opened 2024-10-09
Add local mode to gcp-metadata modifier

rfbgo:
This allows for it to contiune to work in environments where we might not have pdsh, such as GKE

### DIFF
--- a/var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
@@ -27,11 +27,41 @@ class GcpMetadata(BasicModifier):
     maintainers("rfbgo")
 
     mode("standard", description="Standard execution mode")
+    mode(
+        "local", description="Local execution (disables parallel prefix/pdssh)"
+    )
     default_mode("standard")
 
     software_spec("pdsh", pkg_spec="pdsh", package_manager="spack*")
 
     required_variable("hostlist")
+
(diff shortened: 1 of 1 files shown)


linsword13 (approved):

var/ramble/repos/builtin/modifiers/gcp-metadata/modifier.py
```diff
@@ -27,12 +27,42 @@ class GcpMetadata(BasicModifier):
...
     required_variable("hostlist")
 
+    modifier_variable(
+        "metadata_parallel_prefix",
+        default="pdsh -R ssh -N -w {hostlist} '",
+        modes=["standard"],
+        description="Express how parlalelism should be done between nodes",
+    )
+    modifier_variable(
+        "metadata_parallel_prefix",
```
douglasjacobsen, line 46: Pretty nit-picky, but should this be called `metadata_prefix` instead of `metadata_parallel_prefix`, since we attempt to ignore it in serial anyway, but it could technically be used to pre/suffix commands when running locally too.

rfbgo: Yea I had a hard time naming it. I did go with `metadata_prefix` to start but I think that is actually very misleading, since it only affect those commands marked with `per_node=True` (hence the name being my best attempt to convey that).

I think if we went with  `metadata_prefix` people would be pretty surprised with how it acts when setting it from the YAML.

PR #689, opened 2024-10-10
Remove new lines when defining path variables for spack

douglasjacobsen:
This merge fixes an issue where spack names are returned with newlines (which are then left in path variable name definitions) causing path variables to not function properly when actually using spack.

These problems do not show up in dry-runs.

### DIFF
--- a/var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
@@ -1128,11 +1128,17 @@ class SpackRunner(ramble.util.command_runner.CommandRunner):
             name_match = name_regex.match(name)
             if name_match:
                 name = name_match.group("name")
+        else:
+            # Remove newlines and whitespace from name
+            name = name.replace("\n", "").strip()
 
         if location is None:
             location = os.path.join(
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
```diff
@@ -1128,11 +1128,17 @@ def get_package_path(self, package_spec):
             name_match = name_regex.match(name)
             if name_match:
                 name = name_match.group("name")
+        else:
+            # Remove newlines and whitespace from name
+            name = name.replace("\n", "").strip()
```
linsword13, line 1133: Just curious why spack would return something with newlines?

douglasjacobsen: Maybe just so it's printed to the screen nicely?

i.e.

```
$ spack -D software/wrfv4-intel-mpi/ find --format={name} wrf
wrf
$ 
```

linsword13: Ah right

linsword13 (approved):

PR #691, opened 2024-10-11
Front-end commands for managing software

douglasjacobsen:
This merge adds `ramble workspace manage` commands, to be a user entrypoint for manipulating various definitions within a workspace.

The `generate-config` command is aliased to `manage experiments`.

Additionally, this merge adds `manage software` for manipulating package and environment definitions.

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -51,6 +51,7 @@ from ramble.namespace import namespace
 import ramble.util.matrices
 import ramble.util.env
 from ramble.util.logger import logger
+from ramble.util.conversions import list_str_to_list
 
 #: Environment variable used to indicate the active workspace
 ramble_workspace_var = "RAMBLE_WORKSPACE"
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -935,6 +1065,7 @@ def yaml_add_comment_before_key(
 
             return base
 
+        from ramble.util.conversions import list_str_to_list
```
rfbgo: Should this import be higher?

rfbgo (dismissed):
LGTM, one tiny nit :)

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: In this method? Or more like a module level import?

I only imported it here, because it's a kind of weird utility that nothing else needs (kind of like the yaml import below it).

But I could move it up to module level too.

linsword13 (dismissed):

rfbgo (approved):

PR #692, opened 2024-10-11
Set RAMBLE_PYTHON when setup-env is sourced

douglasjacobsen:
This merge ensures RAMBLE_PYTHON is defined when setup-env is sourced, to ensure longer paths don't break ramble's ability to function properly.


rfbgo (dismissed):

linsword13 (approved):

linsword13:
Should the changes trigger cloud-build runs?

douglasjacobsen:
It might be good to fix this (because technically I'm changing the triggers) but I think given the trigger state now they wouldn't.

douglasjacobsen:
OK, I updated them to run now. Let's see if it works. :)

PR #693, opened 2024-10-12
Add more flexible report generation for intel-aps modifier

linsword13:
* Always generate a html version (as well as a text version) of the summary
* Allow specifying `aps_extra_reports` to dictate additional report generation. For instance `aps_extra_reports: "all"` would generate a set of predefined reports
* Add checking to warn of `APS_STAT_LEVEL` is not set to the required level

### DIFF
--- a/var/ramble/repos/builtin/modifiers/intel-aps/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/intel-aps/modifier.py
@@ -67,12 +67,12 @@ class IntelAps(BasicModifier):
     modifier_variable(
         "aps_extra_reports",
         default="",
-        description="""
+        description=f"""
         Comma-separated descriptors specifying extra reports (besides the summary) to generate.
         Syntax definition:
-            aps_extra_reports = spec { "," spec }
+            aps_extra_reports = spec {{ "," spec }}
             spec = pre_defined_spec | custom_spec | "all"
-            pre_defined_spec = <keys in _PREDEFINED_REPORTS>
+            pre_defined_spec = {",".join(_PREDEFINED_REPORTS["mpi"].keys())}
             custom_spec = "custom" ":" options
             options = <letters>
         Examples:
--- a/var/ramble/repos/builtin/modifiers/intel-aps/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/intel-aps/modifier.py
@@ -9,6 +9,22 @@
 from ramble.modkit import *  # noqa: F403
 
 
+# Pre-defined charts and graphs
+# The value is a tuple of (options_for_aps, min_stat_level)
+_PREDEFINED_REPORTS = {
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/modifiers/intel-aps/modifier.py
```diff
@@ -39,6 +55,33 @@ class IntelAps(BasicModifier):
...
     )
 
+    modifier_variable(
+        "aps_stat_level",
+        default="1",
+        description="Used to define the APS_STAT_LEVEL env variable",
+        mode="mpi",
+    )
+
+    modifier_variable(
```
douglasjacobsen, line 67: Can we generate a `values` argument using the keys from `_PREDEFINED_REPORTS`?

var/ramble/repos/builtin/modifiers/intel-aps/modifier.py
```diff
@@ -39,6 +55,33 @@ class IntelAps(BasicModifier):
...
+            pre_defined_spec = <keys in _PREDEFINED_REPORTS>
+            custom_spec = "custom" ":" options
+            options = <letters>
+        Examples:
+        * "transfer-graph,message-sizes,custom:-t"
+          Generates transfer comm. graph, message-size summary and MPI time per rank chart
+        * "all"
+          Generates all graphs defined in _PREDEFINED_REPORTS
+        """,
+        mode="mpi",
```
douglasjacobsen, line 84: Would this only work for `mode="mpi"`? Or should the `_PREDEFINED_REPORTS` dict have a mode sub-dict, to separate report types when we add new modes?

This might be more of a request to add a TODO, so we can update this later.

douglasjacobsen (changes requested):

var/ramble/repos/builtin/modifiers/intel-aps/modifier.py
linsword13: Great call. Yes these reports only make sense for mpi mode. Changed to use a per-mode dict.

var/ramble/repos/builtin/modifiers/intel-aps/modifier.py
linsword13: This `aps_extra_reports` var allows for a comma-separated string that defines multiple report types, it also allows passing in "custom" report specs. Given that it doesn't have a static allow-list.

douglasjacobsen: Ah OK cool. Well can we maybe have it extract the "known" report types from the dict when generating the description string?

linsword13: Good point, updated.

douglasjacobsen (approved):

PR #695, opened 2024-10-14
Allow `workspace concretize` to be additive

douglasjacobsen:
This merge updates the `workspace concretize` command to augment a workspace when new experiments (or modifiers) are added to it.

Additionally, a `--quiet` mode is added to allow it to silently ignore errors when packages would normally be overwritten (and not overwrite the existing definitions)

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -1180,8 +1180,7 @@ class Workspace:
                             logger.debug(f"  Spec 1: {str(info)}")
                             logger.debug(f"  Spec 2: {str(packages_dict[comp])}")
                             raise RambleConflictingDefinitionError(
-                                f"Compiler {comp} would be defined defined "
-                                "in multiple conflicting ways"
+                                f"Compiler {comp} would be defined " "in multiple conflicting ways"
                             )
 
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1183,20 +1176,19 @@ def concretize(self):
                                 ramble.config.add(
                                     config_path, scope=self.ws_file_config_scope_name()
                                 )
-                        elif not specs_equiv(info, packages_dict[comp]):
+                        elif not quiet and not specs_equiv(info, packages_dict[comp]):
                             logger.debug(f"  Spec 1: {str(info)}")
                             logger.debug(f"  Spec 2: {str(packages_dict[comp])}")
                             raise RambleConflictingDefinitionError(
-                                f"Compiler {comp} defined in multiple conflicting ways"
+                                f"Compiler {comp} would be defined defined "
```
linsword13: Too defined ;)

linsword13 (dismissed):

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: 🤣  Fixed!

linsword13 (approved):

PR #703, opened 2024-10-17
Allow package managers to have their own sub-directories under a workspace's software dir

douglasjacobsen:
This merge allows a workspace to have multiple package manager directories under the `software` directory, which can be used to house independent software environments.

This allows the same named environment to render differently for different experiments (if needed) without causing collisions.

Tests and documentation are updated for this as well.

### DIFF
--- a/lib/ramble/docs/workspace_config.rst
+++ b/lib/ramble/docs/workspace_config.rst
@@ -964,7 +964,9 @@ Ramble automatically generates definitions for the following variables:
 * ``experiment_index`` - Index, in set, of experiment. If part of a chain,
   shares a value with its root.
 * ``env_path`` - Absolute path to
-  ``$workspace_root/software/{env_name}.{workload_name}``
+  ``$workspace_root/software/{package_manager_name}/{env_name}.{workload_name}``
+  if no package manager is used, ``{package_manager_name}`` is replaced with
+  ``no-package-manager``.
 * ``log_dir`` - Absolute path to ``$workspace_root/logs``
 * ``log_file`` - Absolute path to
   ``{experiment_run_dir}/{experiment_name}.out``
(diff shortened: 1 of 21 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1608,6 +1608,11 @@ def shared_license_dir(self):
         """Path to the shared license directory"""
         return os.path.join(self.shared_dir, workspace_shared_license_path)
 
+    def package_manager_dir(self, package_manager):
+        if package_manager is not None:
+            return os.path.join(self.software_dir, package_manager.name)
+        return os.path.join(self.software_dir, "no-package-manager")
```
rfbgo: Do you think when `package_manager` is the "no package manager" package manager it should still have a valid object to give a more consistent interface?

douglasjacobsen: I don't dislike the idea, but I also can't decide what is a reasonable name for that package manager.

TBF, I think this case should *never* happen (because if the package manager is none, then it will never actually try to create a software environment, and it won't need to have it's own directory).

So this is more a very unnecessary fallback mode to avoid some gross error from popping up if you call this method without a package manager.

douglasjacobsen: I updated this so that the `package_manager_dir` is an attribute of the package manager, rather than the workspace. This should get around the error case where `package_manager` is None now to help simplify the interface.

linsword13 (commented):
LGTM! I haven't thought through it, but is there backward-compatibility concerns? (Probably not, and it's just that running `workspace setup` would not reuse stuff from the old path.)

douglasjacobsen:
@linsword13 There are backwards compatibilty concerns if you're using a static workspace from before this PR and you just try to blindly update to the latest version (After this is merged) and run `ramble on`.

Running `ramble workspace setup` should resolve all of the issues though (and won't require new software to be installed).

linsword13 (approved):

PR #712, opened 2024-10-21
Expand deployment arguments and clean URLs

douglasjacobsen:
This merge performs two primary functions. The first is to clean deployment URLs using the `get_maybe_local_path` helper utility. This fixes an issue when uploading to any cloud storage provider with a URL that has trailing back-slashes.

The second allows arguments on `deployment push` to be expanded using the workspace top-level variables.

Documentation is added around the second enhancement.

### DIFF
--- a/lib/ramble/ramble/util/path.py
+++ b/lib/ramble/ramble/util/path.py
@@ -102,9 +102,21 @@ def canonicalize_path(path):
     return path
 
 
-def get_maybe_local_path(path):
-    """Convert a scheme-less path to absolute local path"""
-    parsed = urllib.parse.urlparse(path)
+def normalize_path_or_url(path):
+    """Convert a scheme-less path to absolute local path
+    Also, remove trailing back-slashes from the input path
+
+    Args:
+        path (str): Input path
+
+    Returns:
+        (str): Absolute local path or cleaned remote url
+    """
+
+    # Remove trailing back-slashes from path
+    real_path = path.rstrip("/")
+
+    parsed = urllib.parse.urlparse(real_path)
     if not parsed.scheme:
-        return os.path.abspath(path)
-    return path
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/util/path.py
```diff
@@ -103,8 +103,23 @@ def canonicalize_path(path):
 
 
 def get_maybe_local_path(path):
-    """Convert a scheme-less path to absolute local path"""
-    parsed = urllib.parse.urlparse(path)
+    """Convert a scheme-less path to absolute local path
+
+    Also, remove trailing back-slashes from the input path
```
linsword13, line 107: It may be a bit confusing given that the function has "local" in it, but the "removing trailing" applies to any path. Maybe rename the helper to something along the lines of "normalize_path"?

May also be worthwhile to add a test case into the existing test of this helper.

lib/ramble/ramble/util/path.py
```diff
@@ -103,8 +103,23 @@ def canonicalize_path(path):
...
+
+    Args:
+        path (str): Input path
+
+    Returns:
+        (str): Absolute local path or cleaned remote url
+    """
+    real_path = path
+
+    # Remove trailing back-slashes from path
```
linsword13, line 116: Can this be achieved with `rstrip`?

lib/ramble/ramble/util/path.py
douglasjacobsen: Updated. LMK if you think it should be changed anymore.

lib/ramble/ramble/util/path.py
douglasjacobsen: Yes, thanks. I also added tests.

linsword13 (approved):

PR #717, opened 2024-10-22
Reports v2

rfbgo:
This PR adds the `ramble results report` functionality 

It includes some basic documentation and unit tests


lib/ramble/ramble/application.py
```diff
@@ -1764,11 +1766,18 @@ def calculate_statistics(self, workspace):
         namespace.
         """
 
-        def is_numeric(value):
-            """Returns true if a fom value is numeric"""
+        # TODO: Think about if this should move to FomType class or new FOM class
```
douglasjacobsen: My gut reaction is that this *should* be part of the FomType class.

lib/ramble/ramble/application.py
```diff
@@ -2063,6 +2105,9 @@ def _analysis_dicts(self, criteria_list):
                 "units": conf["units"],
                 "origin": conf["origin"],
                 "origin_type": conf["origin_type"],
+                # FIXME: this 'to_dict' is getting something strange passed
```
douglasjacobsen: Does this still need to be fixed?

lib/ramble/ramble/cmd/results.py
```diff
@@ -6,25 +6,122 @@
...
+        metavar="WRKSPC",
+        action="store",
+        help="the workspace to report on",
+    )
+    report_parser.add_argument(
+        "--strong-scaling",
+        dest="strong_scaling",
+        nargs="+",
+        action="append",
+        help="generate a scaling report, requires two args: [performance metric] [scaling metric]"
```
douglasjacobsen: This is slightly nit-picky, but it's not super clear to me which axis scaling / performance metrics are for. Mostly because I can't tell if `scaling` is "The metric we scale performance with" of "The metric we expect to scale" ?

Performance seems somewhat clearly y-axis IMO, until I read `scaling metric`.

lib/ramble/ramble/cmd/results.py
```diff
@@ -6,25 +6,122 @@
...
+        required=False,
+    )
+    report_parser.add_argument(
+        "-n",
+        "--normalize",
+        dest="normalize",
+        action="store_true",
+        help=(
+            "Normalize charts where possible. For scaling charts, this requires fom_type to be "
+            "specified as either 'time' or 'throughput'."
```
douglasjacobsen: Maybe state where the specification needs to happen (i.e. in an application definition). This kind of sounds like something you could set on the CLI.

lib/ramble/ramble/cmd/results.py
douglasjacobsen: It might actually be better to just call them `y-axis metric` and `x-axis metric` since we refer to the axes below?

lib/ramble/ramble/test/data/reports/results.test.json
```diff
@@ -0,0 +1,1318 @@
+{
```
douglasjacobsen: Similar to our discussion on https://github.com/GoogleCloudPlatform/ramble/pull/699, is there a way to generate this with less code, so we don't have to track such a large file? (this applies to the other formats too)

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1526,38 +1526,40 @@ def dump_results(self, output_formats=["text"], print_results=False, summary_onl
...
                                     name = fom["name"]
-                                    if name not in fom_summary.keys():
-                                        fom_summary[name] = []
-                                    stat_name = fom["origin_type"]
-                                    value = fom["value"]
-                                    units = fom["units"]
-
-                                    output = f"{stat_name} = {value} {units}\n"
-                                    fom_summary[name].append(output)
+                                    if fom["origin_type"].startswith("summary::"):
```
douglasjacobsen: We should at least make a task for moving some of this logic into the FOM class, so we don't need manipulate / inspect dicts directly here.

requirements.txt
```diff
@@ -8,3 +8,5 @@ pyarrow==3.0.0;python_version<="3.6"
 google-cloud-bigquery
 tqdm
 deprecation
+pandas
```
douglasjacobsen, line 11: Should we split these out? I forget what the outcome of our discussion was the other day.

share/ramble/cloud-build/Dockerfile-apt
```diff
@@ -18,8 +18,9 @@ RUN cd /opt &&  \
     spack load py-pip ^python@${PYTHON_VER} && \
     wget https://raw.githubusercontent.com/GoogleCloudPlatform/ramble/develop/requirements.txt && \
     wget https://raw.githubusercontent.com/GoogleCloudPlatform/ramble/develop/requirements-dev.txt && \
-    python -m pip install -r /opt/requirements-dev.txt
-
+    python -m pip install -r /opt/requirements-dev.txt \
+    python -m pip install pandas \
```
douglasjacobsen: Are these pip install lines necessary? `requirements-dev` should pick them up if they are in the base requirements.txt file.

share/ramble/cloud-build/Dockerfile-yum
```diff
@@ -18,8 +18,9 @@ RUN cd /opt &&  \
     spack load py-pip ^python@${PYTHON_VER} && \
     wget https://raw.githubusercontent.com/GoogleCloudPlatform/ramble/develop/requirements.txt && \
     wget https://raw.githubusercontent.com/GoogleCloudPlatform/ramble/develop/requirements-dev.txt && \
-    python -m pip install -r /opt/requirements-dev.txt
-
+    python -m pip install -r /opt/requirements-dev.txt \
+    python -m pip install pandas \
```
douglasjacobsen: Are these pip install lines necessary? requirements-dev should pick them up if they are in the base requirements.txt file.

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+================
+Ramble Results
+================
+
+Ramble provides support and tooling around processing results data.
+
+See ``ramble workspace analyze`` (and associated :doc:`workspace documentation<workspace>`) for generating such data.
+
+Most notably, the two supported functionalities are:
+  * Uploading of arbitrary JSON
```
douglasjacobsen: "arbitrary but specifically formatted" or "arbitrary JSON that adheres to Ramble's output schema"

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+  * Generation of plots and analysis as part of a report
+
+-------------------
+Upload
+-------------------
+
+The `ramble results upload` command is used to imports experiment results from
```
douglasjacobsen: "imports" -> "import"

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+a JSON file, and uploads them as specified in the upload block of Ramble's
```
douglasjacobsen: "uploads" -> "upload"

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+config file.
```
douglasjacobsen: Maybe cross reference the relevant part of the config block?

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+
+This can be used for both results from an arbitrary ramble workspace, or for
+data that is generated by tools which conform to rambles data schema.
+
+-------------------
+Report
+-------------------
+The `ramble results report` command is experimental functionality that provides
+the user a way to generate plots and reports from the output of a workspace.
+This output is in the form of individual `png` plots, and a combined `pdf`
```
douglasjacobsen, line 39: Does single tick marks here format it the right way? I kind of expected it to be double tick marks, so this is more for my information that asking for a change...

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+  * Scaling (line) plots (both strong and weak)
+  * Comparison bar plots
+  * General FOM bar plots
+
+For these plots, common operations are supported such as log axis, data
+grouping, and idealized line plotting.
+
+While this functionality is very useful and powerful, it is not intended as a
+wholesale replacement for uploading the data into more advanced offline
+analysis tools.
```
douglasjacobsen: "at this time" ?

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Both commands take two sub-arguments:
+  * The performance metric: the FOM to plot on the y axis
+  * The scaling metric: the variable to plot on the x axis
+
+And example call may look something like:
+
+.. code-block:: console
+
+   ramble -d results report --strong-scaling 'simpleFoam Time' n_nodes -f ~/ramble-results.json
```
douglasjacobsen: Can we add an example of grouping by context as well?

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+
+(a very similar command should also work for `--weak-scaling`
```
douglasjacobsen: Missing `)`

lib/ramble/docs/results.rst
douglasjacobsen: Also, probably remove the `-d` argument

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+
+^^^^^^^^^^^^^^^^^^^
+Comparison and FOM plots
+^^^^^^^^^^^^^^^^^^^
+
+A comparison plot compares a specific FOM across all valid experiments, and generates a bar plot with one bar per experiment. It can be invoked by:
+
+.. code-block:: console
+
+    ramble -d results report --compare "Total Elapsed Time" -f ~/ramble-results.json
```
douglasjacobsen: Remove `-d`

lib/ramble/docs/results.rst
```diff
@@ -0,0 +1,86 @@
...
+
+A FOM plot gives the user a quick way to visualize all FOMS in a workspace as a bar chart, for general data discovery. It can be invoked by:
+
+.. code-block:: console
+
+    ramble -d results report --foms -f ~/ramble-results.json
```
douglasjacobsen: Remove `-d`

lib/ramble/docs/results.rst
douglasjacobsen: Also, can we add an example of normalizing the data? This is less important to me than the grouping though.

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,989 @@
...
+    Check for results in this order:
+        1. via ``ramble results report -f FILENAME``
+        2. via ``ramble -w WRKSPC`` or ``ramble -D DIR`` or
+        ``ramble results report --workspace WRKSPC``(arguments)
+        3. via a path in the ramble.workspace.ramble_workspace_var environment variable.
+    """
+    results_dict = {}
+
+    if args.file:
+        # FIXME: why does this call back into cmd?
```
douglasjacobsen: Do we still need this FIXME?

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,989 @@
...
+
+
+def is_repeat_child(experiment):
+    try:
+        if int(experiment["RAMBLE_VARIABLES"][keywords.repeat_index]) > 0:
+            return True
+        else:
+            return False
+    except KeyError:
+        # TODO: Remove this old implementation. I keep it here since the JSON update was so recent
```
douglasjacobsen: Still needed?

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,989 @@
...
+
+    Output (pd.DataFrame):
+             name   fom_name    fom_value  n_nodes
+        0   exp_1      fom_1         42.0        1
+        1   exp_1      fom_2           35        1
+        2   exp_2      fom_1         56.5        2
+        3   exp_2      fom_2           73        2
+    """
+
+    # TODO: is this sufficient?
```
douglasjacobsen: This seems unncessary?

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,989 @@
...
+    # panda = pd.json_normalize(results['experiments'])
+    # print(panda['CONTEXTS']) # would need to flatten contexts too
+    # Then remove things we don't need, like failures and repeats
+
+    unnest_context = []
+    skip_exps = []
+    # first unnest dictionaries
+    for exp in results["experiments"]:
+        # TODO: this does not work as expected for repeats since success is not clear
```
douglasjacobsen: Unnecessary?

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,989 @@
...
+
+                        # older data exports may not have fom_type stored
+                        if "fom_type" not in exp_copy:
+                            exp_copy["fom_type"] = FomType.UNDEFINED
+                            exp_copy[ReportVars.BETTER_DIRECTION.value] = (
+                                BetterDirection.INDETERMINATE
+                            )
+
+                    # Exclude vars that aren't needed for analysis, mainly paths and commands
+                    dir_regex = r"_dir$"
```
douglasjacobsen, line 185: Maybe these definitions should be higher, since they are static?

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,989 @@
...
+                            pdf_report,
+                            results_df,
+                            logx,
+                            logy,
+                            split_by,
+                        )
+                    )
+
+        if not plots:
+            # TODO: should this be checked in cmd?
```
douglasjacobsen: Still needed?

lib/ramble/ramble/reports.py
```diff
@@ -0,0 +1,989 @@
...
+                from_col=ReportVars.FOM_VALUE_MIN.value,
+            )
+            self.normalize_data(
+                selected_data,
+                scale_to_index=True,
+                to_col=ReportVars.FOM_VALUE_MAX.value,
+                from_col=ReportVars.FOM_VALUE_MAX.value,
+            )
+
+    # TODO: these args come from the spec, so don't need to be passed and could be stored at init
```
douglasjacobsen: Is this easy to do?

lib/ramble/ramble/cmd/results.py
rfbgo: Yea I think I agree, I will change and push

lib/ramble/ramble/test/data/reports/results.test.json
rfbgo: Not that I can think of. Do you have ideas?

lib/ramble/ramble/workspace/workspace.py
rfbgo: Agreed. Will make a task (internally)

requirements.txt
rfbgo: My suggestion was that we keep them here, but we can split them out if you prefer?

share/ramble/cloud-build/Dockerfile-apt
rfbgo: Good catch, will remove!

share/ramble/cloud-build/Dockerfile-yum
rfbgo: Good catch, will remove!

lib/ramble/docs/results.rst
rfbgo: Yes, as discussed offline single is more like italics, and double is more like "inline code block"

lib/ramble/ramble/reports.py
rfbgo: Removed :)

lib/ramble/docs/results.rst
rfbgo: Added context example, and a reference to `-n`. PTAL

lib/ramble/ramble/reports.py
rfbgo: I think this is actually no longer true, will remove

lib/ramble/ramble/reports.py
rfbgo: I think since they are only used here, it's OK to keep them local. I think if we need them elsewhere too that is the time to move them?

lib/ramble/docs/results.rst
rfbgo: I don't think this will ever totally be true? We will always want to support both CUJs? 

Or put another way this is never going to be so good we remove `--upload`?

douglasjacobsen: Definitely not, but I also think eventually it might be good enough that people who don't ever use `--upload` can do what they want with it.

For us, we always want `--upload` so we can have the results stored in a common place later.

douglasjacobsen (approved):

PR #721, opened 2024-10-25
Add expander support for NotIn

linsword13:


lib/ramble/ramble/expander.py
```diff
@@ -799,8 +799,10 @@ def _eval_comparisons(self, node):
         """Handle a comparison node in the ast"""
 
         # Extract In nodes, and call their helper
-        if len(node.ops) == 1 and isinstance(node.ops[0], ast.In):
-            return self._eval_comp_in(node)
+        if len(node.ops) == 1 and (
+            isinstance(node.ops[0], ast.In) or isinstance(node.ops[0], ast.NotIn)
+        ):
+            return self._eval_comp_in_or_not(node, reverse=isinstance(node.ops[0], ast.NotIn))
```
douglasjacobsen: Can the reverse logic be encapsulated in the eval method?

lib/ramble/ramble/expander.py
```diff
@@ -829,8 +831,8 @@ def _eval_comparisons(self, node):
         except KeyError:
             raise SyntaxError("Unsupported binary comparison operator")
 
-    def _eval_comp_in(self, node):
-        """Handle in nodes in the ast
+    def _eval_comp_in_or_not(self, node, reverse):
+        """Handle in and not in nodes in the ast
 
         Perform extraction of `<variable> in <experiment>` syntax.
         Raises an exception if the experiment does not exist.
```
douglasjacobsen, line 839: Assuming the reverse logic does not move into this method, can you update the doc string to describe the arguments?

douglasjacobsen (changes requested):

lib/ramble/ramble/expander.py
linsword13: Oh right indeed, that'd be much simpler. Updated.

lib/ramble/ramble/expander.py
linsword13: Moved the logic up.

douglasjacobsen (approved):

PR #738, opened 2024-10-31
Allow `required_variable` to only enforce on certain modes

linsword13:
gcp-metadata mod can use this, as it does not need `hostlist` when run in local mode.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -829,7 +829,7 @@ class ApplicationBase(metaclass=ApplicationMeta):
             self._modifier_instances.append(mod_inst)
 
             # Add this modifiers required variables for validation
-            self.keywords.update_keys(mod_inst.required_vars)
+            self.keywords.update_keys(mod_inst.get_required_variables())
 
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/application.py
```diff
@@ -829,7 +829,19 @@ def build_modifier_instances(self):
             self._modifier_instances.append(mod_inst)
 
             # Add this modifiers required variables for validation
-            self.keywords.update_keys(mod_inst.required_vars)
+            required_vars = mod_inst.required_vars
```
douglasjacobsen: I think I'd prefer if we push the logic to extract required variables into the modifier rather than having the application class pull them out directly.

douglasjacobsen (changes requested):
This is great, thanks! I've actually wanted this for a while, so it's nice to see it would help someone else. :)

I have one small request, but otherwise this LGTM.

lib/ramble/ramble/application.py
linsword13: Good point, updated.

douglasjacobsen (approved):

PR #741, opened 2024-11-01
Add nemo executable to create transformer cache

douglasjacobsen:
This merge adds an executable to the py-nemo app which will ensure the transformer cache is created prior to executing the nemo experiment.

### DIFF
--- a/var/ramble/repos/builtin/applications/py-nemo/application.py
+++ b/var/ramble/repos/builtin/applications/py-nemo/application.py
@@ -1371,38 +1371,39 @@ class PyNemo(ExecutableApplication):
 
         final_regex = re.compile(self.final_epoch_regex)
 
-        with open(log_file, "r", encoding="ISO-8859-1") as f:
-            data = f.read()
+        if os.path.exists(log_file):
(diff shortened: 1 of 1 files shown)


rfbgo (dismissed):

var/ramble/repos/builtin/applications/py-nemo/application.py
```diff
@@ -27,6 +27,12 @@ class PyNemo(ExecutableApplication):
 
     tags("ml-framework", "machine-learning")
 
+    executable(
+        "setup_transformer_cache",
+        'bash -c "python3 -c \'from transformers import AutoTokenizer; AutoTokenizer.from_pretrained(\\"gpt2\\")\'"',
```
linsword13, line 32: Is this setup applicable to non-gpt models?

douglasjacobsen: I think so, but I am not the expert on this.

I think all this does is allow the transformers module to setup a cache that it will store transformers in.

linsword13 (dismissed):

linsword13 (dismissed):

linsword13 (approved):

PR #742, opened 2024-11-01
Add a `ramble workspace logs` command

douglasjacobsen:
This merge adds a new `ramble workspace logs` command which can be used to get information about logs in experiments.

It includes flags to easily filter out only failed experiments as well.

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -51,6 +51,7 @@ subcommands = [
     "info",
     "edit",
     "mirror",
+    "experiment-logs",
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -51,6 +51,7 @@
     "info",
     "edit",
     "mirror",
+    "experiment-logs",
```
linsword13, line 54: May be for later, but do you think there's value to make this subcommand look for logs from other pipelines (setup, archive, etc.)?

linsword13 (approved):

PR #755, opened 2024-11-07
Add an apptainer modifier

douglasjacobsen:
This merge adds a modifier which can be used to create workflows using apptainer for experiments.

### DIFF
--- a/var/ramble/repos/builtin/modifiers/apptainer/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/apptainer/modifier.py
@@ -1,0 +1,289 @@
+# Copyright 2022-2024 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import re
+import os
+
+from ramble.modkit import *
+from ramble.util.hashing import hash_string
+
+import llnl.util.filesystem as fs
+
+
+class Apptainer(BasicModifier):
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/modifiers/apptainer/modifier.py
```diff
@@ -0,0 +1,289 @@
...
+    name = "apptainer"
+
+    tags("container")
+
+    maintainers("douglasjacobsen")
+
+    mode("standard", description="Standard execution mode for apptainer")
+    default_mode("standard")
+
+    required_variable("container_name")
```
linsword13, line 40: nit: since this is likely the only place where these two variables are "defined", would be nice to have some descriptions on them.

var/ramble/repos/builtin/modifiers/apptainer/modifier.py
```diff
@@ -0,0 +1,289 @@
...
+            expanded_exp_mount = app_inst.expander.expand_var(exp_mount)
+
+            if (
+                exp_mount not in input_mounts
+                and expanded_exp_mount not in input_mounts
+            ):
+                add_mod = self._usage_mode not in self.variable_modifications
+                add_mod = (
+                    add_mod
+                    or self._usage_mode in self.variable_modifications
```
linsword13: nit: perhaps it would be easier to parse by adding brackets around the `and` clause?

var/ramble/repos/builtin/modifiers/apptainer/modifier.py
douglasjacobsen: Added comments, and made a task to add descriptions to required variables, since that doesn't exist yet.

var/ramble/repos/builtin/modifiers/apptainer/modifier.py
douglasjacobsen: Ack.

linsword13 (approved):

PR #757, opened 2024-11-08
Add a nccl-env modifier to manage NCCL environment variables

douglasjacobsen:


var/ramble/repos/builtin/modifiers/nccl-env/modifier.py
```diff
@@ -0,0 +1,714 @@
+# Copyright 2022-2024 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.modkit import *  # noqa: F403
```
linsword13: nit: the noqa comment should not be needed.

var/ramble/repos/builtin/modifiers/nccl-env/modifier.py
```diff
@@ -0,0 +1,714 @@
...
+        "nccl_nsocks_perthread",
+        default="",
+        description="Specification of number of sockets opened by each helper thread.",
+        modes=["standard"],
+    )
+
+    modifier_variable(
+        "nccl_cross_nic",
+        default="",
+        values=[0, 1, 2],
```
linsword13, line 71: Does this essentially make it a required variable?

douglasjacobsen: No, this just allows `ramble info --type modifiers nccl-env` to print "suggested values" for this variable. It's more for documentation than any kind of enforcement.

var/ramble/repos/builtin/modifiers/nccl-env/modifier.py
douglasjacobsen: Thanks, fixed.

linsword13 (approved):

PR #759, opened 2024-11-09
Add a modifier for GCP cloud logging

douglasjacobsen:
This merge adds a modifier which will upload logs into Google Cloud Platform's Cloud Logging.

### DIFF
--- a/var/ramble/repos/builtin/modifiers/gcp-cloud-logging/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/gcp-cloud-logging/modifier.py
@@ -1,0 +1,71 @@
+# Copyright 2022-2024 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.modkit import *  # noqa: F403
+from ramble.util.shell_utils import last_pid_var
+
+
+class GcpCloudLogging(BasicModifier):
+    """Upload experiment logs to Google Cloud Platform Cloud Logging.
+
+    https://cloud.google.com/logging?e=48754805&hl=en
+    """
+
+    name = "gcp-cloud-logging"
+
+    tags("info")
+
+    maintainers("douglasjacobsen")
+
+    mode("standard", description="Standard execution mode")
+
+    default_mode("standard")
+
+    target_shells("bash")
+
+    modifier_variable(
+        "cloud_logging_tag",
+        default="{experiment_namespace}",
+        description="Tag to prefix cloud logging entries with",
+        modes=["standard"],
+    )
+
+    modifier_variable(
+        "cloud_logging_status_file",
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/modifiers/gcp-cloud-logging/modifier.py
```diff
@@ -0,0 +1,43 @@
...
+    default_mode("standard")
+
+    register_builtin(
+        "start_cloud_logger", required=True, injection_method="prepend"
+    )
+
+    def start_cloud_logger(self):
+        return [
+            "tail -f {log_file} | logger -t {experiment_namespace} &",
+            "export LOGGER_PID=$!",
```
linsword13: nit: there's a `last_pid_var` util in shell_utils.py that can be used here.

var/ramble/repos/builtin/modifiers/gcp-cloud-logging/modifier.py
```diff
@@ -0,0 +1,43 @@
...
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.modkit import *  # noqa: F403
+
+
+class GcpCloudLogging(BasicModifier):
+    """Upload experiment logs to Google Cloud Platform Cloud Logging.
```
linsword13, line 14: Just curious, is cloud-logging something that gets configured upon cluster creation? Would we want to support some configurability in the modifier? One thing I may want to configure is how frequently the log gets uploaded, and/or if the logging agent has some knobs on resource control. Mostly wonder if there could be some performance hit when this is enabled.

var/ramble/repos/builtin/modifiers/gcp-cloud-logging/modifier.py
douglasjacobsen: Thanks, fixed!

I also moved the `'{experiment_namespace}'` to a modifier variable, since people probably want to be able to control this.

var/ramble/repos/builtin/modifiers/gcp-cloud-logging/modifier.py
douglasjacobsen: I haven't explored the performance impact yet, but I don't know of any user-side control of the cloud-logging agent.

I can research this some more, but this was mostly to support other workflows within our organization.

douglasjacobsen: I ran some performance impact studies which I can share with you on Monday, but this has no noticeable impact on the performance of the workloads I ran.

linsword13: Thanks for checking that! One last annoying question: do you see the need to verify the logging agent is actually up and running? I imagine something like `service google-fluentd status`, although I don't really know if that's the right agent service.

linsword13 (dismissed):

var/ramble/repos/builtin/modifiers/gcp-cloud-logging/modifier.py
douglasjacobsen: Done. :)

linsword13 (approved):

PR #764, opened 2024-11-12
Defend against -1 values for n_nodes (and update examples)

rfbgo:


douglasjacobsen (changes requested):
I think we probably should have the default `ramble.yaml` (defined in workpsace.py) updated as well, right?

rfbgo:
> I think we probably should have the default `ramble.yaml` (defined in workpsace.py) updated as well, right?

It actually was by you ~2 months ago, but we missed a few other changes like examples :)

https://github.com/GoogleCloudPlatform/ramble/commit/00acbac319acabd6f7f32bba95cd01c8d126c331

douglasjacobsen:
Ah, OK! Thanks! :)

douglasjacobsen (approved):

PR #768, opened 2024-11-14
Allow description for `required_variable` directive

linsword13:
Update several required vars with descriptions.
Verified that `ramble info -v` prints out the description.

### DIFF
--- a/var/ramble/repos/builtin/modifiers/conditional-psm3/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/conditional-psm3/modifier.py
@@ -38,7 +38,11 @@ class ConditionalPsm3(BasicModifier):
 
     executable_modifier("apply_psm3")
 
-    required_variable("psm3_mpi")
+    required_variable(
+        "psm3_mpi",
+        description="The name of the package used for MPI "
+        "within an experiment's software environment",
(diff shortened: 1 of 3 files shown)


var/ramble/repos/builtin/modifiers/conditional-psm3/modifier.py
```diff
@@ -38,7 +38,9 @@ class ConditionalPsm3(BasicModifier):
 
     executable_modifier("apply_psm3")
 
-    required_variable("psm3_mpi")
+    required_variable(
+        "psm3_mpi", description="The MPI defined in the software environment"
```
douglasjacobsen: nit; Maybe "The name of the package used for MPI within an experiment's software environment" ?

Basically, it's unclear (to me) from this description that the variable should be the name of an MPI package.

douglasjacobsen (changes requested):

var/ramble/repos/builtin/modifiers/conditional-psm3/modifier.py
linsword13: Great call, updated.

douglasjacobsen (approved):

PR #774, opened 2024-11-19
Add argument to clean command to remove reports

dapomeroy:
Adds `ramble clean --reports` to remove pdf and image files generated by the reports feature.
```
(base) foo@user:~/ramble$ ramble clean -r
==> Removing pdf and image files generated by ramble reports
==> Really remove all reports and images from /usr/local/foo/.ramble/reports? [y/N] n
==> Error: Will not remove any files
(base) foo@user:~/ramble$ ramble clean -r
==> Removing pdf and image files generated by ramble reports
==> Really remove all reports and images from /usr/local/foo/.ramble/reports? [y/N] y
(base) foo@user:~/ramble$ 
```


lib/ramble/ramble/cmd/clean.py
```diff
@@ -78,3 +89,12 @@ def remove_python_caches():
...
                     shutil.rmtree(dname)
+
+
+def remove_reports_files():
+    reports_path = ramble.reports.get_reports_path()
+    if reports_path:
+        for dir in os.listdir(reports_path):
+            dname = os.path.join(reports_path, dir)
+            logger.debug(f"Removing {dname}")
+            shutil.rmtree(dname)
```
rfbgo: Do we think this is sufficiently safe? It seems a bit like we could accidentally delete stuff that was not intended?

It feels like overkill but maybe we should do something more like delete pngs, then empty dirs, then pdfs?

dapomeroy: That sounds reasonable. Do you think we should add a "really delete reports? [y/N]" prompt like we have for workspaces?

rfbgo: Yea, I think we should follow the workflow in workspaces closely where possible 

In general I'm pretty shy of deleting files indiscriminately, a user could run `rm -rf $RAMBLE_REPORTS_DIR` or something easily if that's what they want

dapomeroy: OK, I've updated this to only delete pdf and png files and empty directories, and to ask y/N first

PR #777, opened 2024-11-21
Add argument to 'clean' command to remove reports files

dapomeroy:
Adds `ramble clean --reports` to remove pdf and image files generated by the reports feature.
```
(base) foo@user:~/ramble$ ramble clean -r
==> Removing pdf and image files generated by ramble reports
==> Really remove all reports and images from /usr/local/foo/.ramble/reports? [y/N] n
==> Error: Will not remove any files
(base) foo@user:~/ramble$ ramble clean -r
==> Removing pdf and image files generated by ramble reports
==> Really remove all reports and images from /usr/local/foo/.ramble/reports? [y/N] y
(base) foo@user:~/ramble$ 
```


lib/ramble/ramble/reports.py
```diff
@@ -326,6 +330,23 @@ def add_minmax_data(self, selected_data, min_data, max_data, scale_var):
...
+        """
+        try:
+            with open(self.inventory_path) as f:
+                inventory = syaml.load(f)
+        except FileNotFoundError:
+            inventory = {"files": []}
+
+        inventory["files"].append(filename)
+
+        with open(self.inventory_path, "w+") as f:
```
douglasjacobsen: Instead of reading and writing the inventory file for each file we write, can we batch the updates to the inventory file?

Doing this with a large report (i.e. lots of png / pdf files) could cause a lot of thrash on a shared filesystem.

douglasjacobsen (changes requested):

lib/ramble/ramble/reports.py
dapomeroy: I've updated it to only write once per inventory file, and I think it can skip the 'read' and just initialize a new empty inventory because each new plot generator / reports folder has a unique timestamped path created on invocation.

douglasjacobsen (approved):

lib/ramble/ramble/reports.py
douglasjacobsen: Awesome, thanks!

PR #782, opened 2024-12-02
Update application.py

napulath:
Removing hydra.searchpath from application.py


douglasjacobsen:
This is included in #783

PR #784, opened 2024-12-03
Fix env var propagation between experiments

douglasjacobsen:
This merge fixes an issue where environment variables were incorrectly propagated between experiments. A test has also been added to help ensure this issue doesn't show up again.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -967,26 +967,21 @@ class ApplicationBase(metaclass=ApplicationMeta):
         if self.expander.workload_name in self.workloads:
             workload = self.workloads[self.expander.workload_name]
 
+            new_env_vars = {}
             for name, env_var in workload.environment_variables.items():
                 action = "set"
                 value = env_var.value
 
-                # Since the type coming from the schema can either be None, or
-                # a complex polymorphic type we need to ensure it has a
-                # sensible base structure when it is not given
-                if not self._env_variable_sets:
-                    self._env_variable_sets.append({"set": {}})
-
-                # Since the type coming from the schema can either be None, or
-                # a complex polymorphic type we need to ensure it has a
-                # sensible base structure when it is not given
-                if not self._env_variable_sets:
-                    self._env_variable_sets.append({"set": {}})
-
+                add = True
                 for env_var_set in self._env_variable_sets:
                     if action in env_var_set:
-                        if env_var.name not in env_var_set[action].keys():
-                            env_var_set[action][env_var.name] = value
+                        if env_var.name in env_var_set[action].keys():
+                            add = False
+
+                if add:
+                    new_env_vars[env_var.name] = value
+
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/application.py
```diff
@@ -967,26 +967,21 @@ def _set_default_experiment_variables(self):
...
+                add = True
                 for env_var_set in self._env_variable_sets:
                     if action in env_var_set:
-                        if env_var.name not in env_var_set[action].keys():
-                            env_var_set[action][env_var.name] = value
+                        if env_var.name in env_var_set[action].keys():
+                            add = False
+
+                if add:
+                    new_env_vars[env_var.name] = value
```
linsword13, line 982: What's causing the leak? Is it due to pollution of class attributes? Modification within the loop?

douglasjacobsen: I think there's two parts. One that I haven't quite figured out yet (because it doesn't seem strictly wrong to me) and one that this PR fixes.

Fundamentally, it appears that the env-vars defined in application.py files end up in the workspace context (https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/experiment_set.py#L65).

I suspect that somewhere this context's env-var set is being added by reference rather than having a copy added (the RHS of the previous link returns the `ramble.config.get.get("env_vars")`, and I tested adding this with `.copy()` with no success).

The second part is that in the code before this PR, application.py environment variables are incorrectly added to *every* env-var set within the application, as opposed to only being added once. This means that if the workspace env-vars are added by reference, the application's env-vars are then injected into this env-var set (and that persists after the experiment is complete).

This PR fixes the second of these issues by only adding application defined env-vars once, in it's own env-var set, at the end of the list of env-var sets. One change that this makes that is different from previous behavior is that application defined env-vars are only defined if they are not defined in a separate env-var set. For example, if an application has:

```
environment_variable("MY_NEW_VAR", "new_val")
```

And a `ramble.yaml` has:
```
ramble:
  env_vars:
    set:
      MY_NEW_VAR: 'other_val'
```

Then the resulting experiment (that uses the application) would not have `export MY_NEW_VAR="new_val"` but would have `export MY_NEW_VAR="other_val"`.

With the previous behavior, the resulting experiment would have (from what I can tell) both exports, with `export MY_NEW_VAR="new_val"` likely taking precedence and potentially occurring multiple times.

EDIT: Just FYI, when I said env-vars in the `application.py` end up in the workspace context, I should have added that this is only the case if the workspace has a top level env-vars section.

If the workspace does not have:
```
ramble:
  env_vars:
    ...
```
Then this does not happen. So I suspect this would occur with any env-vars from an application as long as the workspace (or higher level) has an env-vars section defined.

linsword13: Thanks for the detailed explanation!

linsword13 (approved):

PR #789, opened 2024-12-05
Add a slurm workflow manager

linsword13:
Example usage:
    
```yaml
ramble:
  variants:
    workflow_manager: slurm
  variables:
    processes_per_node: 1
  applications:
    hostname:
      workloads:
        local:
          experiments:
            test:
              variables:
                n_nodes: 1
```

```
  # The batch_submit is defined by the workflow object
ramble on

  # Executors for query and cancel are available
ramble on --executor "{batch_query}"
ramble on --executor "{batch_cancel}"

  # The status report is also available in analyze
ramble workspace analyze -p
```

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -66,7 +66,7 @@ from enum import Enum
 
 experiment_status = Enum(
     "experiment_status",
-    ["UNKNOWN", "SETUP", "RUNNING", "COMPLETE", "SUCCESS", "FAILED", "CANCELLED"],
+    ["UNKNOWN", "SETUP", "SUBMITTED", "RUNNING", "COMPLETE", "SUCCESS", "FAILED", "CANCELLED"],
 )
 
 _NULL_CONTEXT = "null"
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -65,7 +65,8 @@ from ramble.util.output_capture import output_mapper
 from enum import Enum
 
 experiment_status = Enum(
-    "experiment_status", ["UNKNOWN", "SETUP", "RUNNING", "COMPLETE", "SUCCESS", "FAILED"]
+    "experiment_status",
+    ["UNKNOWN", "SETUP", "RUNNING", "COMPLETE", "SUCCESS", "FAILED", "CANCELLED"],
 )
 
 _NULL_CONTEXT = "null"
@@ -1745,10 +1746,13 @@ class ApplicationBase(metaclass=ApplicationMeta):
                     success = True
         success = success and criteria_list.passed()
 
-        if success:
-            self.set_status(status=experiment_status.SUCCESS)
-        else:
-            self.set_status(status=experiment_status.FAILED)
+        status = experiment_status.SUCCESS if success else experiment_status.FAILED
+        # When workflow_manager is present, only use app_status when workflow is completed.
+        if self.workflow_manager is not None:
+            wm_status = self.workflow_manager.get_status(workspace)
+            if not wm_status == experiment_status.COMPLETE:
+                status = wm_status
+        self.set_status(status)
 
         self._init_result()
 
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -181,6 +181,8 @@ class ApplicationBase(metaclass=ApplicationMeta):
         self.license_path = ""
         self.license_file = ""
 
+        self.workflow_manager = None
+
         ramble.util.directives.define_directive_methods(self)
 
     def experiment_lock(self):
@@ -236,7 +238,10 @@ class ApplicationBase(metaclass=ApplicationMeta):
                              experiment.
         """
         self.variants = variants.copy()
+        self._set_package_manager()
+        self._set_workflow_manager()
 
+    def _set_package_manager(self):
         if namespace.package_manager in self.variants:
             pkgman_name = self.expander.expand_var(
                 self.variants[namespace.package_manager], typed=True
@@ -275,6 +280,24 @@ class ApplicationBase(metaclass=ApplicationMeta):
                                 "level": ramble.keywords.output_level.variable,
                             }
                         }
+                    )
+
+    def _set_workflow_manager(self):
+        if namespace.workflow_manager in self.variants:
+            workflow_name = self.expander.expand_var(
+                self.variants[namespace.workflow_manager], typed=True
+            )
+
+            if workflow_name is not None:
+                try:
+                    wfman_type = ramble.repository.ObjectTypes.workflow_managers
+                    self.workflow_manager = ramble.repository.get(workflow_name, wfman_type).copy()
+                    self.workflow_manager.set_application(self)
+                except ramble.repository.UnknownObjectError:
+                    logger.die(
+                        f"{workflow_name} is not a valid workflow manager. "
+                        "Valid workflow managers can be listed via:\n"
+                        "\tramble list --type workflow_managers"
                     )
 
     def build_phase_order(self):
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -502,6 +502,10 @@ class ApplicationBase(metaclass=ApplicationMeta):
             for var in self.package_manager.package_manager_variables.values():
                 self.variables[var.name] = var.default
 
+        if self.workflow_manager is not None:
+            for var in self.workflow_manager.wm_vars.values():
+                self.variables[var.name] = var.default
+
         ##########################################
         # Expand used variables to track all usage
         ##########################################
@@ -964,6 +968,9 @@ class ApplicationBase(metaclass=ApplicationMeta):
 
         for mod_inst in self._modifier_instances:
             var_sets.append(mod_inst.mode_variables())
+
+        if self.workflow_manager is not None:
+            var_sets.append(self.workflow_manager.wm_vars)
 
         for var_set in var_sets:
             for var, val in var_set.items():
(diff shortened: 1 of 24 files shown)


var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -0,0 +1,108 @@
...
+    """Slurm workflow manager"""
+
+    name = "slurm"
+
+    maintainers("linsword13")
+
+    tags("workflow", "slurm")
+
+    render_content(
+        name="batch_submit",
```
douglasjacobsen, line 82: How does this work when we define `batch_submit` in the variables? Does this override the definition of the `batch_submit` variable?

linsword13: Yeah it does override the `batch_submit` variable. I do that so that `ramble on` works without needing to specify an `executor`. Do you see this being undesirable in some cases?

douglasjacobsen: Nope, I was more concerned that the variable would override the template.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -0,0 +1,197 @@
...
+#!/bin/bash
+{_ensure_job_id_snippet}
+{self.runner.generate_cancel_command("$job_id")}
+        """.strip()
+
+    render_content(
+        name="slurm_execute_experiment", content_func="_slurm_execute_script"
+    )
+
+    def _slurm_execute_script(self):
```
douglasjacobsen: Can we maybe move this into the base class, as a general "Build pragmas" function, just since I think this will be used fairly frequently.

Maybe it should be something other than pragmas specifically though... more like something around conditional expansion?

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def set_run_dir(self, run_dir):
+        """
+        Set the experiment_run_dir of this runner
+        """
+        self.run_dir = run_dir
+
+    def generate_query_command(self, job_id):
+        return rf"""
+status=$(squeue -h -o "%t" -j {job_id})
```
douglasjacobsen: I'm not sure how picky I am about this, but I would almost prefer the status map be in python rather than in bash.

And maybe some of the logic mapped into the base class too.

lib/ramble/ramble/test/conftest.py
```diff
@@ -187,18 +199,25 @@ def mock_modifiers(mock_mods_repo_path):
...
-        mock_mods_repo_path, object_type=obj_type
-    ) as mock_mods_repo:
-        yield mock_mods_repo
+        mock_pkg_mans_repo_path, object_type=obj_type
+    ) as mock_pkg_mans_repo:
+        yield mock_pkg_mans_repo
+
+
+@pytest.fixture(scope="function")
+def mock_workflow_managers(mock_wms_repo_path):
```
douglasjacobsen, line 212: Not related to this PR, mostly just something to think about.

Do you have any ideas on how we can clean up this stuff, and make it so we don't have to define new fixtures for each object type?

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -0,0 +1,197 @@
...
+        default="{application_name}_{workload_name}_{experiment_name}",
+        description="Slurm job name",
+    )
+
+    workflow_manager_variable(
+        name="extra_sbatch_headers",
+        default="",
+        description="Extra sbatch headers added to slurm job script",
+    )
+
```
douglasjacobsen, line 62: Can you add a `workflow_manager_variable` to define `hostlist` as `$SLURM_JOB_NODELIST` too? That might make composition with modifiers that require this to be define a bit easier.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -0,0 +1,197 @@
...
+    def _slurm_execute_script(self):
+        headers = ["#!/bin/bash"]
+        expander = self.app_inst.expander
+        # Adding pre-defined sets of headers
+        pragmas = [
+            ("#SBATCH -N {}", "n_nodes"),
+            ("#SBATCH -p {}", "partition"),
+            ("#SBATCH --ntasks-per-node {}", "processes_per_node"),
+            ("#SBATCH -J {}", "job_name"),
```
douglasjacobsen: Can you also add `--gpus-per-node` = `{gpus_per_node}` here?

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -0,0 +1,197 @@
...
+
+    maintainers("linsword13")
+
+    tags("workflow", "slurm")
+
+    def __init__(self, file_path):
+        super().__init__(file_path)
+
+        self.runner = SlurmRunner()
+
```
douglasjacobsen, line 50: We should also think about how to allow this modifier to have a default `mpi_command` (that could still be overridden)

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Definitely, added.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Great suggestion, added.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Makes sense, updated.

lib/ramble/ramble/test/conftest.py
linsword13: We can perhaps have a fixture that defaults to all the builtin paths, and can be overridden to point to builtin.mock path for specific object type. But that makes the test itself slightly more verbose, so not sure if that's much of an improvement.

douglasjacobsen: OK cool, thanks. Maybe we can keep this in the back of our minds to see if we can make this nicer in the future.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Good call. I moved the get_status call inside the runner to be in python instead of relying on this bash script, so now `ramble workspace analyze` would rely on python. Do you prefer even the `--executor "{batch_query}"` to be handled in python instead of the bash script right now?

linsword13:
Marking this as a draft, given that this PR will need to be updated if https://github.com/GoogleCloudPlatform/ramble/pull/804 is in.

douglasjacobsen:
Something I just found out, if a job is completed and you run `ramble on --executor='{query_job}'` then it'll say the job failed (even if you haven't analyzed it.

We might want to see if there's a way to use something like scontrol to get the status of completed jobs.

linsword13:
> Something I just found out, if a job is completed and you run `ramble on --executor='{query_job}'` then it'll say the job failed (even if you haven't analyzed it.
> 
> We might want to see if there's a way to use something like scontrol to get the status of completed jobs.

Hm that's unexpected. I did have a `sacct` command to try to query for status of completed jobs, if `squeue` comes up empty. I will look into it.

douglasjacobsen:
Let me try it again, it might be that analyzing the job failed for some reason.

douglasjacobsen:
Yeah, I think I just had a weird issue with my experiments.

Can we also add default `-o` and `-e` pragmas to the generated submission script? This can place the slurm logs into the experiment directory.

linsword13:
> Yeah, I think I just had a weird issue with my experiments.
> 
> Can we also add default `-o` and `-e` pragmas to the generated submission script? This can place the slurm logs into the experiment directory.

Thanks for trying out! Also added the `-o` and `-e`.

linsword13:
> Marking this as a draft, given that this PR will need to be updated if #804 is in.

This is now updated. It currently uses changes in https://github.com/GoogleCloudPlatform/ramble/pull/814.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Added. It works to add that as a `workflow_variable`, and that's overridable.

lib/ramble/ramble/application.py
```diff
@@ -65,7 +65,8 @@
 from enum import Enum
 
 experiment_status = Enum(
-    "experiment_status", ["UNKNOWN", "SETUP", "RUNNING", "COMPLETE", "SUCCESS", "FAILED"]
+    "experiment_status",
+    ["UNKNOWN", "SETUP", "RUNNING", "COMPLETE", "SUCCESS", "FAILED", "CANCELLED"],
```
douglasjacobsen: Can we add a "SUBMITTED" status between "SETUP" and "RUNNING?

douglasjacobsen (changes requested):
This is looking great. One small request. :)

lib/ramble/ramble/application.py
linsword13: Definitely, added.

douglasjacobsen (dismissed):

douglasjacobsen (approved):

PR #798, opened 2024-12-12
Fix HPL preventing variable overrides from workspace config

douglasjacobsen:
This merge fixes an issue where variables defined in `ramble.yaml` would not propagate into the HPL.dat.

As an example, `"variables:swapping_threshold:999"` would not show up in the generated HPL.dat file when the calculator workload was used. After this chage, these settings propagate correctly.

Additionally, some logic around defining HPL settings is cleaned up to make it easier to maintain in the future.


rfbgo (approved):

samskillman:
Confirmed this addresses the issue

samskillman (approved):

PR #799, opened 2024-12-12
Allow typed dictionaries, and dictionary subscripts from expander

douglasjacobsen:
This merge enables the expander to handle dictionaries better.

The most obvious change from this is that given:
```
variables:
  test_dict:
    key1: val1
    key2: val2
```
then `'{test_dict["key1"]}'` should expand to `"val1"`.

A more subtle change from this is that given the variable definition above, inside object definitions one can now do:
```
test_dict = app_inst.expander.expand_var_name("test_dict", typed=True)

for key, val in test_dict.items():
  print(f"Key {key} = {val}")
```

Tests and documentation have also been added for this.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -334,6 +334,8 @@ class Expander:
     Additionally, math will be evaluated as part of expansion.
     """
 
+    _ast_dbg_prefix = "EXPANDER AST:"
+
     def __init__(self, variables, experiment_set, no_expand_vars=set()):
 
         self._keywords = ramble.keywords.keywords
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/expander.py
```diff
@@ -906,24 +906,41 @@ def _eval_unary_ops(self, node):
...
+                    step = _get_with_default(slice_node, "step", 1)
+                    return operand[slice(lower, upper, step)]
+                elif operand in self._variables and isinstance(self._variables[operand], dict):
+                    op_dict = self.expand_var_name(operand, typed=True)
+
+                    # TODO: Remove after support for python 3.9 is dropped
+                    # DEPRECATED: ast.Index was dropped in python 3.9
+                    if hasattr(ast, "Index") and isinstance(slice_node, ast.Index):
+                        key = self.eval_math(slice_node.value)
+                        return op_dict[key]
```
linsword13: What happens when key is not in the dict?

douglasjacobsen: Ah good catch. Currently, you get a `KeyError`, but I don't think that's what we want.

I'll work on fixing this.

douglasjacobsen: Fixed!

linsword13 (approved):

PR #803, opened 2024-12-18
Add new yaml files to `ramble workspace edit`

dapomeroy:
Adds ability for `ramble workspace edit` to open applications.yaml, modifiers.yaml, and software.yaml within the workspace config directory.


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -881,12 +908,28 @@ def workspace_edit(args):
         )
 
     config_file = ramble.workspace.config_file(ramble_ws)
+    applications_file = ramble.workspace.applications_file(ramble_ws)
+    modifiers_file = ramble.workspace.modifiers_file(ramble_ws)
+    software_file = ramble.workspace.software_file(ramble_ws)
```
linsword13: Are there potentially more such "include" files that we want to be able to edit? Would it make sense to have an `--all` option to open up all yaml files under the config directory, and then maybe another `--name` option to allow opening up specific ones (like `ramble workspace edit --name applications` to edit the applications.yaml file?)

dapomeroy: I'd love some feedback from Bob/Doug here. Yes, users could definitely add more files than just these and I think your two suggestions are good additions. It might also be nice for '--all' to open included files as well.

And as I'm using Ramble myself, I don't want to have to type out all those chars, which don't tab complete, to open just the 'applications.yaml' file. I like the convenience of having '-a' short flag for it. Do you think that having all these individual flags and options is making the interface too complex, or is it ok?

rfbgo: In general I like and agree with all of the suggestion from both.  I think `--all` is good idea too (but definitely don't want that to be the default behavior) 

Slight preference for `-a` over `--type applications` personally, but am OK with either 

My overall take is that we should make the `top(n)` operations quick and easy, but we do not have to solve complex or edge cases and the user should `cd $RAMBLE_WORKSPACE; vim ...` for that

dapomeroy: Thanks Bob. The way it's written now, it keeps current behavior of opening ramble.yaml and templates by default. If they exist, it also opens applications, modifiers, and software. Adds flags for those 3 to support quick editing of them. Why don't I add in '--all' to cover cases where someone may use other filenames, and leave it there. Nothing changes in the workflow for existing users with just ramble.yaml, it works automatically if a user chooses to create app/mod/software files, and adds an easy option to open every yaml just in case someone wants it.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -292,6 +292,18 @@ def licenses_file(path):
     return get_yaml_filepath(path, licenses_file_name)
 
 
+def all_config_files(path):
+    """Returns path to all yaml files in workspace config directory"""
+    config_files = []
+
+    config_path = os.path.join(path, workspace_config_path)
+    for f in os.listdir(config_path):
+        if f.endswith(".yaml"):
```
rfbgo, line 283: Is this right? Do we only care about `.yaml` or do we also need to compare `.tpl` files?

dapomeroy: I was thinking `--all` was all config files, but that's a good point. It's probably better to have it be all of the above, including template files

rfbgo: Yea to me it feels like something might be "wrong" if `ramble workspace edit` gives you more files than `--all`

dapomeroy: OK, I've updated it. This method still only returns the yaml files because there's already an 'all template' method, so I combined the two in cmd instead of here.

linsword13 (dismissed):

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: I think I like the `--all` addition, but I'd almost prefer we had something like:
`ramble workspace edit --file <file_name>`

e.g.
`ramble workspace edit --file applications.yaml`
`ramble workspace edit --file test.tpl`
`ramble workspace edit --file auxiliary_software_files/concretizer.yaml`

I don't really like the `--applications-only` / `--modifiers-only` flags, as that implies to me that every configuration section needs its own `--*-only` flag to edit it easily.

douglasjacobsen (changes requested):

lib/ramble/ramble/cmd/workspace.py
dapomeroy: OK, I've removed the 3 filesnames/args and added the --file option

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -883,15 +900,20 @@ def workspace_edit(args):
     config_file = ramble.workspace.config_file(ramble_ws)
     template_files = ramble.workspace.all_template_paths(ramble_ws)
 
-    edit_files = [config_file] + template_files
+    edit_files = [config_file]
+    edit_files.extend(template_files)
 
-    if args.config_only:
+    if args.filename:
+        edit_files = [ramble.workspace.get_yaml_filepath(ramble_ws, args.filename)]
```
douglasjacobsen: It might be worth changing the name of `get_yaml_filepath` to be more like `get_filepath` since it could be any file (like a `*.tpl` file)

douglasjacobsen (changes requested):

lib/ramble/ramble/cmd/workspace.py
dapomeroy: This is done

douglasjacobsen (approved):
Thanks for the changes!

PR #804, opened 2024-12-24
Add a `register_template` directive to objects

linsword13:
This directive allows for specifying per-experiment template rendering, with the template defined as separate files alongside the main object source.

The testing app `template` is defined as an example.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -70,6 +70,8 @@ experiment_status = Enum(
 
 _NULL_CONTEXT = "null"
 
+_DEFAULT_CONTENT_PERM = stat.S_IRWXU | stat.S_IRWXG | stat.S_IROTH | stat.S_IXOTH
+
 
 def _get_context_display_name(context):
     return (
@@ -1173,6 +1175,7 @@ class ApplicationBase(metaclass=ApplicationMeta):
             self._set_input_path()
 
             self._derive_variables_for_template_path(workspace)
+            self._define_object_template_vars()
             self._vars_are_expanded = True
 
     def _inputs_and_fetchers(self, workload=None):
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/application.py
```diff
@@ -1376,7 +1379,9 @@ def _make_experiments(self, workspace, app_inst=None):
                     f.write(
                         self.expander.expand_var(template_conf["contents"], extra_vars=exec_vars)
                     )
-                os.chmod(expand_path, stat.S_IRWXU | stat.S_IRWXG | stat.S_IROTH | stat.S_IXOTH)
+                os.chmod(expand_path, _DEFAULT_CONTENT_PERM)
+
+            self._render_object_templates(exec_vars)
```
douglasjacobsen, line 1384: This seems a bit out of place for `_inputs_and_fetchers`. Should we update the name of this method? Also, can we add some information to the docstring that this will render object defined templates as well?

lib/ramble/ramble/test/end_to_end/test_template.py
```diff
@@ -0,0 +1,57 @@
...
+        f.write(test_config)
+    ws._re_read()
+
+    workspace("setup", "--dry-run", global_args=["-w", workspace_name])
+    run_dir = os.path.join(ws.experiment_dir, "template/test_template/test/")
+    script_path = os.path.join(run_dir, "bar.sh")
+    assert os.path.isfile(script_path)
+    with open(script_path) as f:
+        content = f.read()
+        assert "echo hello santa" in content
```
douglasjacobsen, line 53: 🤣

douglasjacobsen (changes requested):
Looks great! One request but otherwise I think this is awesome.

lib/ramble/ramble/application.py
linsword13: Hm this might be a weird display by Github. This call is actually inside the `_make_experiments` function, not in the `_inputs_and_fetchers` one.

douglasjacobsen: Ah OK, thanks! Yeah, it was just a weird display thing... this time it showed something different so seems good to me. :)

douglasjacobsen (approved):

PR #814, opened 2025-01-08
Add the ability to define template-specific extra variables

linsword13:
The ability to generate dynamic variables is often needed when defining more complicated templates (for instance, programmatically generate slurm job headers.) Having this template-specific option to generate extra rendering vars reduces the need to ingest (more global-scoped) variables via `define_variables` via phase hooks.

### DIFF
--- a/lib/ramble/ramble/language/shared_language.py
+++ b/lib/ramble/ramble/language/shared_language.py
@@ -486,6 +486,7 @@ def register_template(
     src_name: str,
     dest_name: str,
     define_var: bool = True,
+    extra_vars: Optional[dict] = None,
     extra_vars_func: Optional[str] = None,
     output_perm=None,
 ):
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/language/shared_language.py
```diff
@@ -480,7 +482,12 @@ def _execute_target_shells(obj):
 
 @shared_directive("templates")
 def register_template(
-    name: str, src_name: str, dest_name: str, define_var: bool = True, output_perm=None
+    name: str,
+    src_name: str,
+    dest_name: str,
+    define_var: bool = True,
+    extra_vars_func: Optional[str] = None,
```
douglasjacobsen, line 490: Could we also support a:

`extra_vars: Optiona[dict] = {}`?

Or does this not work at this level?

linsword13: Good call, added.

douglasjacobsen (approved):

PR #826, opened 2025-01-15
Allow templates registered into other workspace paths

douglasjacobsen:
This merge allows workspace path variables to be replaced in destination paths for templates that are registered in objects.

This allows a template to be rendered into the shared directory of a workspace, as an example.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -48,6 +48,7 @@ import ramble.util.directives
 import ramble.util.stats
 import ramble.util.graph
 import ramble.util.class_attributes
+import ramble.util.path
 import ramble.util.lock as lk
 from ramble.util.foms import FomType
 from ramble.util.logger import logger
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/application.py
```diff
@@ -2311,7 +2312,7 @@ def _get_template_config(obj, tpl_config, obj_type):
             for tpl_conf in obj.templates.values():
                 yield _get_template_config(obj, tpl_conf, obj_type=obj_type)
 
-    def _render_object_templates(self, extra_vars):
+    def _render_object_templates(self, extra_vars, replacement_vars={}):
```
linsword13: nit: I recommend avoiding using mutable object as the default value: https://google.github.io/styleguide/pyguide.html#2124-decision.

linsword13 (changes requested):
Cool stuff! One nit request, but otherwise LGTM.

linsword13 (approved):

PR #830, opened 2025-01-16
Add a `batch_wait` script

linsword13:
The script polls (with 10s-interval) for the job-id, until it does not show up in squeue anymore.

### DIFF
--- a/var/ramble/repos/builtin/workflow_managers/slurm/batch_wait.tpl
+++ b/var/ramble/repos/builtin/workflow_managers/slurm/batch_wait.tpl
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# Ensure job_id is present
+job_id=$(< {experiment_run_dir}/.slurm_job)
+if [ -z "${job_id:-}" ]; then
+    echo "No valid job_id found" 1>&2
+    exit 1
+fi
+
+echo "Waiting for job {job_name} with id ${job_id} to complete..."
+
+while true; do
+    status=$(squeue -h -o "%t" -j "${job_id}" 2>/dev/null)
+    # The absence of the job from squeue indicates it's completed.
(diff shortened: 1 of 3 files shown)


var/ramble/repos/builtin/workflow_managers/slurm/batch_wait.tpl
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Ensure job_id is present
+job_id=$(< {experiment_run_dir}/.slurm_job)
+if [ -z "${job_id:-}" ]; then
+    echo "No valid job_id found" 1>&2
```
rfbgo, line 6: I think a lot of users will want something with a semantic like "wait for all jobs created by this workspace". Is it easy to extend to support that? Is the expected CUJ to `ramble on --executor {batch_wait}` and have it cycle through all one by one?

linsword13: Yeah it already does that via the existing `ramble on` mechanism. If one uses `ramble on --executor "{batch_wait}"`, then it will run this script against all experiments (and `--where` can be used to pick the experiments one wants to wait for.) Although it does that in a "serial" fashion. Example execution:

```
==> Running executors...
Waiting for job hostname_parallel_test.1 with id 382...
job hostname_parallel_test.1 with id 382 is done
Waiting for job hostname_parallel_test.2 with id 383...
job hostname_parallel_test.2 with id 383 is done
```

douglasjacobsen: I imagine that the CUJ for the time being is what @rfbgo said. And eventually we can have something like `ramble workflow wait` instead to call that.

douglasjacobsen (approved):

PR #833, opened 2025-01-17
Add in a hook for supplying variables for template render

linsword13:
This is used as the basis to allow workflow manager to provide a set of variables for all its `register_template` directives.


douglasjacobsen:
How should we think about this if we want different templates to have different variables that they render with?

linsword13:
> How should we think about this if we want different templates to have different variables that they render with?

Yeah for that we still have the `extra_vars` and `extra_vars_func` arguments via the `register_template`. And those have higher precedence over this `template_render_vars`.

douglasjacobsen:
OK, thanks!

douglasjacobsen (approved):

PR #834, opened 2025-01-18
Make slurm execute script customizable

linsword13:
This is done by making the template source path pointing to a
workflow_manager_variable.

The major change is around extending `register_template`'s `src_path`
prop (renamed from `src_name`) to allow it to point to a user-specified
path.

Another change related to template rendering is to define `template_render_vars` as reserved variables.


var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -83,35 +83,44 @@ def __init__(self, file_path):
         description="partition to submit job to, if unspecified, it uses the default partition",
     )
 
+    workflow_manager_variable(
+        name="slurm_execute_template_path",
+        default="slurm_execute_experiment.tpl",
+        description="Path to the custom template (under workspace_config dir) for generating "
```
douglasjacobsen: I think `under workspace_config dir` is a bit misleading here. The default is that it's under the configs dir, but allowing other workspace path variables (or an absolute path) this file can be anywhere on a file system, right?

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -83,35 +83,44 @@ def __init__(self, file_path):
...
+        src_path="batch_wait.tpl",
         dest_path="batch_wait",
     )
 
     register_template(
         name="slurm_execute_experiment",
-        src_name="slurm_execute_experiment.tpl",
-        dest_path="slurm_execute_experiment",
+        src_path="{slurm_execute_template_path}",
+        dest_path="execute_experiment",
```
douglasjacobsen: I am not sure how I feel about this hijacking `execute_experiment`. But I do understand why you did it (partly because I complained about how it felt confusing).

It's also kind of weird to have a template file not end up rendering as a template (I'll send you more about this in work chat).

douglasjacobsen (changes requested):

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Indeed, the description was wrong. Updated.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Ack. Reverted back to the `slurm_execute_experiment`.

douglasjacobsen (approved):

PR #836, opened 2025-01-21
Clean up NVIDIA HPC Benchmark application definitions

douglasjacobsen:
A large portion of the logic to allow exectuion of the NVIDIA HPC Benchmarks using the NVIDIA HPC Benchmark container is common regardless of which benchmark is being executed.

This merge extracts this logic into a common base_application definition, and applies it to the application definitions for each benchmark.

### DIFF
--- a/var/ramble/repos/builtin/base_applications/nvidia-hpc-benchmarks/base_application.py
+++ b/var/ramble/repos/builtin/base_applications/nvidia-hpc-benchmarks/base_application.py
@@ -1,0 +1,127 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+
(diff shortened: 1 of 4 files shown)


var/ramble/repos/builtin/base_applications/nvidia-hpc-benchmarks/base_application.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+
+    workload_variable(
+        "hpl_use_nvshmem",
+        default="0",
+        description="Whether to use NVSHMEM or not",
+        workload_group="all_workloads",
+    )
+    environment_variable(
+        "HPL_USE_NVSHMEM",
```
linsword13: I am not sure about having these "HPL-looking" stuff in this base app, but I also understand it'd be annoying to repeat them in the two HPL derivatives. Would it be too involved an inheritance chain, to have a base app `nvidia-hpl` that inherits from this, and that gets inherited by both the mxp and non-mxp apps?

douglasjacobsen: Thanks this should be fixed now. Reading the docs again, it's unclear if these env-vars only apply to HPL or if they also apply to HPL-MxP. So, I only put them on the HPL workload for the time being.

linsword13 (approved):

PR #865, opened 2025-02-04
Remove deprecated code for release v0.6.0

dapomeroy:
Update version and remove deprecated code for release v0.6.0


lib/ramble/ramble/test/end_to_end/known_applications.py
linsword13, line 1: Not sure if this is something we want to remove. I also don't quite know what the deprecation annotation refers to.

dapomeroy: The deprecation tag was added here: https://github.com/GoogleCloudPlatform/ramble/commit/0c5219e108850d54ef8e1e4b6d3107d0a4c25dba
I had thought that known_applications was rendered obsolete when Doug reworked testing in cloud build, but perhaps this tag is referring to `if package_manager == "None":` on L53?

lib/ramble/ramble/test/application_language.py
```diff
@@ -19,22 +18,6 @@
 ]
 
 
-@deprecation.fail_if_not_removed
-@pytest.mark.parametrize("app_class", app_types)
-def test_application_type_features(app_class):
```
douglasjacobsen: I don't think this should be removed. I think we just remove the deprecation tag on this, if the `SpackApplication` is still here.

lib/ramble/ramble/test/end_to_end/known_applications.py
douglasjacobsen: @linsword13 is right, we do not want this removed either. This is one of our most important tests.

lib/ramble/ramble/test/modifier_language.py
```diff
@@ -17,28 +16,6 @@
 mod_types = [ModifierBase, BasicModifier]  # noqa: F405
 
 
-@deprecation.fail_if_not_removed
```
douglasjacobsen: Same comment here about not removing this test.

douglasjacobsen (changes requested):
Also, if you can merge the newest develop in, that would be great.

lib/ramble/ramble/test/application_language.py
dapomeroy: Updated to just remove the tag

lib/ramble/ramble/test/modifier_language.py
dapomeroy: Updated to just remove the tag

lib/ramble/ramble/test/end_to_end/known_applications.py
dapomeroy: Fixed this, restored the test and removed the tag.

dapomeroy:
This is updated, tests are passing, and I've rebased on develop. Please review the deletions again and let me know if you spot any other issues

douglasjacobsen (approved):

PR #869, opened 2025-02-06
feat: update lammps reaxff to use newer builds

vsoch:
Problem: we want to run apptainer with lammps, and not require any spack or previous software installed. This adds an example, and also a new variable that exposes the version of lammps to download.

@douglasjacobsen this successfully runs our lammps container with ramble, with the updated problem (and I figured out what I needed to in order to get it to work, thanks for the help!)

I added a new directory to examples that can explicitly be for apptainer, the reason because I think some users will want to skip spack, etc. and jump straight to that. 

## Questions

I have questions about usability too (and sorry if I missed these in the docs).

- I noticed on the first run I did that the input.txt file, even when I specified a new input_file and ran setup again, did not update. I had to manually remove the experiment directory and then setup again and rerun for the change to propagate. **Could that be a bug or is there a different command I need to run in order to properly cleanup?**
- I'd like to be able to do `ramble on` and see the output stream to the terminal (right now it doesn't, it runs silently and I have to cat the output file) - **is there a flag or command to stream output to the terminal**?
- **Is there a programmatic way to create the workspace from the file** instead of manually copy pasting into the `ramble workspace edit --config`? We will need that for automated runs.
- **Where do I specify a number of iterations for an experiment?**
- **Is there a command to cleanup a ramble workspace (not delete it, just remove the experiment results to do again)?**
- For top level variables, I'm wondering why `mpi_command` and `batch_submit` need to be set, when they are basically using defaults (an empty string and `{execute_command}`. It added some confusion for my first try because I had no idea what they means.
- I'm guessing that for some run with a workload manager, I'd set the `batch_submit` to something with my workload manager (e.g., flux)? And would I put the nodes / tasks in there as well?

Thanks for answering my questions! When I know about the above, I'll probably next test this running from inside a container with a lammps executable directly, and then with flux.

### DIFF
--- a/examples/apptainer/lammps-reaxff.yaml
+++ b/examples/apptainer/lammps-reaxff.yaml
@@ -1,0 +1,51 @@
+# Copyright 2022-2025 The Ramble Authors
+
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+ramble:
+  modifiers:
+  - name: apptainer
+  env_vars:
+    set:
+      OMP_NUM_THREADS: '{n_threads}'
+  variables:
+    mpi_command: ''
+    container_uri: docker://ghcr.io/converged-computing/metric-lammps-cpu:libfabric-zen4-reax
+    container_name: lammps-reax
+    batch_submit: '{execute_experiment}'
+    processes_per_node: 1
+    gpus_per_node: 0
+    # This is how you can add flags
+    # apptainer_run_args: --nv --bind {container_mounts} --writable-tmpfs
+  applications:
+    lammps: # Application name
+      workloads:
+        hns-reaxff: # Workload name from application
+          experiments:
+            strong_scale: # Arbitrary experiment name
+              variables:
+                env_name: apptainer
+                n_nodes: 1
+                lammps_path: /usr
+                # Add other lammps flags
+                lammps_flags: "-nocite"
+                # This is the release that is downloaded
+                input_stage: stable_29Aug2024_update1
+                # The input file used for lammps from the input_path directory
+                input_file: in.reaxff.hns
+                # Problem size dimensions for lammps
+                xx: 2
+                yy: 2
+                zz: 2
+  software:
+    packages:
+      apptainer:
+        pkg_spec: apptainer
+    environments:
+      lammps:
+        packages:
+        - apptainer
--- a/var/ramble/repos/builtin/applications/lammps/application.py
+++ b/var/ramble/repos/builtin/applications/lammps/application.py
@@ -72,12 +72,10 @@ class Lammps(ExecutableApplication):
 
     input_file(
         "lammps-stage",
-        url="https://github.com/lammps/lammps/archive/refs/tags/stable_23Jun2022_update3.tar.gz",
-        target_dir="{application_input_dir}/stable_23Jun2022_update3",
-        description="Stage of lammps source from 20220623.3 release",
-        sha256="8a276a01b50d37eecfe6eb36f420f354cde51936d20aca7944dea60d3c098c89",
+        url="https://github.com/lammps/lammps/archive/refs/tags/{input_stage}.tar.gz",
+        target_dir="{application_input_dir}/lammps-input-stage",
+        description="Stage of lammps source from release",
     )
-
     executable(
         "copy",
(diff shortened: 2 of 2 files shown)


vsoch:
I'm also wondering if we can support a generic application executor. For example, I don't need to hard code every specific argument for some application, I could just define a container, an input line with parameters, and then use ramble to programatically (and easily) run a lot of experiments without having to write an application.py for each. It's a more flexible approach that is less prone to not having, for example, an updated download of data, and it would allow for using ramble to reproduce experiments without needing to write a lot of those complex files. What do you think?

douglasjacobsen:
Hey @vsoch here are some hopefully helpful responses...

* It looks to me like the `input.txt` file is created at run-time, rather than setup time. https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/applications/lammps/application.py#L81-L99 So, I would expect the input to change if you: Updated the config, ran `ramble workspace setup` and then executed the experiment.
* There are a few different solutions for this. It ultimately depends on what  you want to have happened *after* the run completes, but I might suggest using the `gcp-cloud-logging` modifier. Outside of that, maybe look at using the `{unformatted_command_sans_logs}` variable when creating execution templates. The only downside with that, is Ramble won't help you extract figures of merit if you use that.
* Yes! Look at `ramble workspace manage experiments`. We need to write a similar command for managing modifiers, but haven't gotten around to that yet. You can also look at using `ramble deployment` or `ramble workspace archive -t` and then moving the tar around between the nodes. You can also put your configs in a repo, and then use `ramble workspace manage includes` to pull them in.
* What do you mean by iterations? Do you mean the number of iterations in the lammps experiment? Or how many repeat experiments you want to run to get statistics from? If you mean lammps, some workloads expose a timesteps variable (https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/applications/lammps/application.py#L162-L167). If you mean repeats, https://ramble.readthedocs.io/en/latest/configuration_files.html#experiment-repeats
* No. We've discussed this several times, but the current recommendation is `rm -rf experiments; ramble workspace setup`
* This is a great question, and given the workflow managers we added recently we will probably revisit these definitions. Previously, they were required things to be defined, and we wrote them in the default config mostly as a way to communicate to users things that you might want to edit. Given users have different workflows and MPI implementations they want to use, these change pretty frequently from one system to another.
* Yes, but also, we have workflow managers that you can define (we don't have docs for this yet) in the `variants:workflow_manager:slurm` block. Slurm is currently the only option. This would fill out most of the things for you automatically. For flux, you would update the `batch_submit` definition to be the submission command. Referring to `{execute_experiment}` gives the path to the rendered execute experiment script (which could be a flux submission script).

Regarding your question about generic application executors.

There are a few reasons you would not want to do this, just FYI. You end up missing out on a lot of the things Ramble can do for you (like unifying the outputs, archiving files from the experiment, and things like that).

However, you can turn any of the existing application definitions into a generic experiment. I use hostname for this frequently. With that, you can define internals that are your own custom execution steps: https://ramble.readthedocs.io/en/latest/configuration_files.html#internals-config

You can also see this tutorial for more information on how this might work: https://ramble.readthedocs.io/en/latest/tutorials/11_using_internals.html

Again though, don't expect `ramble workspace analyze` to give you much useful output from this.

lib/ramble/ramble/pipeline.py
```diff
@@ -572,6 +572,7 @@ def _execute(self):
             app_inst.add_expand_vars(self.workspace)
             exec_str = app_inst.expander.expand_var(self.executor)
             exec_parts = shlex.split(exec_str)
+
```
douglasjacobsen: nit; Can you remove this change?

douglasjacobsen (changes requested):
One small change, but otherwise it looks good. Thanks!

lib/ramble/ramble/pipeline.py
vsoch: oops, sorry about that, leftover from an interactive debugging change! Removed.

douglasjacobsen (dismissed):

douglasjacobsen:
Looks like there are some style issues.

You can try running `ramble style -f` to fix them. But they are:
1) the license header seems to be wrong, probably on the YAML file you added.
2)
```
Step #1 - "ramble-style-tests": var/ramble/repos/builtin/applications/lammps/application.py:278: [E131] continuation line unaligned for hanging indent
Step #1 - "ramble-style-tests": var/ramble/repos/builtin/applications/lammps/application.py:279: [E131] continuation line unaligned for hanging indent
Step #1 - "ramble-style-tests": var/ramble/repos/builtin/applications/lammps/application.py:280: [E131] continuation line unaligned for hanging indent
```

douglasjacobsen (approved):

PR #876, opened 2025-02-13
Add a stub package manager that does nothing

douglasjacobsen:
This merge adds a stub package manager that performs no actions, as a way to work around required paths.

### DIFF
--- a/var/ramble/repos/builtin/package_managers/stub/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/stub/package_manager.py
@@ -1,0 +1,22 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.pkgmankit import *  # noqa: F403
+
+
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/package_managers/stub/package_manager.py
```diff
@@ -0,0 +1,22 @@
...
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.pkgmankit import *  # noqa: F403
+
+
+class Stub(PackageManagerBase):
```
rfbgo, line 12: We chatted a bit offline about the naming here. After some more though I think I would personally prefer something like:

- existing `None` is rename to something like `userManaged` 
- To me this is closer to `None` or `doNothing` 

Also I think we should start emitting a package manager in the default yaml 

Happy to chat more, but figured concrete suggestions might be useful. Overall happy to merge the concept (but I don't personally love the `stub` name the more I think about it)

rfbgo (changes requested):

var/ramble/repos/builtin/package_managers/stub/package_manager.py
douglasjacobsen: Thanks, I'm going to close this in favor of #882

douglasjacobsen:
Closing in favor of #882

PR #882, opened 2025-02-21
Create a user-managed package manager, update None package manager

douglasjacobsen:
This merge creates a user-managed package manager that behaves as the previous None package manager was.

Additionally, the None package manager has been converted to not do anything. Paths are no longer required to be defined when using a None package manager anymore. The default package manager is updated to be `user-managed`.

### DIFF
--- a/var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
@@ -1,0 +1,56 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.pkgmankit import *  # noqa: F403
+
+
+class UserManaged(PackageManagerBase):
+    """Package manager representing a user managed environment.
+
+    This package manager is used when the software required for the experiments
+    is manually installed outside of Ramble. Generally, a user will need to
+    convey the paths to these installed packages to Ramble through specific
+    variable definitions.
+    """
+
+    name = "user-managed"
+
+    _spec_prefix = "user_managed"
+
+    register_phase(
+        "define_requirements",
+        pipeline="setup",
+        run_before=["get_inputs"],
+    )
+
+    def _define_requirements(self, workspace, app_inst=None):
+        """Define requirements for user managed software stack
+
+        Extracts all required packages from experiments and modifiers, then
+        creates required variables to convey the installation locations to
+        Ramble.
+        """
+
+        package_objects = [app_inst]
+        package_objects.append(self)
+        if app_inst.workflow_manager is not None:
+            package_objects.append(app_inst.workflow_manager)
+        package_objects.extend(app_inst._modifier_instances)
+
+        for obj in package_objects:
+            for pkgname, _ in obj.required_packages.items():
+                app_inst.keywords.update_keys(
+                    {
+                        f"{pkgname}_path": {
+                            "type": ramble.keywords.key_type.required,
+                            "level": ramble.keywords.output_level.variable,
+                        }
+                    }
+                )
+
+        app_inst.validate_experiment()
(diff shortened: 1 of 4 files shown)


var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -0,0 +1,56 @@
...
+    def _define_requirements(self, workspace, app_inst=None):
+        """Define requirements for user managed software stack
+
+        Extracts all required packages from experiments and modifiers, then
+        creates required variables to convey the installation locations to
+        Ramble.
+        """
+
+        package_objects = [app_inst]
+        package_objects.append(self)
```
linsword13: nit: maybe just `package_objects = [app_inst, self]`?

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -0,0 +1,56 @@
...
+        if app_inst.workflow_manager is not None:
```
linsword13: Do we need to safeguard when `app_inst` is `None`?

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -0,0 +1,56 @@
...
+            package_objects.append(app_inst.workflow_manager)
+        package_objects.extend(app_inst._modifier_instances)
```
linsword13: Maybe for later, but it may be nice to use for instance `app_inst._objects`  to access these objects, so that this package manager doesn't need to have direct knowledge of how to enumerate the various objects.

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -0,0 +1,56 @@
...
+
+        for obj in package_objects:
+            for pkgname, _ in obj.required_packages.items():
```
linsword13: nit: maybe use `keys()`?

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
douglasjacobsen: `app_inst` should never be `None` in practice (i.e. if it is None, then we don't have a valid experiment). I could add a guard though, just to help with unit testing. Really the only reason we default `app_inst=None` in the interface was that we added the `app_inst` argument after we had defined the `register_phase` directive. So this was more of a "try to be backwards compatible", but it's also been long enough that that doesn't worry much much anymore.

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
douglasjacobsen: Ah that is a good idea. I can update that.

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
linsword13: I see, thanks. I think this sounds good then.

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -0,0 +1,55 @@
...
+
+        Extracts all required packages from experiments and modifiers, then
+        creates required variables to convey the installation locations to
+        Ramble.
+        """
+
+        if app_inst is None:
+            package_objects = [self]
+        else:
+            package_objects = app_inst._objects
```
linsword13: The `_objects` is a callable, and it returns a list of tuples containing both the type and the actual object.

douglasjacobsen: I should have looked. 🤣 Thanks

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -0,0 +1,55 @@
...
+
+        Extracts all required packages from experiments and modifiers, then
+        creates required variables to convey the installation locations to
+        Ramble.
+        """
+
+        if app_inst is None:
+            package_objects = [self]
+        else:
+            package_objects = app_inst._objects()
```
linsword13, line 42: I think you probably want only the second part of the tuple

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -0,0 +1,55 @@
...
+
+        for _, obj in package_objects:
```
linsword13, line 44: I guess technically this would be an issue for the app_inst == None branch, although as you mentioned we probably never hit that. But maybe just make a fake tuple like `(None, self)`?

douglasjacobsen: Ah yeah, that's a good point.

linsword13 (approved):

PR #885, opened 2025-02-24
Update default semantics for variants in new workspaces

douglasjacobsen:
This merge updates the semantics around variants within newly created workspaces. Previously, externally defined variants would not show up within a workspace. This merge updates this behavior so now any variants that are defined outside of workspaces are propagated into the workspace when it is created.

This is primarily as a mechanism to communicate to users who create new workspaces what is defined within their workspace and to help allow them to change this more easily.

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -576,7 +576,21 @@ cd "{experiment_run_dir}"
 
     @classmethod
     def _default_config_yaml(self):
-        return """\
+
+        # Construct string for default variants
+        variant_string = ""
+        variant_dict = ramble.config.get("variants")
+        variant_defs = []
+
+        for var, val in variant_dict.items():
+            variant_defs.append(f"    {var}: {val}")
+
+        if variant_defs:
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -594,20 +604,20 @@ def _default_config_yaml(self):
...
 
 ramble:
   env_vars:
     set:
-      OMP_NUM_THREADS: '{n_threads}'
-  variables:
-    mpi_command: mpirun -n {n_ranks}
-    batch_submit: '{execute_experiment}'
+      OMP_NUM_THREADS: '{{n_threads}}'
+{variant_string}  variables:
```
linsword13: Would it be cleaner to put "variables:" in a new line?

douglasjacobsen: Cleaned up. I was annoyed by it injecting a blank new line, and messing with the formatting. But I fixed it, so it's better now.

rfbgo (approved):

linsword13 (approved):

PR #887, opened 2025-02-25
Allow package managers to ignore defined software environments

douglasjacobsen:
This merge fixes an issue where the `user-managed` package required an environment to be used, when it doesn't actually consume it.

Package managers now have a `uses_software_environments` attribute that can be used to prevent them from requiring a software environment.


samskillman (approved):
Tested - Fixes things for me!

rfbgo (approved):

PR #906, opened 2025-03-04
Add a shell util for generating associative array

linsword13:


lib/ramble/ramble/util/shell_utils.py
```diff
@@ -69,5 +69,29 @@ def cmd_sub_str(shell, cmd: str):
...
+    Args:
+        var_name: the variable name for the generated dict in the shell.
+        dict: the input dict.
+        shell: the shell used. Currently this is bash-only.
+    Return:
+        The string of the generated shell commands.
+    """
+
+    if shell != "bash":
+        raise UnsupportedError("`define_dict` is only supported in bash")
```
douglasjacobsen, line 87: Do you think there's a path to supporting other shells with this?

I'm a bit worried that adding this is going to make other shells unable to use the workflow managers (though, TBF I don't know that that is any different than what we currently have...).

Would it be better to render a python script to determine which status we have instead of using shell arrays?

linsword13: I think this will only not work if the system does not have bash installed. Even if a user is using another shell (say `fish`), as long as their system has bash, then the workflow manager scripts (which specify bash via the shebang) should still work fine.

Using a python script sounds like a good idea. Although there may be some things to work through in terms of importing (if the workflow manager and the script both want to share some common things like the status_map)

douglasjacobsen: 👍🏻  OK, thanks.

douglasjacobsen (approved):

PR #907, opened 2025-03-05
Fix variants in chained experiments

douglasjacobsen:
This merge fixes two issues with chained experiments.

First, it updates the name of the chained experiments to avoid a warning about invalid decimal literals.

Second, it propagates variants into chained experiments, to make sure things like package managers and workflow managers are setup properly.

Tests are added for this as well.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -738,8 +738,8 @@ class ApplicationBase(metaclass=ApplicationMeta):
                 if "order" in cur_exp_def:
                     order = cur_exp_def["order"]
 
-                chained_name = f"{chain_idx}.{cur_exp_name}"
-                new_name = f"{parent_namespace}.chain.{chained_name}"
+                chained_name = f"chain.{chain_idx}.{cur_exp_name}"
+                new_name = f"{parent_namespace}.{chained_name}"
 
                 new_run_dir = os.path.join(
                     parent_run_dir, namespace.chained_experiments, chained_name
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -229,6 +229,8 @@ class ApplicationBase(metaclass=ApplicationMeta):
         new_copy.set_template(False)
         new_copy.repeats.set_repeats(False, 0)
         new_copy.set_chained_experiments(None)
+        if self.variants:
+            new_copy.set_variants(self.variants)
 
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/test/end_to_end/chained_experiment_var_inheritance.py
```diff
@@ -25,7 +25,9 @@
 
 
 @pytest.mark.filterwarnings("ignore:invalid decimal literal")
-def test_chained_experiment_variable_inheritance(mutable_config, mutable_mock_workspace_path):
+def test_chained_experiment_variable_inheritance(
+    mutable_config, mutable_mock_workspace_path, request
```
linsword13: nit: since `mutable_config` and `mutable_mock_workspace_path` are listed above with `usefixtures`, they don't need to be listed here explicitly? 

Also now that the invalid decimal issue is addressed, is the `filterwarnings` still needed?

lib/ramble/ramble/test/end_to_end/chained_experiment_var_inheritance.py
```diff
@@ -152,7 +154,7 @@ def test_chained_experiment_variable_inheritance(mutable_config, mutable_mock_wo
             chained_script = os.path.join(
                 parent_dir,
                 "chained_experiments",
-                f"{chain_idx}" + r".intel-mpi-benchmarks.collective.collective_chain",
+                rf"chain.{chain_idx}.intel-mpi-benchmarks.collective.collective_chain",
```
linsword13: Just curious, is the r prefix needed here (given that there isn't a backslash?)

lib/ramble/ramble/test/end_to_end/chained_experiment_var_inheritance.py
douglasjacobsen: Ah, good point. Thanks. :)

lib/ramble/ramble/test/end_to_end/chained_experiment_var_inheritance.py
douglasjacobsen: Yeah, thanks. I think we had the `r` before to try and work around the invalid decimal error, but it didn't do anything and ended up being left around.

Good thing to clean up. :)

linsword13 (approved):

PR #914, opened 2025-03-09
Print additional slurm job info from the query script

linsword13:
These info can be useful for various purpose, such as debugging.

Example output:

```
job foo with id 1934 has status: COMPLETE
additional job info:
  nodes:nodeset-[11,29]
  start:2025-03-09T06:45:22
  end:2025-03-09T06:55:40
```

Also add in a few job status mappings.

### DIFF
--- a/var/ramble/repos/builtin/workflow_managers/slurm/batch_query.tpl
+++ b/var/ramble/repos/builtin/workflow_managers/slurm/batch_query.tpl
@@ -17,11 +17,15 @@ if [ ! -z "$status" ]; then
         status=${status_map["$status"]}
     fi
 fi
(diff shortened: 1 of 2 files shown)


douglasjacobsen:
This is great. :)

Do you think we can add this to a file in the experiment directory too, in case the `.slurm_job` file is broken / removed?

linsword13:
> This is great. :)
> 
> Do you think we can add this to a file in the experiment directory too, in case the `.slurm_job` file is broken / removed?

Great idea! added.

As a future idea, it may be interesting to run this query as part of analyze, and then having workflow manager-specific FOMs for these info.

douglasjacobsen (approved):

PR #918, opened 2025-03-11
Add experiment diff prints to help debug non-unique experiment errors

douglasjacobsen:
This merge primarily adds some print statements when a workspace with non-unique experiments is acted on. These help users determine what the differences between the two non-unique experiments are, so they can fix the workspace configuration.

Additionally, a bug is fixed that overrides an explicit definition for n_threads=0 to be n_threads=1.

### DIFF
--- a/lib/ramble/ramble/experiment_set.py
+++ b/lib/ramble/ramble/experiment_set.py
@@ -251,7 +251,7 @@ class ExperimentSet:
         elif not n_nodes:
             variables[self.keywords.n_nodes] = 1
 
-        if n_threads is None:
+        if not n_threads:
             variables[self.keywords.n_threads] = 1
 
     def _prepare_experiment(self, exp_template_name, variables, context, repeats):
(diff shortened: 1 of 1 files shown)


lib/ramble/ramble/experiment_set.py
```diff
@@ -491,6 +491,47 @@ def _ingest_experiments(self):
...
+                    left_unique_vars = {}
+                    right_unique_vars = {}
+                    diff_vars = {}
+                    left_exp_vars = self.experiments[final_exp_namespace].variables
+                    right_exp_vars = experiment_vars
+
+                    logger.warn("Two experiments are defined with the name {final_exp_namespace}")
+
+                    # Determine variables that are only in one of the two experiments
+                    for var in left_exp_vars:
```
linsword13: nitpick: I wonder if some of these for loops can be simplified using set operations:

```
lkeys = set(left_exp_vars.keys())
rkeys = set(right_exp_vars.keys())

left_unique_vars = lkeys - rkeys
right_unique_vars = rkeys - lkeys
common_vars = lkeys & rkeys
...
```

linsword13 (dismissed):

lib/ramble/ramble/experiment_set.py
douglasjacobsen: Good suggestion. :)

LMK if you like that more.

linsword13: Looks great!

linsword13 (approved):

PR #928, opened 2025-03-17
Allow templates in nested directories inside configs

douglasjacobsen:
This merge enables templates within the configs directory to be in nested directories. The directories are then created within the resulting experiment directories, and the variables that contain their paths are named "{directory_name/template_name}".


lib/ramble/ramble/test/end_to_end/nested_config_templates.py
linsword13, line 1: Would it be useful to include testing for the variable name behavior (that `{templates/test_template}` renders to the path)?

douglasjacobsen: Good idea. Added!

linsword13 (approved):

PR #929, opened 2025-03-18
Allow `ramble workspace edit` to have unknown arguments

douglasjacobsen:
This merge allows Ramble to pass through additional argument to `ramble workspace edit`. This allows users to invoke the editor with additional arguments easily.

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -9,6 +9,7 @@
 import os
 import sys
 import tempfile
+import argparse
 
 import llnl.util.tty as tty
 import llnl.util.tty.color as color
@@ -855,7 +856,7 @@ def workspace_edit_setup_parser(subparser):
 
     subparser.add_argument(
         "-c",
-        "--config_only",
+        "--config-only",
         dest="config_only",
         action="store_true",
         help="Only open config files",
@@ -864,7 +865,7 @@ def workspace_edit_setup_parser(subparser):
 
     subparser.add_argument(
         "-t",
-        "--template_only",
+        "--template-only",
         dest="template_only",
         action="store_true",
         help="Only open template files",
@@ -873,7 +874,7 @@ def workspace_edit_setup_parser(subparser):
 
     subparser.add_argument(
         "-l",
-        "--license_only",
+        "--license-only",
         dest="license_only",
         action="store_true",
         help="Only open license config files",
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -892,8 +903,14 @@ def workspace_edit_setup_parser(subparser):
         "-p", "--print-file", action="store_true", help="print the file name that would be edited"
     )
 
+    subparser.add_argument(
```
linsword13, line 896: Is this used?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1473,19 +1501,20 @@ def setup_parser(subparser):
         setup_parser_cmd(subsubparser)
 
 
-def workspace(parser, args):
+def workspace(parser, args, unknown_args):
     """Look for a function called workspace_<name> and call it."""
     action = subcommand_functions[args.workspace_command]
-    action(args)
+    action(args, unknown_args)
```
linsword13: Just thinking aloud if it's possible to somehow make the decision here. One somewhat hacky approach could be to inspect the arguments expected by the subcommand, and then only delegate the decision to subcommands that expect the `unknown_args` argument. It's not super kosher, but has the benefit of leaving other subcommands except `workspace_edit` unchanged.

```
def workspace(parser, args, unknown_args):
    """Look for a function called workspace_<name> and call it."""
    action = subcommand_functions[args.workspace_command]
    arg_len = len(inspect.signature(action).parameters)
    if arg_len <= 1:
        # When the subcommand accepts a single arg, assume it should throw upon unknown args
        if unknown_args:
            logger.die("...")
        else:
            action(args)
    else:
        # Otherwise, leave the handling of unknown_args to the individual subcommand
        action(args, unknown_args)
```

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: No. Mostly because there's no way to actually populate it. But this is intended as more of documentation so that users will see they can pass in extra args with `ramble workspace edit -h`

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Thanks, that's a good idea.

I updated this to try and unify checking subcommands for unknown_args. Let me know if you like this more.

linsword13 (approved):

PR #947, opened 2025-04-03
Add a workflow manager to support running hpc apps in k8s cluster

linsword13:
It uses:

* `mpi-operator` from `kubeflow` to manage the worker pods
* `kustomize` to port locally generated assets into the pods

Example workflow for running HPL:

* Ramble workspace config: (the `container_image` points to an image containing HPL dependency, courtesy of @akiki-liang0)

```yaml
ramble:
  variants:
    package_manager: None
    workflow_manager: gke-mpi
  applications:
    hpl:
      workloads:
        calculator:
          experiments:
            gketest:
              env_vars:
                set:
                  OMP_NUM_THREADS: '{n_threads}'
                  OMPI_ALLOW_RUN_AS_ROOT: 1
                  OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
                  LD_LIBRARY_PATH: /usr/local/lib/openmpi:/openBLAS:/root/hpl:/root/hpl/bin:/root/hpl/bin/linux:/usr/local/lib
                  OMP_PLACES: 'core'
                  OMP_PROC_BIND: 'true'
              variables:
                mpi_command: mpirun -np {n_ranks} --npernode {processes_per_node} -v --allow-run-as-root --bind-to core -x PATH -x LD_LIBRARY_PATH --mca btl tcp,self
                processes_per_node: 192
                n_nodes: 2
                memory_per_node: 762
                array_size: 6400000000
                N-NBMINs: 1
                NBMINs: 4
                NPFACTs: 1
                PFACTs: 1
                N-RFACTs: 1
                RFACTs: 1
                extra_container_config_files: |
                  '{experiment_run_dir}/HPL.dat'
                container_image: '<container_image_uri>'
                container_work_dir: '/root/hpl/bin/linux'
```

* Run `ramble workspace setup` to set up the `kustomize` templates

* Optionally, can run `ramble on --executor '{batch_print_deployment}'` to inspect the generated deployment template

* Run `ramble on` to submit the job to an existing gke cluster

* During the run, can use `ramble on --executor '{batch_query}'` to look at the job info and get the launcher log

* Run `ramble workspace analyze` upon completion, this fetches in the launcher log (the hpl isn't optimized, so ignore the perf result):

```
$ ramble workspace analyze -p
From Workspace: gke-hpl (hash: 1f96b0b9cf8fe9b0c67e29740404175ae559bae9ec75757150e2a715eec2e0ae)
Experiment hpl.calculator.gketest figures of merit:
  Status = SUCCESS
  Tags = ['benchmark', 'benchmark-app', 'linpack']
  N-NB-P-Q = 405504-384-16-24 context figures of merit:
    Time = 2832.29 s
    GFlops = 1.5695e+04 GFLOP/s
```

* At the end, run `ramble on --executor '{batch_cancel}'` to delete the launcher pod. This is intentionally configured to stay running after the job completion, so that its logs can be fetched. That means for now it needs to be manually cleaned up after the run (i.e., running this cancel command).

### DIFF
--- a/var/ramble/repos/builtin/workflow_managers/gke-mpi/workflow_manager.py
+++ b/var/ramble/repos/builtin/workflow_managers/gke-mpi/workflow_manager.py
@@ -1,0 +1,186 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import os
+import shutil
+import textwrap
+
+from ramble.wmkit import *
+
+from spack.util.executable import Executable, ProcessError
+
+
+class GkeMpi(WorkflowManagerBase):
+    """GKE workflow manager that uses the MPI operator"""
+
+    name = "gke-mpi"
+
+    maintainers("linsword13")
+
+    tags("workflow", "gke", "mpi")
+
+    workflow_manager_variable(
+        name="job_name",
+        default="{application_name}-{workload_name}-{experiment_name}",
+        description="GKE job name",
+    )
+
+    workflow_manager_variable(
+        name="extra_metadata",
+        default="",
+        description="Extra line-separated key:val pairs for the metadata section",
+    )
+
+    workflow_manager_variable(
+        name="cores_per_node",
+        default="{processes_per_node}",
+        description="Cores per node",
+    )
+
+    workflow_manager_variable(
+        name="container_image",
+        default="",
+        description="url to the container image",
+    )
+
+    workflow_manager_variable(
+        name="extra_container_config_files",
+        default="",
+        description="extra line-separated list of config files to be mapped to containers",
+    )
+
+    workflow_manager_variable(
+        name="container_work_dir",
+        default="/config",
+        description="working directory inside the container",
+    )
+
+    workflow_manager_variable(
+        name="launcher_execute_script_template",
+        default="launcher_execute_script.tpl",
+        description="execute script template for the launcher",
+    )
+
+    register_template(
+        name="launcher_execute_script",
+        src_path="{launcher_execute_script_template}",
+    )
+
+    workflow_manager_variable(
+        name="worker_execute_script_template",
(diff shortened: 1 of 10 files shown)


var/ramble/repos/builtin/workflow_managers/gke-mpi/workflow_manager.py
```diff
@@ -0,0 +1,184 @@
...
+    )
+
+    workflow_manager_variable(
+        name="container_image",
+        default="",
+        description="url to the container image",
+    )
+
+    workflow_manager_variable(
+        name="extra_config_files",
```
douglasjacobsen: nit; We should probably discuss this, but maybe this would be better if it were clear the config files were used for k8s?

var/ramble/repos/builtin/workflow_managers/gke-mpi/workflow_manager.py
```diff
@@ -0,0 +1,184 @@
...
+        default="",
+        description="extra line-separated list of config files to be mapped to containers",
+    )
+
+    workflow_manager_variable(
+        name="gke_run_dir",
```
douglasjacobsen: nit; Annoyingly, this might be better to be a bit more general... like "container_work_dir" or something. I think there is something similar in slurm-pyxis based workflows, and it might be nice to have this generalized between them.

var/ramble/repos/builtin/workflow_managers/gke-mpi/workflow_manager.py
douglasjacobsen: Or maybe it would just be good to call out that it's the workflow config files? like `extra_workflow_config_files` ?

douglasjacobsen (changes requested):
This is awesome. Thanks for working on this.

I have two small nits, which we can discuss in our meeting later today.

var/ramble/repos/builtin/workflow_managers/gke-mpi/workflow_manager.py
linsword13: Good call, I now rename it to `extra_container_config_files`.

var/ramble/repos/builtin/workflow_managers/gke-mpi/workflow_manager.py
linsword13: SG, updated to `container_work_dir`.

douglasjacobsen (approved):

PR #948, opened 2025-04-04
Tolerate some "none-looking" names in variant setting

linsword13:
With the change, all the following would point to the None pkgman:

```
package_manager: null
package_manager: None
package_manager: ''
package_manager: 'None'
package_manager: 'none'
```

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -51,6 +51,7 @@ from ramble.error import RambleError
 from ramble.experiment_result import ExperimentResult
 from ramble.language.application_language import ApplicationMeta
 from ramble.language.shared_language import SharedMeta, register_builtin, register_phase
+from ramble.util import conversions
 from ramble.util.foms import FomType
 from ramble.util.logger import logger
 from ramble.util.naming import NS_SEPARATOR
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/application.py
```diff
@@ -130,6 +130,15 @@ def _get_phase_func_wrapper(workspace, phase_func, phase_name):
     return profiler(phase_func)
 
 
+def _canonical_none(maybe_none):
```
douglasjacobsen: Can we move this to a utility, so we can use it in other places? It might be nice to be able to use it in the expander so we can expand to None types too.

douglasjacobsen (changes requested):

lib/ramble/ramble/application.py
linsword13: SG, moved to a util. I initially thought of building it directly into the expander, but I guess converting empty string to None is not a generally good idea (even when `typed=True`)

douglasjacobsen: Yeah, that's a good point. Thanks!

douglasjacobsen (approved):

PR #949, opened 2025-04-04
Fix documentation issues

dapomeroy:
Fix various documentation issues like broken formatting and links.


lib/ramble/docs/tutorials/shared/gromacs_vector_workspace.rst
```diff
@@ -65,7 +65,7 @@ contents of ``$RAMBLE_ROOT/examples/vector_matrix_gromacs_config.yaml``:
 requires ``spack`` is installed and available in your path. Modifications to
 the ``package_manager`` variant will change this behavior.
 
-.. literalinclude:: ../../../../examples/vector_matrix_gromacs_config.yaml
+.. literalinclude:: /../../../examples/vector_matrix_gromacs_config.yaml
```
dapomeroy, line 68: There's some kind of a bug in Sphinx where the relative path is being calculated wrong for the literalinclude yaml file inside of an included rst. If it's set to the 4th parent it resolves to ramble_parent/ramble/lib/examples and if it's set to the 5th parent it resolves to ramble_parent/examples. Very weird! So this switches it to being based off the sphinx root dir (where conf.py is located) and it just works.

Some of the other alternatives I looked at were more complicated, like trying to do substitutions of absolute paths, using rst_epilog or environment variables, etc. This one is simple and effective if perhaps a bit unintuitive. We'll need to do this for any of the files in the shared dir that have a literalinclude.

dapomeroy:
I set up a dev build of readthedocs for my fork, and the updated docs in this PR are available to view there: https://ramble-dev-dapomeroy.readthedocs.io/en/update_documentation/

lib/ramble/docs/workspace_config.rst
```diff
@@ -187,7 +186,6 @@ Dictionary references are supported:
 
 .. _ramble-escaped-variables:
 
-~~~~~~~~~~~~~~~~~
```
linsword13, line 190: Curious why this is removed?

lib/ramble/ramble/context.py
```diff
@@ -111,12 +111,13 @@ def create_context_from_dict(context_name, in_dict):
         'exclude': {},
         'zips': {},
         'matrices': {} or [],
-        'tags': []
-        'n_repeats': '',
+        'tags': [],
+        'n_repeats': ''
+        }
```
linsword13, line 118: nit: the indentation seems wrong?

lib/ramble/ramble/language/application_language.py
```diff
@@ -92,7 +92,7 @@ def workload_group(name, workloads=None, mode=None, **kwargs):
 
     Args:
         name (str): The name of the group
-        workloads (list(str) | None): A list of workloads to be grouped
```
linsword13: Is it complaining about this union operator? I kinda prefer using `|` here...

lib/ramble/ramble/language/modifier_language.py
```diff
@@ -149,7 +149,7 @@ def write_exec_name(self, executable_name, executable, app_inst=None):
             mpi=False,
             redirect='{log_file}',
             output_capture=OUTPUT_CAPTURE.DEFAULT
-        )]
+            )]
```
linsword13: nit: wrong indentation?

linsword13 (commented):
Thanks for cleaning these up! Some minor comments, but otherwise LGTM.

lib/ramble/ramble/language/application_language.py
dapomeroy: Just trying to apply some consistency. I don't have a preference for one over the other, but elsewhere in the file it used 'or' so I standardized on that. Happy to go back the other way and standardize on the pipe.

lib/ramble/ramble/context.py
dapomeroy: This was what sphinx wanted it to be, but it just occurred to me that perhaps the sample dict should be in a code block instead of plain text. If I put it in a code block, it will allow the right indentation

lib/ramble/docs/workspace_config.rst
dapomeroy: This depth of nesting is the paragraph level. Overline is required for the top two levels of heading and optional for all levels below. We don't have a lot of examples of nesting this deep, but in the tutorials most of the paragraph-level headings omit the overline. Do you like it better one way or the other? Maybe I should run through and add overlines to all of them to be consistent?

lib/ramble/ramble/language/application_language.py
linsword13: Yeah I prefer the union `|` here as that's a valid type hint.

lib/ramble/ramble/context.py
linsword13: Ah makes sense, yeah a code block sounds good.

lib/ramble/docs/workspace_config.rst
linsword13: I see, thanks for the context. I think this is good as-is.

lib/ramble/ramble/language/modifier_language.py
dapomeroy: Reverted as it's inside a code block

lib/ramble/ramble/language/application_language.py
dapomeroy: I've updated throughout

linsword13 (approved):

PR #950, opened 2025-04-07
Add support for variants, a when clause, and the default_args clause

douglasjacobsen:
This merge adds support for variants and two new directives.

Variants can be defined within object definitions to give them default values and allow experiments to define them.

The two new directives that are added are `when` and `default_args`. These were backported from Spack's implementation, and allow default definitions for shared arguments to be passed through with blocks.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -650,12 +650,16 @@ class Expander:
             )
 
     def satisfies(
-        self, reqs: List = None, variant_set=None, extra_vars=None, merge_used_stage: bool = True
+        self,
+        reqs: List[str] = None,
+        variant_set=None,
+        extra_vars=None,
+        merge_used_stage: bool = True,
     ):
         """Determine an experiment's variants satisfy a query
 
         Args:
-            req: requirement to check if experiment satisfies
+            reqs: List of string requirements to check if experiment satisfies
             extra_vars: Variable definitions to use with highest precedence
             merged_used_stage: Whether used variables are merged into the
                                set of used variables or not.
@@ -927,12 +931,7 @@ class Expander:
 
             found = False
             for comp in node.comparators:
-                if isinstance(comp, ast.List):
-                    for elt in comp.elts:
-                        rhs_value = self.eval_math(elt)
-                        if lhs_value == rhs_value:
-                            found = True
-                elif isinstance(comp, ast.Set):
+                if isinstance(comp, (ast.List, ast.Set)):
                     for elt in comp.elts:
                         rhs_value = self.eval_math(elt)
                         if lhs_value == rhs_value:
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -11,7 +11,7 @@ import math
 import operator
 import random
 import string
-from typing import Dict
+from typing import Dict, List
 
 import ramble.error
 import ramble.keywords
@@ -649,6 +649,37 @@ class Expander:
                 f'a non-boolean string: "{evaluated}"'
             )
 
+    def satisfies(
+        self, reqs: List = None, variant_set=None, extra_vars=None, merge_used_stage: bool = True
+    ):
+        """Determine an experiment's variants satisfy a query
+
+        Args:
+            req: requirement to check if experiment satisfies
+            extra_vars: Variable definitions to use with highest precedence
+            merged_used_stage: Whether used variables are merged into the
+                               set of used variables or not.
+
+        Returns:
+            boolean: True or False, based if the experiment's variants satisfy
+                     the input requirement.
+        """
+
+        if variant_set is not None:
+            variant_definitions = variant_set.as_set()
+        else:
+            variant_definitions = set()
+
+        satisfied = True
+        if reqs is not None:
+            for req in reqs:
+                exp_req = self.expand_var(
+                    req, extra_vars=extra_vars, merge_used_stage=merge_used_stage
+                )
+
+                satisfied = satisfied and exp_req in variant_definitions
+        return satisfied
+
     @staticmethod
     def expansion_str(in_str):
         return f"{ExpansionDelimiter.left}{in_str}{ExpansionDelimiter.right}"
@@ -897,6 +928,11 @@ class Expander:
             found = False
             for comp in node.comparators:
                 if isinstance(comp, ast.List):
+                    for elt in comp.elts:
+                        rhs_value = self.eval_math(elt)
+                        if lhs_value == rhs_value:
+                            found = True
+                elif isinstance(comp, ast.Set):
                     for elt in comp.elts:
                         rhs_value = self.eval_math(elt)
                         if lhs_value == rhs_value:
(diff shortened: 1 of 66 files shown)


lib/ramble/ramble/variants.py
```diff
@@ -0,0 +1,313 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import collections.abc
```
linsword13: nit: looks like only `Sequence` is used? Maybe just `from collections.abc import Sequence`?

lib/ramble/ramble/expander.py
```diff
@@ -901,6 +932,11 @@ def _eval_comp_in(self, node):
                         rhs_value = self.eval_math(elt)
                         if lhs_value == rhs_value:
                             found = True
+                elif isinstance(comp, ast.Set):
```
linsword13: nit: maybe this can be combined with the above with `isinstance(comp, (ast.List, ast.Set))`?

linsword13 (commented):
Exciting stuff! A couple nit comments for now ;)

lib/ramble/ramble/expander.py
```diff
@@ -649,6 +649,37 @@ def evaluate_predicate(self, in_str, extra_vars=None, merge_used_stage: bool = T
                 f'a non-boolean string: "{evaluated}"'
             )
 
+    def satisfies(
```
linsword13, line 652: Maybe have some expander unit tests for it?

lib/ramble/ramble/expander.py
```diff
@@ -649,6 +649,37 @@ def evaluate_predicate(self, in_str, extra_vars=None, merge_used_stage: bool = T
...
+        self, reqs: List = None, variant_set=None, extra_vars=None, merge_used_stage: bool = True
```
linsword13: Is `reqs` a list of strings?

lib/ramble/ramble/variants.py
douglasjacobsen: Thanks, fixed!

lib/ramble/ramble/expander.py
douglasjacobsen: Done!

lib/ramble/ramble/expander.py
douglasjacobsen: Ahh good call out. I should have done this. But I added some now.

lib/ramble/ramble/expander.py
douglasjacobsen: Yup, thanks. I updated the type hint, and the description.

lib/ramble/ramble/language/shared_language.py
```diff
@@ -135,23 +140,35 @@ def define_compiler(name, pkg_spec, compiler_spec=None, compiler=None, package_m
     """
 
     def _execute_define_compiler(obj):
+        when_list = ramble.language.language_helpers.build_when_list(
+            when, obj, name, "define_compiler"
+        )
+
+        if package_manager is not None:
```
linsword13, line 147: Is there concern that this `package_manager` option is in effect ignored?

lib/ramble/ramble/language/language_base.py
```diff
@@ -79,6 +82,22 @@ def __new__(cls, name, bases, attr_dict):
 
         return super().__new__(cls, name, bases, attr_dict)
 
+    @staticmethod
```
linsword13: Can these be made as module-level functions instead? https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators

lib/ramble/ramble/language/workflow_manager_language.py
```diff
@@ -45,3 +45,18 @@ def _define_wm_variable(wm):
...
+
+@workflow_manager_directive(dicts=())
+def workflow_manager_family(*names: str):
+    """Add a new family to this workflow manager
+
+    Args:
+        name (str): Name of family to apply to this workflow manager
+    """
+
+    def _define_workflow_manager_family(pm):
```
linsword13: nit: `wm`?

lib/ramble/ramble/language/language_base.py
```diff
@@ -174,7 +193,33 @@ class Foo(ApplicationBase):
...
+                kwargs = dict()
+                for default_args in DirectiveMeta._default_args:
+                    kwargs.update(default_args)
+                kwargs.update(_kwargs)
+
+                # Inject when arguments from the context
+                if DirectiveMeta._when_constraints_from_context:
+                    # Check that directives not yet supporting the when= argument
+                    # are not used inside the context manager
+                    if decorated_function.__name__ == "version":
```
linsword13: Is `version` the only unsupported one?

lib/ramble/ramble/language/shared_language.py
douglasjacobsen: I think there are three options...

1: Try to convert the `package_manager` directive into a when clause (TBF, this won't generally work, as we typically write `package_manager='spack*'` and that's not a valid when clause)
2: Warn, and let it continue without an error.
3: Hard error and make people convert them.

I've implemented (2) but I can also see an argument for (3). IMO (1) is not a good idea though.

lib/ramble/ramble/language/language_base.py
douglasjacobsen: No. Honestly, `version` isn't even a directive.

But, 90% of our directives are unsupported currently with when. I aim to support all of them though. I'll invert this so it is more accurate.

lib/ramble/ramble/language/workflow_manager_language.py
douglasjacobsen: Done

lib/ramble/ramble/language/language_base.py
douglasjacobsen: I think this is done. Let me know if I didn't do what you intended.

lib/ramble/ramble/language/shared_language.py
linsword13: Yeah sounds good with the current impl then.

linsword13 (dismissed):

rfbgo (dismissed):

var/ramble/repos/builtin.mock/applications/when-variants/application.py
```diff
@@ -0,0 +1,48 @@
...
+
+    variant(
+        "inc_zlib",
+        default=True,
+        values=[True, False],
+        description="Test boolean variant",
+    )
+
+    with when("package_manager_family=spack"):
+        with when("+inc_zlib"):
```
rfbgo: Q: Do you think it's worth the effort to support some sort of `and` semantic on a single line instead of exclusively nesting?

linsword13 (approved):

dapomeroy (approved):

PR #952, opened 2025-04-07
Support import same-directory sources from an object file

linsword13:
This is done by adding a `is_package` method to the `RepoLoader`.


douglasjacobsen (approved):

douglasjacobsen:
LGTM. I don't have any concerns, but I want @rfbgo to review too, just to make sure.

PR #959, opened 2025-04-12
Update format specifier extraction to be more robust

douglasjacobsen:
This merge updates the way the expander extracts format specifiers. Previously, it would split the input string on colons, and had a hard coded assumption that there would only be two portions to this split. After splitting, it would try to use the first as the keyword, and the second as the format specifier (then throwing everything after that away).

When the input string is a formatted dictionary, this breaks and results in improperly formatted output.

This merge changes this behavior to use regular expression to detect the format specifier (if one exists), and then attempts to apply it to the detected keyword.

Tests are added for this case.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -211,11 +211,6 @@ class ExpansionNode:
                     self.value = "{}"
                     return
 
-                # TODO:
-                # - Make sure there's only one format specifier
-                # - Detect it from the end backwards
-                # - Do not format broken format specifiers
-
                 keyword = replaced_contents[1:-1]
                 format_match = format_spec_regex.search(keyword)
                 required_passthrough = False
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -262,9 +262,9 @@ class ExpansionNode:
                         )
                         required_passthrough = False
                     except ValueError:
-                        self.value += f":{format_spec}"
+                        self.value = replaced_contents
                     except KeyError:
-                        self.value += f":{format_spec}"
+                        self.value = replaced_contents
 
                 if required_passthrough:
                     self.value = f"{{{self.value}}}"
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/expander.py
```diff
@@ -80,6 +81,9 @@ def _maybe(expander, var_name, default=""):
     "re_search": _re_search,
 }
 
+# Format Spec Regex:
+format_spec_regex = re.compile(r"(?P<kw>.*):(?P<format_spec>\S+)$")
```
linsword13, line 85: For an input like `{"a": "b"}`, is it essentially using the space to separate out "dictionary-like" versus "format-specifier"?

douglasjacobsen: Yes, but also if you wrote `{"a":"b"}` with this newer format specifier detection, it should result in an output string of: `{"a":"b"}`. Whereas before it would have given a weird: `"a"`

linsword13: For an input like `"{'key1':'val1'}"`,  looks like that will get expanded into `"key1:'val1'"` (i.e., without the curly bracket and the `'val1'` part treated as a format specifier.) I guess my concern is that that outcome is quite different from the input  `"{'key1': 'val1'}"`, which has an extra whitespace after `:`.

douglasjacobsen: OK, this should be resolved now. I added a test for this as well.

linsword13, line 85: One more nit: is it OK to make the "kw" group non-greedy `(?P<kw>.*?)`, or is it intentional that we want it to consume all the way up to the last ":"?

linsword13 (dismissed):

lib/ramble/ramble/expander.py
douglasjacobsen: Yes, it is supposed to match the last `:` specifically.

linsword13 (approved):

PR #960, opened 2025-04-12
Add a modifier to support running experiment from a directory other than experiment_run_dir

linsword13:


var/ramble/repos/builtin/modifiers/run-directory/modifier.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def retain_files(self):
+        glob_pattern = self.expander.expand_var_name("retained_files_glob")
+        if not glob_pattern:
+            return []
+        target_dir = self.expander.expand_var_name("target_directory")
+        source_dir = self.expander.expand_var_name("source_directory")
+        if target_dir == source_dir:
+            return []
+        return [f"cp -r {target_dir}/{glob_pattern} {source_dir}/"]
```
douglasjacobsen: It might be better to do an rsync, if that is common enough to depend on. That might be less expensive when you have the same large files in both directories.

linsword13: Great call. I initially didn't use it as rsync is not available by default in our hpc image. But indeed it'd be a good perf improvement.

Added the rsync use if it's available. (I thought about having it required, but figured that might be too strict.)

douglasjacobsen: Awesome, thanks!

douglasjacobsen (approved):

PR #962, opened 2025-04-14
Add slurm-pyxis workflow manager

dapomeroy:
Adds a slurm-pyxis workflow manager to add pyxis-specific srun args to the experiment command when using Slurm.


var/ramble/repos/builtin/workflow_managers/slurm-pyxis/workflow_manager.py
```diff
@@ -0,0 +1,45 @@
...
+class SlurmPyxis(SlurmBase):
+    """Pyxis support for Slurm workflow manager"""
+
+    name = "slurm-pyxis"
+
+    maintainers("dapomeroy")
+
+    tags("workflow", "slurm", "pyxis")
+
+    def __init__(self, file_path):
```
linsword13: This `init` can be skipped given that it only calls into the parent init.

var/ramble/repos/builtin/workflow_managers/slurm-pyxis/workflow_manager.py
```diff
@@ -0,0 +1,45 @@
...
+
+    workflow_manager_variable(
+        name="container_path",
+        default="",
+        description="Path to container image",
+    )
+
+    workflow_manager_variable(
+        name="container_env_vars",
+        default="",
```
linsword13: Would the empty default cause an error to `srun` (i.e., `--container-env` without an argument)?

var/ramble/repos/builtin/workflow_managers/slurm-pyxis/workflow_manager.py
dapomeroy: Done

var/ramble/repos/builtin/workflow_managers/slurm-pyxis/workflow_manager.py
dapomeroy: It didn't cause an error when I tried it, but it's probably better to be safe and check if it's empty when constructing srun args.

douglasjacobsen (approved):
I think in general there are things we should consider to improve on this in the future, but for the time being I think they are fine to ignore.

These include:
- How to extend the default `srun_args` with other flags easily
- What should srun_args do if {container_env_vars} is empty?
- What should srun_args do if {container_mounts} is empty (this is less likely)?

PR #963, opened 2025-04-15
Add a modifier to output various os-level tunables

linsword13:
Example output:

```
default (null) context figures of merit:
    possible hostname = linsword-13.c.googlers.com
    modifier::tunables::address-space-randomization = 2
    modifier::tunables::numa-balancing = 0
    modifier::tunables::smt-active = 1
    modifier::tunables::thp-enabled = always
    modifier::tunables::thp-defrag = madvise
    modifier::tunables::explicit-hugepage-count = 0
    modifier::tunables::explicit-hugepage-size = 2048 kB
```

### DIFF
--- a/var/ramble/repos/builtin/modifiers/tunables/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/tunables/modifier.py
@@ -1,0 +1,120 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import os
+from collections import defaultdict
+
+from ramble.modkit import *
+
+_TUNABLE_LOG = "{experiment_run_dir}/tunables.log"
+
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/modifiers/tunables/modifier.py
```diff
@@ -0,0 +1,122 @@
...
+    ]
+
+    register_builtin("get_tunable_info", injection_method="append")
+
+    def get_tunable_info(self):
+        """Collect tunable information."""
+        cmd_list = []
+        for conf in self.info_list:
+            include_name = conf.get("include_name_in_output", True)
+            prefix = f'{conf["name"]}: ' if include_name else ""
```
douglasjacobsen: This seems to have an assumption that the configuration on one compute node is the same for all compute nodes. I don't necessarily see any issues with that, but can you think of a reason we might want this to be all of the compute nodes instead?

(I'm fine if the answer is no, just wanted to make sure we gave it some thought)

linsword13: Yeah that's a keen observation. I did assume uniformity when writing this. But thinking about it, that's honestly not something to rely upon. Will update to fetch from all involved nodes.

linsword13: Updated to check for all compute nodes ;)

douglasjacobsen (approved):

PR #969, opened 2025-04-18
Add class to represent software info produced by package managers

rfbgo:
PR primarily introduces a class called `SoftwareInfo` that is adopted by all package managers. This also allows the user managed to output the packages if specified in yaml

I snook in a few other things:

1. Added an `output_prefix` to package managers to let them output with `_` re-coverted to `-`
2. Removed dead file `lib/ramble/ramble/package_manager_types/runner.py`
3. Fix a bug in spack package manager output formatting (oversplitting on spaces)


var/ramble/repos/builtin/package_managers/environment-modules/package_manager.py
```diff
@@ -119,6 +119,7 @@ def get_package_list(self, workspace):
         list_file = app_inst.expander.expand_var(
             f"{{experiment_run_dir}}/{self._list_file}"
         )
+        print(list_file)
```
douglasjacobsen: Do you want this print here?

rfbgo: Nope! Thanks for catching it

douglasjacobsen (approved):

PR #970, opened 2025-04-19
Enforce .tpl extension to all template sources used by register_template

linsword13:


lib/ramble/ramble/application.py
```diff
@@ -2437,6 +2437,7 @@ def _object_templates(self, workspace):
         run_dir = self.expander.experiment_run_dir
         replacements = workspace.workspace_paths()
         expander = self.expander
+        tpl_ext = ".tpl"
```
douglasjacobsen: I kind of think this extensions should live somewhere else. We use it in the workspace as well, so that should probably use the same extension variable.

douglasjacobsen (changes requested):
This is great, thanks! One small change (and maybe adding something that's not in this PR already)

lib/ramble/ramble/application.py
linsword13: Great point, updated.

rfbgo (approved):
LGTM!

douglasjacobsen (approved):

PR #973, opened 2025-04-21
decouple meaning of RAMBLE_STATUS and experiment_status

rfbgo:
This PR changes the semantics of RAMBLE_STATUS and experiment_status in the analyze output 

Previously they had the identical bool (FAILED/SUCCESS) value. This change makes `experiment_status` able to retain it's value for unsuccessful experiments (so you can see if they were cancelled, etc)

eg:

```
7457       "RAMBLE_STATUS": "FAILED",
7458       "N_REPEATS": 0,
7459       "application_name": "osu-micro-benchmarks",
7460       "application_namespace": "osu-micro-benchmarks",
7461       "simplified_application_namespace": "osu-micro-benchmarks",
7462       "workload_name": "osu_mbw_mr",
7463       "workload_namespace": "osu-micro-benchmarks.osu_mbw_mr",
7464       "simplified_workload_namespace": "osu-micro-benchmarks-osu-mbw-mr",
7465       "experiments_file": "/Users/robertbird/ramble/var/ramble/workspaces/osu_mbw_mr/all_experiments",
7466       "experiment_name": "h3_2_88.5",
7467       "experiment_hash": "cdd0c44c209e9ad6a845d5233128eff4e8628b63e0b176eaa912e89b900d82fc",
7468       "experiment_status": "SETUP",
```

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -2309,19 +2309,29 @@ class ApplicationBase(metaclass=ApplicationMeta):
             with lk.ReadTransaction(exp_lock):
                 with open(status_path) as f:
                     status_data = spack.util.spack_json.load(f)
-                self.variables[self.keywords.experiment_status] = ExperimentStatus(
-                    status_data[self.keywords.experiment_status]
-                )
+                    self.set_status(ExperimentStatus(status_data[self.keywords.experiment_status]))
         else:
             self.set_status(ExperimentStatus.UNKNOWN)
 
     def set_status(self, status=ExperimentStatus.UNKNOWN):
         """Set the status of this experiment"""
         self.variables[self.keywords.experiment_status] = status
+        self.set_ramble_status(status)
 
     def get_status(self):
         """Get the status of this experiment"""
         return self.variables[self.keywords.experiment_status]
+
+    def get_ramble_status(self):
+        """Get the RAMBLE_STATUS of this experiment (boolifyied status)"""
+        return self.variables[self.keywords.RAMBLE_STATUS]
+
+    def set_ramble_status(self, status):
+        """Set the RAMBLE_STATUS (boolifyied status) of this experiment"""
+
+        self.variables[self.keywords.RAMBLE_STATUS] = status
+        if status != ExperimentStatus.SUCCESS:
+            self.variables[self.keywords.RAMBLE_STATUS] = ExperimentStatus.FAILED
 
     register_phase("write_status", pipeline="analyze", run_after=["analyze_experiments"])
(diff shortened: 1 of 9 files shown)


douglasjacobsen:
Thanks, this is awesome. Do you know if it would be possible to make `RAMBLE_STATUS` something that can be referred to in `where` clauses?

It might be nice to be able to easily do something like:

```
--where "'{RAMBLE_STATUS}' == 'FAILED'"
```

would would map in from the various `experiment_status` values.

rfbgo:
> Thanks, this is awesome. Do you know if it would be possible to make `RAMBLE_STATUS` something that can be referred to in `where` clauses?
> 
> It might be nice to be able to easily do something like:
> 
> ```
> --where "'{RAMBLE_STATUS}' == 'FAILED'"
> ```
> 
> would would map in from the various `experiment_status` values.

I don't think it's that hard, maybe a bit fiddly

As I understand it now (without trying) RAMBLE_STATUS only exists in the ExperimentResult object (as `app_inst.result.status`) so would need to be propagated to the expander (not sure what that looks like, but feels possible)

Incidentally this is also true for a few other things that might be nice to query on, such as n_repeats (and everything else in init): https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/experiment_result.py#L33

rfbgo:
> > Thanks, this is awesome. Do you know if it would be possible to make `RAMBLE_STATUS` something that can be referred to in `where` clauses?
> > It might be nice to be able to easily do something like:
> > ```
> > --where "'{RAMBLE_STATUS}' == 'FAILED'"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would would map in from the various `experiment_status` values.
> 
> I don't think it's that hard, maybe a bit fiddly
> 
> As I understand it now (without trying) RAMBLE_STATUS only exists in the ExperimentResult object (as `app_inst.result.status`) so would need to be propagated to the expander (not sure what that looks like, but feels possible)
> 
> Incidentally this is also true for a few other things that might be nice to query on, such as n_repeats (and everything else in init): https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/experiment_result.py#L33

I pushed a commit (9c369e851d6c1bcaa64e8c5450e4745dfc403dcb) to try and directly address this for RAMBLE_STATUS. We might need to do follow on work to identify other relevant vars 

Heads up, I added RAMBLE_STATUS as a capitalized keyword, which did not seem ideal, but is required to maintain backwards compatibility of outputs

douglasjacobsen (approved):

PR #976, opened 2025-04-22
Fix regex

dapomeroy:
Fixes issue parsing scientific notation


douglasjacobsen (approved):
This looks good. We might consider promoting this to a shared location so more applications can easily use scientific notation FOMs without having to replicate this regex.

PR #978, opened 2025-04-22
Add a workflow manager to simplify using intel-mpi with slurm

linsword13:
This workflow sets the `--mpi=pmi2` argument for `srun`, and it also attempts to use a default for setting the `I_MPI_PMI_LIBRARY` env var.

### DIFF
--- a/var/ramble/repos/builtin/workflow_managers/slurm-intel-mpi/workflow_manager.py
+++ b/var/ramble/repos/builtin/workflow_managers/slurm-intel-mpi/workflow_manager.py
@@ -1,0 +1,46 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/workflow_managers/slurm-intel-mpi/workflow_manager.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+    register_builtin("set_pmi2_path", injection_method="prepend")
+
+    def set_pmi2_path(self):
+        lib_path = self.app_inst.expander.expand_var_name("pmi2_lib_path")
+        if not lib_path:
+            return []
+        return [
+            f'export I_MPI_PMI_LIBRARY="${{I_MPI_PMI_LIBRARY:-{lib_path}}}"'
```
douglasjacobsen, line 45: It might not be important for this PR specifically, but it might be nice to use the environment variable things to make this shell agnostic.

FWIW, I'd be happy with this being a follow on task, if you can make it.

douglasjacobsen (approved):

var/ramble/repos/builtin/workflow_managers/slurm-intel-mpi/workflow_manager.py
linsword13: Yeah that's a good point. I thought about doing that, but I want the "only set if not already set" semantic which isn't provided by the shell modification util. I will have this as a follow-on task.

PR #980, opened 2025-04-23
Update default workflow manager to be user-managed

rfbgo:
Changed the default workflow manager to be UserManaged (and not none, which is now disallowed / assumed to be UserManaged) 

Previously we instantiated the base manager. This pulls the implies functionality into a concrete class

### DIFF
--- a/etc/ramble/defaults/variants.yaml
+++ b/etc/ramble/defaults/variants.yaml
@@ -1,3 +1,3 @@
 variants:
   package_manager: user-managed
-  workflow_manager: null
+  workflow_manager: user-managed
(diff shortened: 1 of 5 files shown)


douglasjacobsen (approved):

douglasjacobsen (changes requested):
I almost merged this, but then I remember that I think we need to change the actual default configuration files.

Can you make sure the file `etc/ramble/defaults` is updated to have the correct default for `variants:workflow_manager` ?

rfbgo:
> I almost merged this, but then I remember that I think we need to change the actual default configuration files.
> 
> Can you make sure the file `etc/ramble/defaults` is updated to have the correct default for `variants:workflow_manager` ?

Done! Thanks for catching that (and as discussed, filed a bug around the issue that hid this)

douglasjacobsen (approved):

PR #987, opened 2025-04-28
Allow directives to ignore `when` and `default_args` clauses

dapomeroy:


lib/ramble/ramble/application.py
```diff
@@ -908,11 +908,20 @@ def build_modifier_instances(self):
...
+            for type, obj in self._objects():
+                if type == ramble.repository.ObjectTypes.modifiers:
+                    obj_variants = getattr(obj, "object_variants", None)
+                    if obj_variants is not None:
+                        self.object_variants.merge_multi_value_variants(
+                            getattr(obj, "object_variants")
+                        )
+
             # Add this modifiers required variables for validation
             self.keywords.update_keys(mod_inst.get_required_variables())
```
douglasjacobsen, line 918: Realistically this shouldn't matter, but I think this line should move to the loop below this one. That way a modifier can key off of some other modifier's mode, even if it's added after the current modifier.

dapomeroy: Got it, fixed!

lib/ramble/ramble/application.py
```diff
@@ -916,12 +917,19 @@ def build_modifier_instances(self):
...
+            # Ensure no expand vars are set correctly for modifiers
             for var in mod_inst.no_expand_vars():
                 self.expander.add_no_expand_var(var)
                 mod_inst.expander.add_no_expand_var(var)
 
+            # Set standard variants for all modifiers
+            obj_variants = getattr(mod_inst, "object_variants", None)
+            if obj_variants is not None:
+                self.object_variants.merge_multi_value_variants(
+                    getattr(mod_inst, "object_variants")
```
linsword13: Is this just `obj_variants`?

linsword13 (dismissed):
LGTM with a nit comment

lib/ramble/ramble/application.py
dapomeroy: It is. I just copied the syntax from above: https://github.com/GoogleCloudPlatform/ramble/blob/2884819c56a4b31061190df74053dacc324cb145/lib/ramble/ramble/application.py#L297

Is there a reason to prefer a second getattr call for these instead of just using the local variable?

linsword13: I think it should be fine to just use the local variable.

dapomeroy: OK, I've updated both this and the original to use the local variable

linsword13 (approved):

rfbgo (approved):

PR #994, opened 2025-04-30
Add 'when' to register_phase directive

dapomeroy:


douglasjacobsen (approved):

lib/ramble/ramble/test/when.py
```diff
@@ -0,0 +1,48 @@
...
+
+pytestmark = pytest.mark.usefixtures(
+    "mutable_config", "mutable_mock_workspace_path", "mutable_mock_apps_repo", "mock_modifiers"
+)
+
+config = RambleCommand("config")
+workspace = RambleCommand("workspace")
+
+
+def test_register_phase_when(request):
```
douglasjacobsen, line 20: It might actually be a good idea to replicate this test, and use `when_phase:false` and assert that the test phase was not in the log file.

douglasjacobsen (changes requested):
Thanks, this is great. One small change to help ensure the functionality is what we're looking for.

lib/ramble/ramble/test/when.py
dapomeroy: Updated

douglasjacobsen (approved):

PR #998, opened 2025-05-01
Allow repeated `software_specs` in object definitions

douglasjacobsen:
This merge enables the `software_spec` and `define_compiler` directives to define the same packages multiple times with different definitions that would conflict.

The primary use case for this is while using `when` arguments to change the circumstances for using each of the definition. Errors will still be presented if multiple definitions would actually conflict, to help prevent invalid object definitions.

### DIFF
--- a/lib/ramble/ramble/util/spec_utils.py
+++ b/lib/ramble/ramble/util/spec_utils.py
@@ -7,33 +7,23 @@
 # except according to those terms.
 
 
-def specs_equiv(spec1, spec2):
-    all_keys = set()
-    for key in spec1.keys():
-        if spec1[key] is not None:
-            all_keys.add(key)
-    for key in spec2.keys():
-        if spec2[key] is not None:
-            all_keys.add(key)
+def specs_conflict(new, existing, prefix="", skip_conflicting_when=False):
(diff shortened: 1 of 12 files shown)


lib/ramble/ramble/util/spec_utils.py
```diff
@@ -6,34 +6,120 @@
...
+            escaped_val = val.replace("@", "@@")
+            output += f"{indentation}{color_key}: {escaped_val}\n"
+        return output
+
+    def __str__(self):
+        self_dict = self.to_dict(prefix="")
+        self_dict["prefix"] = self.prefix
+        return str(self_dict)
+
+    def copy(self):
```
rfbgo, line 90: Is this any different from just calling something like `deepcopy`?

rfbgo (dismissed):
LGTM. One nit/questions

lib/ramble/ramble/util/spec_utils.py
douglasjacobsen: Nope, it's not. I pushed an update that swaps it.

linsword13 (approved):

PR #999, opened 2025-05-01
Convert SoftwareInfo to dict ahead of output

rfbgo:


lib/ramble/ramble/test/end_to_end/package_manager_provenance.py
```diff
@@ -68,6 +67,21 @@ def test_spack_package_manager_provenance_zlib(mock_applications, request):
         assert "spack packages:" in data
         assert "zlib" in data
 
+    workspace("analyze", "-f", "json", global_args=global_args)
```
linsword13, line 86: nit: might as well add yaml to the test?

rfbgo: Happy to. I decided not to since the test is annoyingly slow (and I dislike yaml as an output 🥲 ) but happy to if you think it's worth while

linsword13: Haha, fair point ;) Nope having the current test is fine (and it's great that it plugs a hole in our coverage!)

linsword13 (dismissed):

lib/ramble/ramble/test/end_to_end/package_manager_provenance.py
```diff
@@ -59,14 +60,29 @@ def test_spack_package_manager_provenance_zlib(mock_applications, request):
...
+    with open(results_file) as f:
+        import json
+
+        data = json.load(f)
+        for data in data["experiments"]:
+            pkg_list = data["SOFTWARE"]["spack"]
+            names = [pkg["name"] for pkg in pkg_list]
+            assert "SOFTWARE" in data
+            assert "spack" in data["SOFTWARE"]
+            assert "zlib" in names
```
douglasjacobsen, line 101: If we're using an actual installation of zlib, I think something needs to happen to uninstall the environment that was installed here, otherwise we're messing with people's spack install directories just by running tests.

douglasjacobsen (changes requested):

lib/ramble/ramble/test/end_to_end/package_manager_provenance.py
douglasjacobsen: To complicate things more... we probably only want to uninstall zlib if it wasn't installed before we ran `workspace setup`.

So, we might need some logic around making sure we clean this up properly.

rfbgo: Yea that is a really good point.. that makes testing this way harder. I was thinking about this from the CI perspective where I don't mind installing zlib..

douglasjacobsen: Hmm... something you might be able to do is add a config.yaml into the auxiliary software files that changes the install dir to be the same as the environment path...

rfbgo: That's a really great idea! Updated

douglasjacobsen (approved):

PR #1002, opened 2025-05-01
Fix incorrect variable definition splitting

douglasjacobsen:
This merge fixes an issue where `manage experiments` would incorreclty split variable definitions that included multiple equals signs.

A test is added to help ensure this doesn't happen again as well.

### DIFF
--- a/lib/ramble/ramble/test/cmd/workspace_manage.py
+++ b/lib/ramble/ramble/test/cmd/workspace_manage.py
@@ -38,10 +38,18 @@ def test_manage_variable_multiple_equals(request, tmpdir):
         "test_var=foo=bar",
         "-v",
         "test_list_var=[ val1 = val2, val3=val4]",
+        "-v",
+        "test_other_var = test1=test2=test3",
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/test/cmd/workspace_manage.py
```diff
@@ -0,0 +1,54 @@
...
+        "--wf",
+        "water_bare",
+        "-v",
+        "n_nodes=1",
+        "-v",
+        "n_ranks=1",
+        "-v",
+        "test_var=foo=bar",
+        "-v",
+        "test_list_var=[ val1 = val2, val3=val4]",
```
linsword13, line 40: nit: might be fun to add some extra whitespaces around the first "=", just to ensure the "\s*" matching works as expected.

douglasjacobsen: Good call. Added.

linsword13 (approved):

PR #1011, opened 2025-05-06
Add codecov io upload

rfbgo:
First test to enable code cov reporting

### DIFF
--- a/share/ramble/qa/run-unit-tests
+++ b/share/ramble/qa/run-unit-tests
@@ -73,13 +73,9 @@ if [ $? != 0 ]; then
     ERROR=1
 fi
 
-
-echo $PUSH_CODECOV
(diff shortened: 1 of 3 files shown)


linsword13 (dismissed):

share/ramble/qa/run-unit-tests
```diff
@@ -74,6 +74,12 @@ if [ $? != 0 ]; then
 fi
 
 
+echo $PUSH_CODECOV
```
rfbgo: I can remove this

linsword13: Ah good call. Yeah SG to remove this, given that there's another echo in the if block.

rfbgo: done

linsword13 (approved):

PR #1016, opened 2025-05-07
Allow workspace info to print variants

rfbgo:
Example:

```
      Experiment 10: osu-micro-benchmarks.osu_mbw_mr.h3_2_88.3
        Experiment Tags: ['synthetic-benchmarks']
        Variants:
          package_manager: user-managed
          workflow_manager: user-managed
```

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -649,6 +649,10 @@ def workspace_info_setup_parser(subparser):
         "--phases", action="store_true", help="If set, phase information will be printed"
     )
 
+    subparser.add_argument(
+        "--variants", action="store_true", help="If set, experiment variants will be printed"
+    )
+
     arguments.add_common_arguments(subparser, ["where", "exclude_where", "filter_tags"])
 
     subparser.add_argument(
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -802,6 +806,11 @@ def workspace_info(args):
                     if args.tags:
                         color.cprint("        Experiment Tags: " + str(app_inst.experiment_tags))
 
+                    if args.variants:
```
linsword13, line 810: Do we also want to make it so that when `--verbose 2` (or 1) this gets printed?

rfbgo: Really great idea! Updated

linsword13 (dismissed):

douglasjacobsen (changes requested):
Nit;

https://github.com/GoogleCloudPlatform/ramble/pull/1016/files#diff-68927c3d0c18af55615ca21966a2c9d3c35fd2567a438b06654b4c78b8f78d11R665

should be updated to say that it prints variants too.

rfbgo:
> Nit;
> 
> https://github.com/GoogleCloudPlatform/ramble/pull/1016/files#diff-68927c3d0c18af55615ca21966a2c9d3c35fd2567a438b06654b4c78b8f78d11R665
> 
> should be updated to say that it prints variants too.

Good catch, thanks Doug!

douglasjacobsen (approved):

PR #1017, opened 2025-05-08
Add a test that makes it easier to validate object setup and analyze

linsword13:
This new test would look for the presence of `test_cases` directory in the targeted Ramble repo (by default the `builtin` repo), setup the workspace with the given `ramble.yaml`, copy over extra run artifacts, run `workspace analyze`, and check the analyze output matches with the given one.

The goal is to allow object writers an easy way to validate their implementations, by including a `test_cases` directory, without needing to write explicit test code.


conftest.py
```diff
@@ -792,3 +795,17 @@ def pytest_generate_tests(metafunc):
...
+    if "test_case_path" in metafunc.fixturenames:
+        repo_path = pathlib.Path(
+            metafunc.config.getoption("--repo-path") or ramble.paths.builtin_path
+        )
+        test_data_paths = []
+        for child_path in repo_path.rglob("test_cases"):
+            if child_path.is_dir():
+                for test_dir in child_path.iterdir():
+                    if test_dir.is_dir():
+                        config_file = test_dir / "ramble.yaml"
```
douglasjacobsen: Could we also test with a variables.yaml file (Or maybe any *.yaml file?)

Then we can easily meld those into `ramble workspace manage experiments` calls.

linsword13: Great call! Updated to now use a `configs` directory.

var/ramble/repos/builtin/applications/osu-micro-benchmarks/test_cases/osu_latency/configs/ramble.yaml
```diff
@@ -0,0 +1,34 @@
+ramble:
```
douglasjacobsen: Super nitpicky... but can we replace this with minimal config files, and call the `manage` methods in the test instead of having the full YAML here?

douglasjacobsen (changes requested):

var/ramble/repos/builtin/applications/osu-micro-benchmarks/test_cases/osu_latency/configs/ramble.yaml
linsword13: For sure, I updated so that the ramble config is split into separate applications.yaml and software.yaml, and the test itself will look for the presence of `includes` folder (and call `manage includes` accordingly.) Does this look a bit cleaner than before? Or do we want to also use `manage` to generate the experiment and software? (I think that'd be cool, just unsure how we want a test writer to describe things.)

linsword13: Updated now to have a `setup.yaml` to list out the `workspace manage` commands to run, for constructing the ramble config.

douglasjacobsen (approved):

PR #1018, opened 2025-05-08
Package manager load/unload environment commands

douglasjacobsen:
This merge adds methods to package manager classes that can be used to determine their commands for loading or unloading environments within an experiment. These can be used by other objects to construct multi-step templates that might need to re-load (or swap) environments.

### DIFF
--- a/var/ramble/repos/builtin/package_managers/environment-modules/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/environment-modules/package_manager.py
@@ -8,6 +8,7 @@
 
 import os
 import re
+from typing import List
 
 import llnl.util.filesystem as fs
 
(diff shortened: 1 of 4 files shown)


var/ramble/repos/builtin/package_managers/environment-modules/package_manager.py
```diff
@@ -140,3 +140,9 @@ def get_package_list(self, workspace):
                         )
                     )
         return pkg_list
+
+    def environment_load_commands(self) -> List[str]:
```
linsword13, line 145: Looks like it's complaining about the `List` not imported.

linsword13 (commented):
LGTM! Just a minor import issue that fails the unit test.

var/ramble/repos/builtin/package_managers/environment-modules/package_manager.py
douglasjacobsen: Thanks, should be fixed now.

linsword13 (approved):

PR #1024, opened 2025-05-12
Add 'when' to Figures of Merit and FOM Contexts

dapomeroy:
Adds 'when' clause to FOMs and FOM contexts, and enables FOMs to key off both context and FOM name

### DIFF
--- a/lib/ramble/docs/dev_guides/application_dev_guide.rst
+++ b/lib/ramble/docs/dev_guides/application_dev_guide.rst
@@ -312,13 +312,17 @@ package manager variant:
         when=["package_manager_family=eessi"],
     )
 
-The `when` clause can also be used with modifier and modifier mode, using the
-following syntax:
+The following standard variant definitions can be used in the `when` clause,
+along with any variants created in definition files:
 
 .. code-block:: python
 
-    when=["modifier=modifier-name"]
-    when=["modifier-name_mode=mode"]
+    package_manager  
+    package_manager_family  
+    workflow_manager  
+    workflow_manager_family  
+    modifier  
+    <mod-name>_mode 
 
 --------------------------
 Package Manager Directives
--- a/lib/ramble/docs/dev_guides/application_dev_guide.rst
+++ b/lib/ramble/docs/dev_guides/application_dev_guide.rst
@@ -277,6 +277,49 @@ that should be used when referencing file paths in application definitions. This
 helps with Ramble to properly mock out these paths during unit testing, where the
 files may not exist under the dry-run setting.
 
+^^^^^^^^^^^^^^^^^
+Conditional Logic
+^^^^^^^^^^^^^^^^^
+
+Ramble supports conditional logic to allow developers to set directives based on
+their :ref:`variants<variants-config>`. Directives supporting this logic will
+have a `when` argument to accept conditions. Multiple directives can be grouped
+under the same when clause by enclosing them in a `with` statement and using the
+`when` directive.
+
+For example, the Gromacs definition defines a default software spec based on the
+package manager variant:
+
+.. code-block:: python
+
+    with when("package_manager_family=spack"):
+        define_compiler("gcc9", pkg_spec="gcc@9.3.0")
+
+        software_spec(
+            "impi2018",
+            pkg_spec="intel-mpi@2018.4.274",
+        )
+
+        with default_args(compiler="gcc9"):
+            software_spec(
+                "spack_gromacs",
+                pkg_spec="gromacs@2020.5",
+            )
+
+    software_spec(
+        "eessi_gromacs",
+        pkg_spec="GROMACS/2024.1-foss-2023b",
+        when=["package_manager_family=eessi"],
+    )
+
+The `when` clause can also be used with modifier and modifier mode, using the
+following syntax:
+
+.. code-block:: python
+
+    when=["modifier=modifier-name"]
+    when=["modifier-name_mode=mode"]
+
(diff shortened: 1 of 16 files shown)


lib/ramble/docs/dev_guides/application_dev_guide.rst
```diff
@@ -277,6 +277,49 @@ that should be used when referencing file paths in application definitions. This
...
+^^^^^^^^^^^^^^^^^
+Conditional Logic
+^^^^^^^^^^^^^^^^^
+
+Ramble supports conditional logic to allow developers to set directives based on
+their :ref:`variants<variants-config>`. Directives supporting this logic will
+have a `when` argument to accept conditions. Multiple directives can be grouped
+under the same when clause by enclosing them in a `with` statement and using the
+`when` directive.
+
```
douglasjacobsen, line 289: It would probably be nice to have a definitive list of standard variant definitions.

i.e.

```
package_manager
package_manager_family
workflow_manager
workflow_manager_family
modifier
<mod-name>_mode
```
And then we can extend it in the future when / if we add other standard ones.

lib/ramble/ramble/cmd/common/info.py
```diff
@@ -204,6 +268,9 @@ def print_single_attribute(obj, attr, verbose=False, pattern="*", format=support
     if attr == "registered_phases":
         _print_phases(obj, attr, verbose, pattern, format=format)
         return
+    if attr == "figures_of_merit":
```
douglasjacobsen: Maybe this if should be `elif`

lib/ramble/ramble/cmd/common/info.py
```diff
@@ -193,6 +230,33 @@ def _print_phases(obj, attr, verbose=False, pattern="*", format=supported_format
...
+
+    Figures of merit are stored in a nested dictionary.
+    """
+    internal_attr_name = _map_attr_name(attr)
+    internal_attr = getattr(obj, internal_attr_name)
+    print_attribute_header(attr, verbose)
+
+    indentation = " " * 4
+
+    for when_key, context_dict in internal_attr.items():
```
douglasjacobsen: It looks like `when_key` isn't used. Maybe it can just be `_`

lib/ramble/ramble/application.py
```diff
@@ -2216,88 +2221,111 @@ def _analysis_dicts(self, criteria_list):
...
         for mod in self._modifier_instances:
-            fom_contexts.update(mod.figure_of_merit_contexts)
+            fom_sources.append(mod)
+        if self.workflow_manager is not None:
+            fom_sources.append(self.workflow_manager)
+
+        all_contexts = {}
+        for source in fom_sources:
+            for when_fs, source_context_defs in source.figure_of_merit_contexts.items():
+                if self.expander.satisfies(list(when_fs), variant_set=self.object_variants):
```
douglasjacobsen: It might be good to push the conversion of `when_fs` -> `when_list` into the `satisfies` method, just to make it a bit more clean.

I was also thinking about pushing `str` -> `[str]` into that so you could call it on a single string and not have it break.

lib/ramble/ramble/application.py
```diff
@@ -2216,88 +2221,111 @@ def _analysis_dicts(self, criteria_list):
...
 
-            mod_vars = mod.modded_variables(self)
+                for context_fs, source_foms in source_contexts.items():
+                    if not context_fs:  # FOMs with no defined context are set to the null context
+                        context_fs = frozenset([_NULL_CONTEXT])
+                        all_contexts[_NULL_CONTEXT] = {}
+                    for context in context_fs:
+                        if context not in all_contexts:
+                            fom_list = str(list(source_foms.keys()))
+                            logger.die(
```
douglasjacobsen, line 2252: We should probably have a test that hits this if possible.

lib/ramble/ramble/application.py
```diff
@@ -2216,88 +2221,111 @@ def _analysis_dicts(self, criteria_list):
...
+                            definitions[context] = {
+                                "definition": {},
+                                "foms": {},
+                            }
+                            if context != _NULL_CONTEXT:
+                                regex_str = self.expander.expand_var(
+                                    all_contexts[context]["regex"]
+                                )
+                                definitions[context]["definition"] = {
+                                    "regex": re.compile(r"%s" % regex_str),
```
douglasjacobsen: Can we replace `r"%s" % regex_str` with just `regex_str` now? (Since we're editing this code anyway)

@linsword13 is there any reason this shouldn't work?

lib/ramble/ramble/application.py
```diff
@@ -2216,88 +2221,111 @@ def _analysis_dicts(self, criteria_list):
...
+                                    "format": all_contexts[context]["output_format"],
+                                }
+
+                        for fom, source_def in source_foms.items():
+                            if fom in definitions[context]["foms"]:
+                                logger.warn(
```
douglasjacobsen, line 2275: Probably should have a test for this as well.

douglasjacobsen (changes requested):
Thanks for this! It's awesome to see the progress on it so far, and great that it's fixing so many issues we've had. :)

lib/ramble/docs/dev_guides/application_dev_guide.rst
dapomeroy: Revised verbiage

douglasjacobsen (approved):

PR #1025, opened 2025-05-12
Add license_names directive

rfbgo:
This allows licenses.yaml to inherit from it's parent if needed

If no licesne_name is specified, app_inst.name is used to be backwards compatible

Testing for this case is covered in existing tests Testing for the new inherited behavior is added in new tests

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -134,6 +134,8 @@ class ApplicationBase(metaclass=ApplicationMeta):
     maintainers: List[str] = []
     tags: List[str] = []
 
+    license_names: List[str] = []
+
     license_inc_name = "license.inc"
 
     def __init__(self, file_path):
@@ -178,6 +180,9 @@ class ApplicationBase(metaclass=ApplicationMeta):
         self._input_lock = None
         self._software_lock = None
         self._experiment_graph = None
+
+        if not self.license_names:
+            self.license_names.append(self.name)
 
         self.hash_inventory = {
             "application_definition": None,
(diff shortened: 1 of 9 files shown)


lib/ramble/ramble/application.py
```diff
@@ -179,6 +181,9 @@ def __init__(self, file_path):
         self._software_lock = None
         self._experiment_graph = None
 
+        if not self.license_names:
+            self.license_names = [self.name]
```
linsword13: My read may be off, but does it mean that if an app doesn't have the `license_name` directive, then it won't inherit the `license_names` from its parent? While an app that has `license_name` would (inherit from its base)?

rfbgo: This might get a bit confusing, but I think the short answer is that it will get the parent because of this line: https://github.com/GoogleCloudPlatform/ramble/pull/1025/files#diff-b0c6c8cf908894aea4d445b9ea2d810d02449be0eaf77981854c3533d73adb6cR325

Basically there are ~4 possible situations:

1. You parent has `license_name` and you do not => you get your parents 
2. Neither you parent, nor you, have it => you get your own name (same as today) 
3. You parent does not have `license_name` and you do => you get what you ask for (which can explicitly be your parent)
4. You parent has `license_name` and so do you: you get both, which gets downselect to "last" => only yours

I think you're asking about #1 and the directive gets set before the init 

I think the most annoying case is if in #1 your parent does something you do not want.. so the expectation is your either shadow it or remove it from the license.yaml

(and for fun: in the multiple inheritance case you take your "right most" parent! 😂 )

douglasjacobsen: I think your (2) is the main thing @linsword13  is talking about.

But, I think your essentially saying this is true, and by design. And the fix is to make sure at least one of the applications in the hierarchy has a `license_name` directive.

It might be good to have tests that cover these various CUJs to make sure they work the way we want. Or at least make sure the ones we think are important are covered.

It looks like the tests currently exercise (4) and maybe arguably (3). It'd be good to make sure (1) at least is exercised as well. (2) feels more edge and like we would expect that this wouldn't work to acquire your parent's name as a licensed app name.

rfbgo: > You parent has license_name and you do not => you get your parents

Last night I realized this means that you will not pick up licenses with your own name, so maybe the concrete instance should always have `self.name` in the `self.license_names`?

douglasjacobsen (dismissed):

douglasjacobsen (approved):

PR #1027, opened 2025-05-13
Create a base py-nemo application for py-nemo and py-nemo-2 applications

duncanspani:
This change is backwards compatible without any application changes for py-nemo or py-nemo-2.

Both py-nemo and py-nemo-2 application have been tested and are verified working.


.gitignore
```diff
@@ -3,3 +3,5 @@ __pycache__/
 *.txt
 var/ramble/cache
 var/ramble/workspaces
+*.vscode
```
douglasjacobsen: This change, and the one below it arguably shouldn't be in this PR.

Given you can configure this on your own system, I'm reluctant to merge these into ramble.

douglasjacobsen (changes requested):

.gitignore
duncanspani: Done.

douglasjacobsen (approved):

PR #1028, opened 2025-05-13
Add `when` support to formatted executables

douglasjacobsen:
This merge adds support for the `when` arugment to the formatted_executable directive. This does not apply to defining formatted_executables within a config section however.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -1192,7 +1192,7 @@ class ApplicationBase(metaclass=ApplicationMeta):
         self.variables[self.keywords.unformatted_command_without_logs] = "\n".join(
             self._command_list_without_logs
         )
-        formatted_exec_groups = [self._formatted_executables]
+        formatted_exec_groups = [{frozenset(): self._formatted_executables}]
 
         objs_to_extract = [self, self.workflow_manager, self.package_manager]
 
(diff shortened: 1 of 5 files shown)


douglasjacobsen:
I'll rebase this after your FOM PR is mergerd, and then mark it as not a draft.

lib/ramble/ramble/application.py
```diff
@@ -418,7 +418,7 @@ def set_tags(self, tags):
 
     def set_formatted_executables(self, formatted_executables):
         """Set formatted executables for this instance"""
-        self._formatted_executables = formatted_executables.copy()
+        self._formatted_executables[frozenset()] = formatted_executables.copy()
```
dapomeroy: Should this check if `self._formatted_executables` has an existing when set? Would this have the effect of removing the 'when' conditions for any copy?

douglasjacobsen: No, it's not possible for the `self._formatted_executables` to have existing when sets. These are the ones defined from the input YAML rather than those defined in directives inside objects.

Regardless, I had this `frozenset()` reference in the wrong place. I fixed that in the most recent version.

dapomeroy: Got it!

dapomeroy (dismissed):
LGTM

dapomeroy (approved):

PR #1034, opened 2025-05-16
Allow `when` on all variable types

douglasjacobsen:
This merge enables support for `when` conditions on all variable types (`modifier_variable`,`package_manager_variable`, `workflow_manager_variable`, and `workload_variable`). Additionally, it introduces a new shared_language directive called `variable` that can define generic variables for each object type.

The variable definitions using the object specific interface for modifiers, package managers, and workflow managers simply wraps the generic variable directive. Workload variables are handled a little differently, as they still need to be connected nicely to workload groups.

There are additional variants defined for `application_name` and `workload_name`.

### DIFF
--- a/lib/ramble/ramble/language/application_language.py
+++ b/lib/ramble/ramble/language/application_language.py
@@ -204,6 +204,7 @@ def workload_variable(
     workload_group=None,
     expandable: bool = True,
     track_used: bool = True,
+    when=None,
     **kwargs,
 ):
     """Define a new variable to be used in experiments
@@ -226,6 +227,7 @@ def workload_variable(
         track_used (bool): True if the variable should be tracked as used,
                            False if not. Can help with allowing lists without vecotizing
                            experiments.
+        when (list | None): List of when conditions to apply to directive
     """
 
     def _execute_workload_variable(app):
@@ -234,12 +236,17 @@ def workload_variable(
             workload, workloads, app.workloads, "workload", "workloads", "workload_variable"
         )
 
+        when_list = ramble.language.language_helpers.build_when_list(
+            when, app, name, "workload_variable"
+        )
+
         workload_var = ramble.workload.WorkloadVariable(
             name,
             default=default,
             description=description,
             values=values,
             expandable=expandable,
+            when=when_list,
             **kwargs,
         )
 
--- a/lib/ramble/ramble/language/application_language.py
+++ b/lib/ramble/ramble/language/application_language.py
@@ -225,7 +225,7 @@ def workload_variable(
         workload_group (str): Name of workload group this variable is used in
         expandable (bool): True if the variable should be expanded, False if not.
         track_used (bool): True if the variable should be tracked as used,
-                           False if not. Can help with allowing lists without vecotizing
+                           False if not. Can help with allowing lists without vectorizing
                            experiments.
         when (list | None): List of when conditions to apply to directive
     """
--- a/lib/ramble/ramble/language/package_manager_language.py
+++ b/lib/ramble/ramble/language/package_manager_language.py
@@ -21,13 +21,14 @@ class PackageManagerMeta(ramble.language.shared_language.SharedMeta):
 package_manager_directive = PackageManagerMeta.directive
 
 
-@package_manager_directive("package_manager_variables")
+@package_manager_directive(dicts=())
 def package_manager_variable(
     name: str,
     default,
     description: str,
     values: Optional[list] = None,
     expandable: bool = True,
+    when=None,
     **kwargs,
 ):
     """Define a variable for this package manager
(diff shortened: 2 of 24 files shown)


lib/ramble/ramble/variants.py
```diff
@@ -119,7 +119,10 @@ def experiment_variant(self, name: str, value: Any):
             value: The value the variant should take.
         """
         if name in self.default_variants:
-            if value not in self.default_variants[name].values:
+            if (
+                self.default_variants[name].values
```
rfbgo, line 123: What are the conditions where the `not self.default_variants[name].values` bit can be true?

lib/ramble/ramble/language/package_manager_language.py
```diff
@@ -38,18 +40,23 @@ def package_manager_variable(
         description (str): Description of variable's purpose
         values (list): Optional list of suggested values for this variable
         expandable (bool): True if the variable should be expanded, False if not.
+        track_used (bool): True if the variable should be tracked as used,
+                           False if not. Can help with allowing lists without vecotizing
```
dapomeroy: Typo?

lib/ramble/ramble/language/shared_language.py
```diff
@@ -747,6 +747,58 @@ def _define_validator(obj):
...
+    """Define a variable for this modifier
+
+    Args:
+        name (str): Name of variable to define
+        default: Default value of variable definition
+        description (str): Description of variable's purpose
+        values (list): Optional list of suggested values for this variable
+        expandable (bool): True if the variable should be expanded, False if not.
+        track_used (bool): True if the variable should be tracked as used,
+                           False if not. Can help with allowing lists without vecotizing
```
dapomeroy: Typo?

lib/ramble/ramble/language/workflow_manager_language.py
```diff
@@ -33,17 +36,24 @@ def workflow_manager_variable(
         default: Default value if the variable is not defined
         description: Description of the variable
         values: Optional list of suggested values for this variable
+        expandable (bool): True if the variable should be expanded, False if not.
+        track_used (bool): True if the variable should be tracked as used,
+                           False if not. Can help with allowing lists without vecotizing
```
dapomeroy: Typo?

lib/ramble/ramble/language/application_language.py
```diff
@@ -226,6 +227,7 @@ def workload_variable(
         track_used (bool): True if the variable should be tracked as used,
                            False if not. Can help with allowing lists without vecotizing
```
dapomeroy: Typo?

lib/ramble/ramble/workload.py
```diff
@@ -33,15 +34,15 @@ def __init__(
             description (str): Description of variable
             values: List of suggested values for variable
             expandable (bool): True if variable can be expanded, False otherwise
-            track_used (bool): True if variable should be considered used,
-                            False to ignore it for vectorizing experiments
+            when (list | None): List of when conditions to apply to directive
```
dapomeroy, line 39: Looks like the docstring for track_used was accidentally removed

var/ramble/repos/builtin/applications/intel-mlc/application.py
```diff
@@ -186,7 +186,9 @@ def _generate_input_file(self, workspace, app_inst=None):
 
         thread_dist = app_inst.expander.expand_var_name("thread_distribution")
         if thread_dist == "{thread_distribution}":
-            thread_dist = workload.variables["thread_distribution"].default
+            possible_vars = workload.find_variable("thread_distribution")
```
dapomeroy, line 189: Given that this change requires an update to one of the builtin repo applications, is it going to be a breaking change for users who don't update their custom repo apps? Anything we need to do to mitigate?

dapomeroy (commented):
LGTM aside from nits

lib/ramble/ramble/variants.py
douglasjacobsen: So, this check is essentially someone defined a variant in an object file with explicit values:

```
variant(
  "foo",
  values=["bar", "baz"],
  default="bar",
  description="..."
)
```

And then a user is trying to set this variant in YAML, with an invalid value:

```
ramble:
  variants:
    foo: blah
```

Variants defined in objects are called "default variants" while variants defined in YAML are called "experiment variants".

rfbgo: I guess my question was more like does `self.default_variants[name].values` always have a value? I was trying to understand when that is not true 

(I understand the `value not in values` bit)

douglasjacobsen: No, you could define a variant without specifying it's values.

The default for `.values` is None.

var/ramble/repos/builtin/applications/intel-mlc/application.py
douglasjacobsen: This is a great point. However, there's not really any nice ways to mitigate this. So, I consider this a pretty major breaking change.

To be fair, reaching into the internals of the applications for this purpose is kind of atypical and I don't see any use in any of the other main repositories that I know about.

dapomeroy (approved):

PR #1052, opened 2025-05-30
Fix success criteria being overwritten by subsequently defined success criteria

douglasjacobsen:
This merge fixes an issue where workspaces that contained multiple applications / experiments and each had their own scoped success criteria would evaluate success criteria incorrectly.

The original goal was that each experiment would have its own success criteria that was scope for just that experiment. This would allow disjoint criteria to be defined for each experiment and all them to be evaluated correctly. Previously, the workspace owned the success criteria, and as a result only the last experiment in the YAML would have their success criteria defined in the list. This would cause some experiments to erroneously fail as a result of using success criteria that are invalid for their definitions.

The fix was to push success criteria into the context object, and allow the experiments to own their own success criteria instead of storing this information in the workspace.

### DIFF
--- a/lib/ramble/ramble/test/success_criteria_functionality/success_scoping.py
+++ b/lib/ramble/ramble/test/success_criteria_functionality/success_scoping.py
@@ -63,9 +63,18 @@ ramble:
 
         workspace("setup", global_args=["-w", workspace_name])
         ramble_on(global_args=["-w", workspace_name])
-        workspace("analyze", global_args=["-w", workspace_name])
+        workspace(
+            "analyze", "--where", "{experiment_index} == 1", global_args=["-w", workspace_name]
+        )
 
         with open(os.path.join(ws.root, "results.latest.txt")) as f:
             data = f.read()
-            assert "FAILED" in data
-            assert "SUCCESS" in data
+            assert "Status = SUCCESS" in data
+
+        workspace(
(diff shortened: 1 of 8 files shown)


rfbgo (dismissed):

samskillman:
This worked for me!

lib/ramble/ramble/test/success_criteria_functionality/success_scoping.py
```diff
@@ -0,0 +1,71 @@
...
+            f.write(test_config)
+        ws._re_read()
+
+        workspace("setup", global_args=["-w", workspace_name])
+        ramble_on(global_args=["-w", workspace_name])
+        workspace("analyze", global_args=["-w", workspace_name])
+
+        with open(os.path.join(ws.root, "results.latest.txt")) as f:
+            data = f.read()
+            assert "FAILED" in data
```
linsword13: nit: is it possible to make the assertions more specific, so that they tie to individual experiment?

douglasjacobsen: Updated. LMK if you like that more.

linsword13: Yes, very nice!

linsword13 (approved):

PR #1055, opened 2025-06-02
Move base object classes to object repository

douglasjacobsen:
This merge migrates the base classes for all objects into the object repository. The primary motivation for this is to allow base classes to be used in a more consistent manner to the existing object classes. Additionally, this allows external repositories the ability to override aspects of the base classes without having to fork all of Ramble.

### DIFF
--- a/lib/ramble/ramble/modkit.py
+++ b/lib/ramble/ramble/modkit.py
@@ -30,9 +30,5 @@ from ramble.util.logger import logger
 from ramble.util.logger import logger as tty
 from ramble.util.output_capture import OUTPUT_CAPTURE
 
-base_class_type = ramble.repository.ObjectTypes.base_classes
-ModifierBase = ramble.repository.get_obj_class("modifier-base", object_type=base_class_type)
-BasicModifier = ramble.repository.get_obj_class("basic-modifier", object_type=base_class_type)
-DisabledModifier = ramble.repository.get_obj_class(
-    "disabled-modifier", object_type=base_class_type
-)
+ModifierBase = ramble.repository.get_base_class("modifier-base")
+BasicModifier = ramble.repository.get_base_class("basic-modifier")
--- a/lib/ramble/ramble/modkit.py
+++ b/lib/ramble/ramble/modkit.py
@@ -13,10 +13,9 @@ import llnl.util.filesystem
 from llnl.util.filesystem import *
 
 import ramble.language.modifier_language
+import ramble.repository
 from ramble.language.modifier_language import *
 from ramble.language.shared_language import *
-from ramble.modifier import ModifierBase
-from ramble.modifier_types.basic import BasicModifier
 from ramble.spec import Spec
 from ramble.util.command_runner import (
     CommandRunner,
@@ -30,3 +29,10 @@ from ramble.util.file_util import get_file_path
 from ramble.util.logger import logger
 from ramble.util.logger import logger as tty
 from ramble.util.output_capture import OUTPUT_CAPTURE
+
+base_class_type = ramble.repository.ObjectTypes.base_classes
+ModifierBase = ramble.repository.get_obj_class("modifier-base", object_type=base_class_type)
+BasicModifier = ramble.repository.get_obj_class("basic-modifier", object_type=base_class_type)
+DisabledModifier = ramble.repository.get_obj_class(
+    "disabled-modifier", object_type=base_class_type
+)
--- a/lib/ramble/ramble/pipeline.py
+++ b/lib/ramble/ramble/pipeline.py
@@ -19,9 +19,9 @@ import llnl.util.filesystem as fs
 import llnl.util.tty as tty
 from llnl.util.tty.color import cprint
 
-import ramble.application
 import ramble.config
 import ramble.expander
+import ramble.experiment_result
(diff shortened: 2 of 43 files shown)


lib/ramble/ramble/modkit.py
```diff
@@ -30,3 +29,10 @@
 from ramble.util.logger import logger
 from ramble.util.logger import logger as tty
 from ramble.util.output_capture import OUTPUT_CAPTURE
+
+base_class_type = ramble.repository.ObjectTypes.base_classes
+ModifierBase = ramble.repository.get_obj_class("modifier-base", object_type=base_class_type)
+BasicModifier = ramble.repository.get_obj_class("basic-modifier", object_type=base_class_type)
+DisabledModifier = ramble.repository.get_obj_class(
```
linsword13: Do we need this defined here?

lib/ramble/ramble/software_environments.py
```diff
@@ -9,11 +9,16 @@
...
 
 import ramble.error
+import ramble.repository
 import ramble.util.colors as rucolor
 from ramble.namespace import namespace
-from ramble.package_manager import PackageManagerBase
 from ramble.util.logger import logger
 
+base_class_type = ramble.repository.ObjectTypes.base_classes
+PackageManagerBase = ramble.repository.get_obj_class(
```
linsword13, line 18: nit: I wonder if we want to have a `get_base_class` defined in `ramble.repository`, just for convenience?

lib/ramble/ramble/software_environments.py
```diff
@@ -148,7 +153,7 @@ def __init__(
         Args:
             name (str): Name of package
             pkg_info (dict): Package info containing specs for supported package managers
-            package_manager (ramble.package_manager.PackageManagerBase):
+            package_manager (ramble.base_cls.builtin.PackageManagerBase):
```
linsword13: nit: I wonder if there's a slight difference between the comment here and the dynamic nature of the import, in that it doesn't necessarily come from the builtin repo?

lib/ramble/ramble/pipeline.py
```diff
@@ -369,7 +374,7 @@ def _prepare(self):
 
         excluded_secrets = set()
         if not self.include_secrets:
-            excluded_secrets.add(ramble.application.ApplicationBase.license_inc_name)
+            excluded_secrets.add(ApplicationBase.license_inc_name)
```
linsword13, line 377: Should this explicitly use the builtin namespace? Otherwise a custom application base class may inadvertently break this?

lib/ramble/ramble/modkit.py
douglasjacobsen: Probably not. I just added all of the base classes to have references here. But I removed this one.

lib/ramble/ramble/software_environments.py
douglasjacobsen: Good suggestion, I updated this.

lib/ramble/ramble/software_environments.py
douglasjacobsen: I tried to update this. I don't really have any great suggestions, especially given we write our internal stuff kind of assuming people will follow the base classes in the builtin repo, but they definitely don't have to.

lib/ramble/ramble/pipeline.py
douglasjacobsen: I don't really want it to use the builtin namespace explicitly. IMO if someone breaks it, that's because they did something weird with one of the base classes without understanding what it should have done. But in this case, they could still potentially change it and have it work fine.

linsword13: That makes sense. What do you think if we decouple the `license_inc_name` from application, and just put it in some where like `util/constants.py`?

lib/ramble/ramble/software_environments.py
linsword13: I see, sounds good.

var/ramble/repos/builtin/base_classes/basic-modifier/base_class.py
```diff
@@ -7,7 +7,7 @@
 # except according to those terms.
 
 
-from ramble.modifier import ModifierBase
+from ramble.modkit import *  # noqa: F403
```
linsword13, line 10: Do you know if the `noqa` is needed?

lib/ramble/ramble/pipeline.py
douglasjacobsen: I think that's fine. I'll make a task for this, because it probably makes sense to do a more thorough review and migrate constants like this together.

var/ramble/repos/builtin/base_classes/basic-modifier/base_class.py
douglasjacobsen: Yeah, without the noqa lines that import these kits using `*` fail the style checks.

linsword13: Hm, good to know. I probably need to look into style cmd. There's supposed to have some `pattern_exemptions` on such imports.

linsword13 (approved):

rfbgo (approved):

PR #1059, opened 2025-06-03
Add LL128 mode to nccl-tcpxo modifier

douglasjacobsen:


var/ramble/repos/builtin/modifiers/nccl-tcpxo/modifier.py
```diff
@@ -76,3 +80,19 @@ def source_tcpxo(self):
                 )
 
         return cmds
+
+    register_builtin(
+        "source_tcpxo", injection_method="prepend", depends_on=["source_tcpxo"]
```
linsword13: Should this be `set_ll128_env_vars`?

douglasjacobsen: Whoops! It's fixed now.

linsword13 (approved):

PR #1061, opened 2025-06-04
Update tests to use request fixture for workspace names

douglasjacobsen:
This merge changes the workspace names in tests to use the request fixture so workspace names can be based off of the test name instead of having to hardcode strings for the workspace name.

### DIFF
--- a/lib/ramble/ramble/test/application.py
+++ b/lib/ramble/ramble/test/application.py
@@ -438,7 +438,7 @@ def test_define_commands(mutable_mock_apps_repo):
     assert "mpirun" in executable_application_instance.variables["command"]
 
 
-def test_derive_variables_for_template_path(mutable_mock_apps_repo):
+def test_derive_variables_for_template_path(mutable_mock_apps_repo, workspace_name):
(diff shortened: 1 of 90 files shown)


lib/ramble/ramble/test/application.py
```diff
@@ -468,7 +470,7 @@ def test_derive_variables_for_template_path(mutable_mock_apps_repo):
 """
     import os.path
 
-    workspace_name = "test_derive_variables_for_template_path"
+    workspace_name = re.sub("[^0-9a-zA-Z_-]", "_", request.node.name)
```
rfbgo: Should we wrap the request fixture into something that sanitizes and clearly gives a workspace_name? Doing this regex outside of the function every time feels kind of bad

rfbgo (changes requested):

lib/ramble/ramble/test/application.py
douglasjacobsen: All done!

linsword13 (approved):

rfbgo (approved):

PR #1063, opened 2025-06-04
Add modifier to aggregate environment variables

douglasjacobsen:
This merge adds a new modifier which should aggregate environment variables. This aggregation allows prefixing the names of the environment variables. The primary use here is to enable workflows where the parallel runtime needs a list of all environment variables to propagate into the execution environment.

### DIFF
--- a/var/ramble/repos/builtin/modifiers/env-var-aggregator/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/env-var-aggregator/modifier.py
@@ -1,0 +1,180 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import ramble.config
+import ramble.util.executable
+from ramble.modkit import *  # noqa: F403
+from ramble.util import shell_utils
+from ramble.util.logger import logger
+
+
+class EnvVarAggregator(BasicModifier):
+    """
+    This modifier is used to prefix, and aggregate the names of environment
+    variables into a common environment variable definition.
+
+    There are two execution modes:
+     'all' - Includes all environment variables defined in the environment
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/modifiers/env-var-aggregator/modifier.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    name = "env-var-aggregator"
+
+    mode("all", description="Aggregate all environment variables")
+    mode(
+        "explicit",
+        description="Construct an aggregated list of explicit variables",
+    )
+    default_mode("all")
```
linsword13, line 42: For the "all" mode, do we want to have some mechanism so that certain env vars are skipped? Like we probably don't want to include in `HOSTNAME` (if that's set as a env-var in the particular shell)?

var/ramble/repos/builtin/modifiers/env-var-aggregator/modifier.py
```diff
@@ -0,0 +1,144 @@
...
+
+import ramble.config
+import ramble.util.executable
+from ramble.modkit import *  # noqa: F403
+from ramble.util.logger import logger
+
+
+class EnvVarAggregator(BasicModifier):
+    """
+    A modifier that aggregates existing environment variables with a specific prefix
```
linsword13: My first read of this description is that this will only aggregate env vars that start with a given prefix (like only grab ones that start with `I_MPI`,) but it seems what it actually does is to grab all the env vars, and then prepend a prefix.

var/ramble/repos/builtin/modifiers/env-var-aggregator/modifier.py
douglasjacobsen: That's a good suggestion. I just updated that. It's only for the `all` mode though.

var/ramble/repos/builtin/modifiers/env-var-aggregator/modifier.py
douglasjacobsen: I updated this. Let me know if the description is better now.

linsword13: Much better now, thanks!

linsword13 (approved):

PR #1064, opened 2025-06-04
Add `when` support to executable directive

dapomeroy:


lib/ramble/ramble/application.py
```diff
@@ -963,11 +963,21 @@ def _check_object_validators(self):
                         else:
                             logger.warn(err_msg)
 
+    def _get_filtered_executables(self) -> dict:
+        """Returns a dict of executables that satisfy `when` conditions"""
+        filtered_executables = {}
+        all_executables = self.executables.copy()
+        for when_set, executable in all_executables.items():
+            if self.expander.satisfies(when_set, variant_set=self.object_variants):
+                filtered_executables.update(executable)
```
douglasjacobsen: It might be worth throwing an error here if any of the executables are already defined. I think in this case we should consider the experiment invalid and make someone fix it.

We'd also want to test this to make sure the error is raised.

douglasjacobsen (changes requested):
Thanks, this looks good. I had one edge case that I think should be covered and tested for, but otherwise everything looks good to merge.

lib/ramble/ramble/application.py
dapomeroy: Added the error and test

douglasjacobsen (approved):

PR #1067, opened 2025-06-06
Fix order of default variable definition

douglasjacobsen:
Previously, several of the default variable types (workflow manager, package manager, etc..) were defined fairly late in the process of creating an experiment. This meant that they were not defined when experiment validation occurred, which caused workspaces that didn't define a default required variable (such as mpi_command) to error when being set up.

This merge fixes this issue by reordering the definition of variables, and fixing up the tests in places where this was caused an issue.

### DIFF
--- a/lib/ramble/ramble/application.py
+++ b/lib/ramble/ramble/application.py
@@ -227,10 +227,12 @@ class ApplicationBase(metaclass=ApplicationMeta):
         new_clone = type(self)(self._file_path)
         self.generated_experiments.append(new_clone)
 
-        if self._env_variable_sets:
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/application.py
```diff
@@ -386,6 +397,26 @@ def set_env_variable_sets(self, env_variable_sets):
 
         self._env_variable_sets = env_variable_sets.copy()
 
+        # Extract workflow environment variable sets
```
linsword13: nit:  is it "workload" instead?

douglasjacobsen: Good catch. Should be fixed now.

linsword13 (approved):

PR #1069, opened 2025-06-07
Ensure escaped curly braces are properly escaped.

douglasjacobsen:
The '\{' string is considered an invalid escape sequence. However, this is a string Ramble uses to propagate curly braces into the output of expanded strings. This merge adds supression for these warnings so they don't pollute the output of Ramble.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -21,6 +21,10 @@ from ramble.util.logger import logger
 import spack.util.naming
 
 
+def _escaped_str(in_str):
+    return str(in_str).replace(r"\{", r"\\{").replace(r"\}", r"\\}")
+
+
 def _and(a, b):
     return a and b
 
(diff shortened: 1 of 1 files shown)


lib/ramble/ramble/expander.py
```diff
@@ -760,7 +761,12 @@ def perform_math_eval(self, in_str):
 
         """
         try:
-            math_ast = ast.parse(in_str, mode="eval")
+            with warnings.catch_warnings():
+                # Filter '\{' being an invalid escape sequence warnings.
+                warnings.filterwarnings(
+                    action="ignore", message="invalid escape sequence.*{", module=".*"
```
linsword13: Somewhat annoyingly, it sounds like the plan is to eventually turn this into a `SyntaxError`: https://screenshot.googleplex.com/8kXajamrDLmorvm. So perhaps we should promote a more future-proof way of escaping the '{' for Ramble usage.

douglasjacobsen:
This was resolved with #1072

PR #1078, opened 2025-06-12
Add a front-end command for managing modifiers within a workspace

douglasjacobsen:
This merge adds a new front-end command for managing modifiers within a workspace. Supported functionality in this command includes:

 * Listing modifiers and their associated scopes
 * Adding modifiers to a specific scope
 * Removing one or more modifiers from a scope (or based on a globed name)
 
 Tests and documentation are added for this functionality as well.

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -60,7 +60,7 @@ subcommands = [
     "manage",
 ]
 
-manage_commands = ["experiments", "software", "includes"]
+manage_commands = ["experiments", "software", "includes", "modifiers"]
 
 
 def workspace_activate_setup_parser(subparser):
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1446,6 +1446,105 @@ def workspace_manage_includes(args):
...
+    """manage workspace modifiers"""
+    actions = subparser.add_mutually_exclusive_group()
+    actions.add_argument(
+        "--list", "-l", action="store_true", help="whether to print existing modifiers"
+    )
+
+    actions.add_argument(
+        "--add",
+        action="store_true",
+        help="whether to remove an existing include by index",
```
rfbgo: Is this (and below) definitely includes?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1446,6 +1446,105 @@ def workspace_manage_includes(args):
...
+        "Should be the name of a modifier object.",
+    )
+
+    subparser.add_argument(
+        "--mode",
+        "-m",
+        dest="mode",
+        metavar="MODE",
+        default=None,
+        help="name of modifier to manage. Can be a glob when removing, and defaults to '*'. "
```
rfbgo: s/name/mode?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1446,6 +1446,105 @@ def workspace_manage_includes(args):
...
+                mode_pattern=args.mode,
+            )
+    elif args.add:
+        with ws.write_transaction():
+            ws.add_modifier(
+                scope=args.scope,
+                name_pattern=args.name,
+                mode=args.mode,
+                on_executable=args.on_executable,
+            )
```
rfbgo, line 1563: Do you think we should have something that tries to tell (or confirms with) the user that we did something? 

`Added 2 modifiers to 1 experiment`?

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Good catch! Fixed.

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Good catch! Fixed.

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Thanks, that's a good idea.

I also realized I had forgot adding the dry-run option. So I updated this PR with that functionality as well.

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1446,6 +1446,134 @@ def workspace_manage_includes(args):
...
+
+    ws = ramble.cmd.require_active_workspace(cmd_name="workspace manage modifiers")
+
+    if args.list:
+        with ws.read_transaction():
+            ws.print_modifiers()
+    elif args.remove:
+        remove_index = None
+        if args.remove_index:
+            remove_index = int(args.remove_index)
```
rfbgo: This cast feels a little dangerous. Can we have argparse do the validation with something like `type=int` ?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1446,6 +1446,134 @@ def workspace_manage_includes(args):
...
+        default=None,
+        help="mode of modifier to manage. Can be a glob when removing, and defaults to None, "
+        "using the default for the modifier. ",
+    )
+
+    subparser.add_argument(
+        "--on-executable",
+        "-e",
+        dest="on_executable",
+        metavar="EXEC_LIST",
```
rfbgo, line 1513: Do you think something like `nargs='*'` or `nargs='+'` would be more appropriate for the list stuff here?

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1990,6 +1992,253 @@ def workspace_paths(self):
 
         return self._workspace_path_replacements
 
+    def _get_scope_section(self, scope):
+        base_section = None
+        if scope == "workspace":
+            base_section = self.config_sections["workspace"]["yaml"][namespace.ramble]
```
rfbgo, line 1998: nit: Is there a nice way to dry out the magic string `workspace` ?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1446,6 +1446,134 @@ def workspace_manage_includes(args):
...
 
+def workspace_manage_modifiers_setup_parser(subparser):
+    """manage workspace modifiers"""
+    actions = subparser.add_mutually_exclusive_group()
+    actions.add_argument(
+        "--list", "-l", action="store_true", help="whether to print existing modifiers"
+    )
+
+    actions.add_argument(
+        "--add",
```
rfbgo, line 1457: what happens if none of add, remove or list are passed?

rfbgo (changes requested):

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Yes, good suggestion. I've made that change.

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: I think this syntax is more consistent with the rest of our manage commands. I also think `'*'` has some issue in how it parses relative to other arguments. `'+'` might be good, but if we're going to make this change we should do it in the other manage commands as well.

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: I could, but I don't know that it really buys us much here. Really the "workspace" in config_sections means something different from the one on line 1997.

I don't think making them the same dry-ed out string is a good idea. Arguably, there are more references to `"workspace"` in here that just in these lines, so I'll make a task for someone to go in and dry them out.

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Nothing at all.

I could always make it default to `list` though. Either that, or trying to print help info, but I don't know that we have a nice hook in the commands to print help info.

rfbgo: We can make arg parse request one of the three is set by using a mutual exclusive group.. would that be useful? 

Might also be useful to think about what should happen is someone asks for both in the same invocation

douglasjacobsen: This is already using a mutually exclusive group. Maybe I need to add something to say that the default is list.

rfbgo (approved):

PR #1081, opened 2025-06-13
Add `when` support to workload directive

dapomeroy:


lib/ramble/ramble/test/when.py
```diff
@@ -1052,3 +1052,138 @@ def test_input_when(request, input_when, expected_input_file):
...
+            "execute_experiment",
+        )
+
+        with open(exec_file) as f:
+            data = f.read()
+
+            assert exec_test_str[expected_exec] in data
+
+
+def test_workload_errors_when_not_enabled(request):
```
douglasjacobsen, line 1109: Odd question, but what happens if you run info on this workspace?

douglasjacobsen (commented):
This looks good, thanks for working on this. I had one question, which might lead to a follow on request, but we'll see. :)

lib/ramble/ramble/test/when.py
dapomeroy: When `ramble:variants:workload_enabled_when` is true, it works as expected:

```
$ ramble --mock-applications workspace info
Workspace: when_directives

Location: /ramble/var/ramble/workspaces/when_directives

Experiments:
  Application: when-directives
    Workload: test_workload_enabled_when
      Experiment 1: when-directives.test_workload_enabled_when.generated
```

When the variant is false, it errors on `when` conditions:

```
$ ramble --mock-applications workspace info
Workspace: when_directives

Location: /ramble/var/ramble/workspaces/when_directives
==> Error: Workload test_workload_enabled_when is not defined for the active `when` conditions.
```

Does that seem correct to you?

douglasjacobsen: Yes! Thanks. :)

douglasjacobsen (approved):

PR #1085, opened 2025-06-17
Add warning capture to ast.parse calls

douglasjacobsen:
This merge adds logic to catch warnings and verify they are warning types we care about when `ast.parse` is called.

Currently, strings that make it into the math evaluation logic with escaped braces (i.e. '\{' or '\}') cause a SyntaxWarning (that will later become a SyntaxError). This merge updates this logic so the warnings are caught and examined before they are printed. Additionally, a generic `except Execption` block is added to help ensure the future SyntaxError change doesn't break the expander.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -12,6 +12,7 @@ import operator
 import random
 import re
 import string
+import warnings
 from contextlib import contextmanager
 from typing import Dict, FrozenSet, List, Union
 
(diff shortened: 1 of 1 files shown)


lib/ramble/ramble/expander.py
```diff
@@ -794,7 +795,7 @@ def _partial_expand(
 
         return str(in_str)
 
-    def perform_math_eval(self, in_str):
+    def perform_math_eval(self, in_str, allow_retry=True):
```
rfbgo: `allow_retry` is never set? Do we need it?

lib/ramble/ramble/expander.py
```diff
@@ -805,15 +806,24 @@ def perform_math_eval(self, in_str):
...
+            except MathEvaluationError as e:
+                logger.debug(f'   Math input is: "{in_str}"')
+                logger.debug(e)
+            except RambleSyntaxError as e:
+                raise RambleSyntaxError(f'{str(e)} in "{in_str}"')
+            except Exception as e:
+                logger.debug(f"ast.parse hit the following exception on input: {in_str}")
+                logger.debug(f"{e}")
+
+            if isinstance(wal, list) and len(wal) > 0:
```
rfbgo: Isn't `wal` guarnteed to be a list?

> If record is [True](https://docs.python.org/3/library/constants.html#True), a list is returned that is progressively populated with objects as seen by a custom..

If so we can probably drop the `if` and loop over it?

rfbgo (dismissed):

lib/ramble/ramble/expander.py
douglasjacobsen: Yes, sorry, this was me forgetting to remove something. :/

lib/ramble/ramble/expander.py
douglasjacobsen: Yup! Done.

rfbgo (approved):

PR #1088, opened 2025-06-20
Improve performance of variants classes

douglasjacobsen:
This merge updates some of the logic of the variants classes to improve performance.

This includes caching the definition of each variant (so we do not have to rebuild the strings multiple times), along with using sets instead of lists in places where order is not important.

### DIFF
--- a/lib/ramble/ramble/variants.py
+++ b/lib/ramble/ramble/variants.py
@@ -256,10 +256,8 @@ class Variant:
         self.description = description
         self.values = values
         if isinstance(self.default, bool):
-            if self.default:
-                self._definition = f"+{self.name}"
-            else:
-                self._definition = f"~{self.name}"
+            prefix = "+" if self.default else "~"
+            self._definition = f"{prefix}{self.name}"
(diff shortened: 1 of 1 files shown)


lib/ramble/ramble/variants.py
```diff
@@ -255,6 +255,13 @@ def __init__(
         self.default = default
         self.description = description
         self.values = values
+        if isinstance(self.default, bool):
+            if self.default:
+                self._definition = f"+{self.name}"
+            else:
+                self._definition = f"~{self.name}"
```
rfbgo: nit: do you think this would be cleaner more like:

```
       prefix = "+" if self.default else "~"
       self._definition = f"{prefix}{self.name}"
```

rfbgo (dismissed):

lib/ramble/ramble/variants.py
douglasjacobsen: Updated!

rfbgo (approved):

PR #1090, opened 2025-06-23
Add `when` support to env_var_modification directive

dapomeroy:


lib/ramble/ramble/language/language_helpers.py
```diff
@@ -130,6 +130,56 @@ def require_definition(
     )
 
 
+def require_when_or_mode(
```
dapomeroy: This is a mod-specific version of require_definition() that merges mode/modes and when, allowing any of the three to be used. I thought about whether it should be in modifier_language or language_helpers, and lean towards language_helpers. My reasoning was because of the similarity with require_definition and not having helper functions within modifier_language. But I'm open to moving it if others think it fits better in that file. What do you think?

douglasjacobsen: I wonder if we can generalize this a bit here. I think it's fine to be here, but maybe we can remove the mode / modes arguments, allow kwargs, and then take two arguments that are the names of the singular and plural arguments that we need to merge?

Then we could potentially do the same thing for workload and workloads too.

lib/ramble/ramble/test/when.py
```diff
@@ -1257,3 +1257,61 @@ def test_obj_env_var_when(workspace_name, obj, mutable_mock_wms_repo, mutable_mo
...
+
+
+@pytest.mark.parametrize(
+    "env_var_mod_when,expected_exec",
+    [
+        (False, False),
+        (True, True),
+    ],
+)
+def test_env_var_modification_when(request, env_var_mod_when, expected_exec):
```
douglasjacobsen: `request` -> `workspace_name`

douglasjacobsen (changes requested):

lib/ramble/ramble/language/language_helpers.py
dapomeroy: OK - generalized this with just when and mode/modes for now, and ability to add workload/workloads later

lib/ramble/ramble/language/language_helpers.py
```diff
@@ -130,6 +130,57 @@ def require_definition(
     )
 
 
+def require_condition(obj, directive_name, **kwargs):
```
douglasjacobsen: Sorry here more what I meant was that the API should be something like:
```
require_condition(
    obj, directive_name: str,
    singular_name: str=None, plural_name: str=None, **kwargs)
)
```

And then internally, you can validate that `singluar_name` and `plural_name` are in kwargs (if they are not None).

This way, we don't actually need to update it at all to support workload/workloads.

dapomeroy: OK, I think I've got it. Let me know if you see anything else needing to be tweaked. I used the same single/multiple arg names as require_definition for consistency.

douglasjacobsen (approved):

PR #1097, opened 2025-06-26
Add `when` support to variable_modification

dapomeroy:


lib/ramble/ramble/language/modifier_language.py
```diff
@@ -97,31 +98,35 @@ def variable_modification(
...
 
-        for mode_name in all_modes:
-            if mode_name not in mod.variable_modifications:
-                mod.variable_modifications[mode_name] = {}
+        if when_set not in mod.variable_modifications:
+            mod.variable_modifications[when_set] = {}
 
-            if name not in mod.variable_modifications[mode_name]:
-                mod.variable_modifications[mode_name][name] = []
+        if name not in mod.variable_modifications[when_set]:
```
rfbgo, line 118: This is a small thing, but am I right that defining the same modification with the same when should over write the first? 

If so it would be nice to have that explicitly covered by a test

lib/ramble/ramble/test/when.py
```diff
@@ -1326,3 +1326,61 @@ def test_executable_modification_when(workspace_name, exec_mod_when, expected_ex
...
+            "n_nodes=1",
+            "-v",
+            "processes_per_node=1",
+            global_args=global_args,
+        )
+
+        mod_config_path = os.path.join(ws.config_dir, "modifiers.yaml")
+        with open(mod_config_path, "w+") as f:
+            f.write("modifiers:\n")
+            f.write(" - name: when-modifier\n")
```
rfbgo, line 1365: If I'm reading the tests right we only test a pretty simple case, right?

It would be nice if we could test things like multiple whens for more complete coverage

lib/ramble/ramble/language/modifier_language.py
```diff
@@ -97,31 +98,35 @@ def variable_modification(
...
                 "variable_modification directive given an invalid method.\n"
                 f"  Valid methods are {str(supported_methods)}"
             )
 
-        all_modes = ramble.language.language_helpers.require_definition(
-            mode, modes, mod.modes, "mode", "modes", "variable_modification"
+        when_list = ramble.language.language_helpers.require_condition(
+            mod, "variable_modification", "mode", "modes", mode=mode, modes=modes, when=when
         )
+        when_set = frozenset(when_list)
```
rfbgo, line 113: What happens if the user leaves when empty? What is the right behavior then?

I think we will key to an empty list, right? Maybe it's better as a warning/error or something?

rfbgo (changes requested):

lib/ramble/ramble/language/modifier_language.py
dapomeroy: Not quite - the modifications are stored in a list, so the same modification with the same when should get added on to the end of the list, even if it duplicates. This is covered by `test_variable_modification_directive()` which used to check that 2 mods with the same mode name both get added to the list, and now checks that two mods with the same when both get added to the list.

lib/ramble/ramble/language/modifier_language.py
dapomeroy: This is covered by `ramble.language.language_helpers.require_condition()` (L110 above) which requires at least one of when, mode, or modes to be defined. There's also a test of this error under `test_variable_modification_missing_mode_or_when()`

lib/ramble/ramble/test/when.py
dapomeroy: Sure - do you have suggestions for some variations to add? Maybe a variant `when` plus a mode?

dapomeroy: I added a test mode and expanded the test to check that both mode and the bool variant are satisfied. LMK if you have other ideas to improve the coverage

rfbgo (approved):

PR #1099, opened 2025-06-27
Add `when` support to package_manager_requirement

dapomeroy:


var/ramble/repos/builtin/base_classes/modifier-base/base_class.py
```diff
@@ -266,8 +266,11 @@ def all_env_var_modifications(self):
             yield from self.env_var_modifications[when_set].items()
 
     def all_package_manager_requirements(self):
-        if self._usage_mode in self.package_manager_requirements:
-            yield from self.package_manager_requirements[self._usage_mode]
+        for when_set, reqs in self.package_manager_requirements.items():
+            if not self.expander.satisfies(when_set, self.object_variants):
+                continue
+
+            yield from self.package_manager_requirements[when_set]
```
rfbgo, line 262: Is this the same as `yield reqs`?

rfbgo (dismissed):
Overall lgtm! One silly nit/question from me

var/ramble/repos/builtin/base_classes/modifier-base/base_class.py
dapomeroy: The reqs are stored in a list. Previously the list was stored in a dict keyed on the modifier mode, and it would yield from the list using the current mode as key. Now, it's in a dict keyed on the when set (which includes the modifier mode converted to a when condition), and it yields from the list when the when conditions are satisfied. So it still matches the current mode, but also validates any other when conditions.

rfbgo:
Thanks for the reply! Happy to merge once conflicts are resolved

rfbgo (approved):

PR #1107, opened 2025-07-03
added fom latency extraction for imb pingpong

fschuerm:


var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
```diff
@@ -236,6 +236,15 @@ class IntelMpiBenchmarks(ExecutableApplication):
         contexts=["bw-bytes"],
     )
 
+    figure_of_merit(
+        "Latency",
```
douglasjacobsen: Can you change the name of this FOM to `Max Time` instead of Latency? Mostly just because we have separate tests we would use to explore latency, rather than using the BW test.

douglasjacobsen (changes requested):

var/ramble/repos/builtin/applications/intel-mpi-benchmarks/application.py
fschuerm: thanks, modified the FoM name to "Time avg"

douglasjacobsen (approved):

PR #1114, opened 2025-07-08
Set default vars in user-managed workflow manager

linsword13:
The default config template sets `workflow_manager` to `user-managed`, and the `mpi_commands` and `batch_submit` variables are set inside that workflow manager. The intent is to avoid the default config template to set these variables and potentially in conflict with workflow managers like slurm.

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -390,22 +390,26 @@ def get_workspace(args, cmd_name, required=False):
         )
 
 
-def _get_all_wm_var_names(wm):
-    """Return a set of all workflow manager's variables names."""
-    if wm is None:
+def _get_all_obj_var_names(obj, obj_type):
+    """Return a set of all variables names defined in the given object."""
+    if obj is None:
+        if obj_type == ramble.repository.ObjectTypes.package_managers:
+            variant_name = namespace.package_manager
+        elif obj_type == ramble.repository.ObjectTypes.workflow_managers:
+            variant_name = namespace.workflow_manager
+        else:
+            raise ValueError("Only package manager and workflow manager types are supported")
         variants_dict = ramble.config.get(namespace.variants)
-        wm_name = variants_dict.get(namespace.workflow_manager)
-        if wm_name is None:
+        obj_name = variants_dict.get(variant_name)
+        if obj_name is None:
             return set()
     else:
-        wm_name = wm
+        obj_name = obj
     try:
-        wm_inst = ramble.repository.get(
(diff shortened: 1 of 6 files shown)


douglasjacobsen (changes requested):
I think some change is needed here:
https://github.com/GoogleCloudPlatform/ramble/pull/1114/files#diff-5c1470c76df4db058e4d465d82ee4274689c1f4de9657f36e77f9708ce22a79cL1309-L1317

for example:
```
ramble workspace create -d test
ramble -D test workspace manage experiments hostname --wf parallel
```

Adds definitions for `mpi_command` and `batch_submit` even though those should be empty (since they are defined by the workflow manager).

The changes might actually need to be here:
https://github.com/GoogleCloudPlatform/ramble/pull/1114/files#diff-5c1470c76df4db058e4d465d82ee4274689c1f4de9657f36e77f9708ce22a79cL1325-L1327

linsword13:
> I think some change is needed here: https://github.com/GoogleCloudPlatform/ramble/pull/1114/files#diff-5c1470c76df4db058e4d465d82ee4274689c1f4de9657f36e77f9708ce22a79cL1309-L1317
> 
> for example:
> 
> ```
> ramble workspace create -d test
> ramble -D test workspace manage experiments hostname --wf parallel
> ```
> 
> Adds definitions for `mpi_command` and `batch_submit` even though those should be empty (since they are defined by the workflow manager).
> 
> The changes might actually need to be here: https://github.com/GoogleCloudPlatform/ramble/pull/1114/files#diff-5c1470c76df4db058e4d465d82ee4274689c1f4de9657f36e77f9708ce22a79cL1325-L1327

Ah that's a great catch, thanks! Updated to avoid defining vars that can come from the associated workflow manager.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -389,6 +390,25 @@ def get_workspace(args, cmd_name, required=False):
         )
 
 
+def _get_all_wm_var_names(wm):
```
douglasjacobsen: We might want this to be able to extract the variables from package managers as well. I think they are less likely to be as influential to the overall workflow as the workflow manager ones, but it feels like we should ignore any variables that come from the package manager as well.

douglasjacobsen (changes requested):

lib/ramble/ramble/workspace/workspace.py
linsword13: Indeed, good call. Updated.

douglasjacobsen (approved):

PR #1117, opened 2025-07-09
Conditionally generate the push image script

linsword13:


var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
```diff
@@ -575,11 +571,19 @@ def get_package_list(self, workspace):
         description="Tag for the container image",
     )
 
+    variant(
+        "generate_push_container_image_script",
```
douglasjacobsen: Can we name this variant something specific to spack? Maybe:
`spack_generate_push_container_image_script` ? But that is also a bit long.

douglasjacobsen (changes requested):

var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
linsword13: Good call, I changed it to `spack_push_container_image_script` (without the `generate`, as `+spack_push_container_image_script` sounds like "generate" is redundant.)

douglasjacobsen (approved):

PR #1119, opened 2025-07-09
Add option to only set env-var if unset

dapomeroy:
Adds an option to only set an environment variable if it is not already set.

### DIFF
--- a/lib/ramble/ramble/workload.py
+++ b/lib/ramble/ramble/workload.py
@@ -158,6 +158,7 @@ class WorkloadEnvironmentVariable:
         value=None,
         description: str = None,
         when=None,
+        always_set=True,
         **kwargs,
     ):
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/workload.py
```diff
@@ -169,7 +170,7 @@ def __init__(
             when (list | None): List of when conditions to apply to directive
         """
         self.name = name
-        self.value = value
+        self.value = value if always_set is True else f"${{{name}:-{value}}}"
```
linsword13, line 173: This is a nice addition. I do wonder if this is bash only though...

douglasjacobsen:
We decided we'll add this as part of a different task, and instead it'll hopefully be a method of adding / defining env-vars in general so it should be supported by the `env_vars` config section.

PR #1124, opened 2025-07-14
Extend to support in memory FOMs

linsword13:
The motiviation is to enable adding FOMs that are not directly backed by log files. One common use case can be to add such FOMs from the `_prepare_analysis` hook, without needing to first write them into a log file.

With the current impl, the inmem FOMs are resolved after the file-based ones, so in theory this can be used to define derived FOMs down the line.

### DIFF
--- a/lib/ramble/ramble/language/shared_language.py
+++ b/lib/ramble/ramble/language/shared_language.py
@@ -98,13 +98,14 @@ def figure_of_merit_context(name, regex, output_format, when=None, **kwargs):
 @shared_directive("figures_of_merit")
 def figure_of_merit(
     name,
-    fom_regex,
-    group_name,
+    fom_regex=None,
+    group_name=None,
     log_file="{log_file}",
     units="",
     contexts=None,
     fom_type: FomType = FomType.UNDEFINED,
     when=None,
+    inmem_key=None,
     **kwargs,
 ):
     """Adds a figure of merit to track for this object
@@ -122,9 +123,16 @@ def figure_of_merit(
                                    should exist in.
       fom_type (ramble.util.foms.FomType): The type of figure of merit
       when (list | None): List of when conditions to apply to directive
+      inmem_key: If supplied, this is treated as an in-memory (as ooposed to file-based)
+                 figure of merit, and its value is extracted using this key
     """
 
     def _execute_figure_of_merit(obj):
+        if inmem_key is None:
+            if fom_regex is None or group_name is None:
+                raise ramble.language.language_base.DirectiveError(
+                    "`fom_regex` and `group_name` are required for defining file-based FOM"
+                )
         when_list = ramble.language.language_helpers.build_when_list(
             when, obj, name, "figure_of_merit"
         )
@@ -146,6 +154,7 @@ def figure_of_merit(
             "fom_type": fom_type,
             "when": when_list,
             "origin_type": obj.origin_type if hasattr(obj, "origin_type") else "",
+            "inmem_key": inmem_key,
         }
 
     return _execute_figure_of_merit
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/language/shared_language.py
```diff
@@ -122,9 +123,16 @@ def figure_of_merit(
                                    should exist in.
       fom_type (ramble.util.foms.FomType): The type of figure of merit
       when (list | None): List of when conditions to apply to directive
+      inmem_key: If supplied, this is treated as an in-memory (as ooposed to file-based)
```
douglasjacobsen: I wonder if we can come up with a better name for this?

Maybe we could call it `fom_map_key` or something? And then define the in memory storage of foms as the `fom_map`?

`inmem` seems like it could be confusing from a user perspective.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -2203,6 +2200,30 @@ def _prepare_analysis(self, workspace, app_inst=None):
         """
         pass
 
+    def _extract_inmem_foms(self, inmem_fom_defs, fom_values):
```
douglasjacobsen, line 2202: Do we also add all from file FOMs to the in-memory map?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -2946,7 +2983,11 @@ def _try_expand_var_or_none(var: str, expander):
                             logger.debug("Log = %s" % log_path)
                             logger.debug("Conf = %s" % fom_def)
 
-        return files, definitions
+        return files, file_fom_defs, inmem_fom_defs
+
+    def add_inmem_fom_value(self, inmem_key, value):
```
douglasjacobsen: I don't know that I have a strong preference for this yet... mostly just a brain dump... but should the in memory fom table / map / whatever list in the results object instead of the application instance?

How does this work if we want other objects to add FOMs to the inmem area? Does the object (i.e. modifier or whatever) have to call this specifically on the appliaction instance, or do we want each object to have their own in-mem map?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
linsword13: Not currently, although `fom_values` contains all the file-based FOMs extracted. My thought was that we could grab those foms from it, when they are needed by derived FOMs.

lib/ramble/ramble/language/shared_language.py
linsword13: Good call, will update.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
linsword13: Yeah right now for a non-app object, that would need to invoke this method on its app instance.

lib/ramble/ramble/language/shared_language.py
linsword13: Updated

lib/ramble/ramble/language/shared_language.py
```diff
@@ -122,9 +123,16 @@ def figure_of_merit(
                                    should exist in.
       fom_type (ramble.util.foms.FomType): The type of figure of merit
       when (list | None): List of when conditions to apply to directive
+      fom_map_key: If supplied, this is treated as an in-memory (as ooposed to file-based)
```
douglasjacobsen: nit; `ooposed` -> `opposed`

douglasjacobsen (changes requested):
One nit, otherwise looks good!

lib/ramble/ramble/language/shared_language.py
linsword13: Ouch, thanks for the catch. Corrected.

douglasjacobsen (approved):

PR #1126, opened 2025-07-15
Regard ramble object as python packages for import

linsword13:


lib/ramble/ramble/repository.py
```diff
@@ -1532,11 +1532,9 @@ def __init__(self, fullname, repo, object_name):
         super().__init__(self.fullname, self.object_py, prepend=self._object_prepend)
 
     def is_package(self, fullname):
```
douglasjacobsen, line 1534: Can this be a staticmethod, if it only returns true?

linsword13: I think not. It's overriding the method of the inherited import class.

douglasjacobsen (approved):

PR #1128, opened 2025-07-17
Add tpu device count app

cboneti:
This adds a simple application that counts the number of TPU devices available.

### DIFF
--- a/var/ramble/repos/builtin/applications/tpu-device-count/application.py
+++ b/var/ramble/repos/builtin/applications/tpu-device-count/application.py
@@ -6,7 +6,7 @@ class TpuDeviceCount(ExecutableApplication):
 
     name = "tpu-device-count"
 
-    tags("test-app")
+    tags("tpu")
 
     parameter("device_count", default=1, description="Expected number of TPU devices")
 
@@ -43,5 +43,5 @@ class TpuDeviceCount(ExecutableApplication):
         "TPU_cores_found",
         mode="fom_comparison",
         fom_name="TPU Cores",
-        formula="{value} == {device_count}",
+        formula="{value} > 0",
     )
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/tpu-device-count/application.py
```diff
@@ -0,0 +1,45 @@
...
+    """This is an example application that runs on TPUs"""
+
+    name = "tpu-device-count"
+
+    tags("tpu")
+
+    software_spec(
+        "libtpu",
+        pkg_spec="-f https://storage.googleapis.com/jax-releases/libtpu_releases.html",
+        package_manager="pip",
```
douglasjacobsen: This should be: `when="package_manager_family=pip"`

var/ramble/repos/builtin/applications/tpu-device-count/application.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+
+    software_spec(
+        "jax",
+        pkg_spec="jax[tpu]",
+        package_manager="pip",
```
douglasjacobsen: This should be: `when="package_manager_family=pip"`

var/ramble/repos/builtin/applications/tpu-device-count/application.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+
+    required_package("jax", package_manager="pip")
```
douglasjacobsen: This should be: `when="package_manager_family=pip"` instead of `package_manager="pip"`

douglasjacobsen (changes requested):

var/ramble/repos/builtin/applications/tpu-device-count/application.py
cboneti: addressed, PTAL

var/ramble/repos/builtin/applications/tpu-device-count/application.py
cboneti: addressed, PTAL

var/ramble/repos/builtin/applications/tpu-device-count/application.py
cboneti: addressed, PTAL

douglasjacobsen (approved):

PR #1129, opened 2025-07-17
Convert families to class and fix info print issues

dapomeroy:
Converts families to class. Fixes printing of families from `info` to make it more useful for defining when clauses. Adds variants to info. Fixes duplicate prints of variable modifications from `info`. Adds object environment variables to `info`. Fixes formatting of variables and other misc info issues.

### DIFF
--- a/lib/ramble/ramble/definitions/families.py
+++ b/lib/ramble/ramble/definitions/families.py
@@ -35,7 +35,7 @@ class Families:
             self._str_indent = 0
         return self.as_str(n_indent=self._str_indent)
 
-    def as_str(self, n_indent: int = 0):
+    def as_str(self, n_indent: int = 0, **kwargs):
         """String representation of this object's families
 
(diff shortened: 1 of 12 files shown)


lib/ramble/ramble/definitions/families.py
```diff
@@ -0,0 +1,54 @@
...
+    ):
+        """Constructor for families object
+
+        Args:
+            origin_type: Base object type
+            families: List of families
+        """
+        self.family_type = f"{origin_type}_family"
+        self.families = families
+
```
douglasjacobsen, line 28: Maybe we could define `__iter__` here, and then the usage could be `for family in self.families` instead of `for family in self.families.families`.

douglasjacobsen (changes requested):

douglasjacobsen (approved):

PR #1138, opened 2025-07-21
Update variant and family definitions to fix leakage

douglasjacobsen:
This merge update the definition step for variables and variants so they are interleaved to be more correct.

### DIFF
--- a/var/ramble/repos/builtin/base_classes/application-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/application-base/base_class.py
@@ -152,10 +152,11 @@ class ApplicationBase(metaclass=ApplicationMeta):
     def __init__(self, file_path):
         super().__init__()
 
-        if getattr(self, "object_variants", None) is None:
-            self.object_variants = ramble.variants.VariantSet()
-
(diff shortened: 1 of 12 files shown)


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -550,31 +500,75 @@ def set_env_variable_sets(self, env_variable_sets):
...
+            variables (dict): Dictionary of variable definitions for this
+                             experiment.
+            variants (dict): Dictionary of variant controls for this
+                             experiment.
+            experiment_set: Reference to experiment set, for expanding
+                            referenced variables.
         """
 
         self.variables = variables
+        self.variants = variants.copy()
```
linsword13, line 518: It feels a bit mixed in that variables are copied before being passed in, while variants get copied inside this function. Would it be cleaner to unify whether the caller or callee should do the copy?

douglasjacobsen: Updated. :)

linsword13 (dismissed):

linsword13 (approved):

PR #1148, opened 2025-07-23
Add a mixin to reduce duplications among objects

linsword13:


lib/ramble/ramble/object_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.util import format
+
+
+class ObjectMixin:
```
douglasjacobsen: Could we move this to be a base_class in the builtin repo? That way it's easy to extend the mixin.

douglasjacobsen (changes requested):
LGTM With one small change. :)

lib/ramble/ramble/object_mixin.py
linsword13: Ah great point, updated.

douglasjacobsen (approved):

PR #1164, opened 2025-08-04
Update su2 definition to work using upstream deck

rfbgo:
`gcc12` change is carried over as part of another change to move everything forward to spack v1.0 non expired packages, and I can temporarily roll back here if needed

### DIFF
--- a/var/ramble/repos/builtin/applications/su2/application.py
+++ b/var/ramble/repos/builtin/applications/su2/application.py
@@ -24,9 +24,9 @@ class Su2(ExecutableApplication):
     maintainers("linsword13")
 
     with when("package_manager_family=spack"):
-        define_compiler("gcc12", pkg_spec="gcc@12.2.0")
+        define_compiler("gcc12", pkg_spec="gcc@12.5.0")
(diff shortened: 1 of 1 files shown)


linsword13 (dismissed):

var/ramble/repos/builtin/applications/su2/application.py
```diff
@@ -36,10 +36,18 @@ class Su2(ExecutableApplication):
 
     # Input deck archived from https://github.com/su2code/Tutorials/tree/d7991cb74e9e12a08463c579add6a9bf73713628/compressible_flow/Inviscid_Bump
```
linsword13: nit: this comment is redundant now?

linsword13 (commented):
Nice! Didn't know multiple inputs are already supported. One nit comment, otherwise LGTM.

var/ramble/repos/builtin/applications/su2/application.py
rfbgo: Good catch, fixed!

linsword13 (approved):

PR #1169, opened 2025-08-07
 Disable expansion of environment references

douglasjacobsen:
This merge disables expansion (by the expander class) of refernces defined in the user's environment. This includes things like `${HOME}` and `~`.


linsword13 (commented):
Might be good to have some tests to codify the updated behavior?

douglasjacobsen:
@linsword13 All done. :)

linsword13 (approved):

PR #1173, opened 2025-08-11
Fix info printing `when` for executable_modifier

dapomeroy:
Fixes an issue where `when` conditions did not print for executable_modifier using the info command

### DIFF
--- a/lib/ramble/ramble/language/modifier_language.py
+++ b/lib/ramble/ramble/language/modifier_language.py
@@ -185,7 +185,9 @@ def executable_modifier(name, when=None, **kwargs):
         if when_set not in mod.executable_modifiers:
             mod.executable_modifiers[when_set] = {}
 
-        mod.executable_modifiers[when_set][name] = name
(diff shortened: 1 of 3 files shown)


var/ramble/repos/builtin/base_classes/modifier-base/base_class.py
```diff
@@ -239,7 +239,7 @@ def apply_executable_modifiers(
         post_execs = []
         for when_set, exec_mods in self.executable_modifiers.items():
             if self.expander.satisfies(when_set, self.object_variants):
-                for exec_mod in exec_mods:
```
douglasjacobsen: Is this change actually needed? I thought the default was `.keys()` when you used `in` on a dict.

dapomeroy: It sure is. My brain defaulted to being overly explicit but no need here.

douglasjacobsen (approved):

PR #1176, opened 2025-08-13
Update intel-mlc to correct the bin path

linsword13:
With spack the binary is installed under `{intel-mlc_path}/bin`.

### DIFF
--- a/var/ramble/repos/builtin/applications/intel-mlc/application.py
+++ b/var/ramble/repos/builtin/applications/intel-mlc/application.py
@@ -28,7 +28,7 @@ class IntelMlc(ExecutableApplication):
 
     executable(
         "execute_bw",
-        "{intel-mlc_path}/{exec_name} --max_bandwidth {isa_flag} -k{cpu_list} -b{buffer_size} {additional_args}",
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/intel-mlc/application.py
```diff
@@ -28,7 +28,7 @@ class IntelMlc(ExecutableApplication):
 
     executable(
         "execute_bw",
-        "{intel-mlc_path}/{exec_name} --max_bandwidth {isa_flag} -k{cpu_list} -b{buffer_size} {additional_args}",
+        "{mlc_bin_dir}/{exec_name} --max_bandwidth {isa_flag} -k{cpu_list} -b{buffer_size} {additional_args}",
```
douglasjacobsen: Can we include the trailing `/` in the bin_dir? This way if you want to say it's in your path, you can set the bin dir to an empty string?

douglasjacobsen (changes requested):

var/ramble/repos/builtin/applications/intel-mlc/application.py
linsword13: Good point. I updated so that the variable directly points to the whole path of the binary.

douglasjacobsen (approved):

PR #1182, opened 2025-08-19
Add information about `workspace manage` into app developers guide

douglasjacobsen:


lib/ramble/docs/dev_guides/application_dev_guide.rst
```diff
@@ -405,10 +405,25 @@ definition needs to contain a complete definition of at least one workload.
...
+
+In order to have Ramble generate a workspace configuration, you can use the
+``workspace manage`` command from :ref:`workspace-manage`. For example:
+
+.. code-block::
+
+  $ ramble workspace manage experiments <app_name> --wf <workload_name> -v n_nodes=1 -v n_ranks=1
+
+
+The ``--dry-run`` option, from :ref:`workspace-setup`, can be used when setting
```
linsword13, line 418: I kinda like the previous mention of using the dry-run as a way to verify/debug the workspace configuration. The new text reads a bit like one can just use the dry-run option to set up the workspace, and expect that to be functional.

douglasjacobsen: Good point. I've updated it, let me know if that sounds better.

linsword13 (approved):

PR #1185, opened 2025-08-22
Add an optional hugepage config exec in intel-mlc

linsword13:


var/ramble/repos/builtin/applications/intel-mlc/application.py
```diff
@@ -26,6 +26,17 @@ class IntelMlc(ExecutableApplication):
             pkg_spec="intel-mlc",
         )
 
+    # This is an optional executable that can be included via `executable_injection`.
+    # mlc normally requires the presence of nr_hugepages.
+    executable(
```
douglasjacobsen, line 36: Do you think it would be better to add a variant for this, instead of requiring users to inject the executable manually/

linsword13: Indeed, that's much nicer. Updated.

douglasjacobsen (approved):

PR #1189, opened 2025-08-26
Fix invalid success_list reference in experiment logs

douglasjacobsen:
This merge fixes an issue where the experiment-logs command would reference an invalid success list attribute.

A test is also added to help ensure experiment-logs works in the future.


dapomeroy (approved):
LGTM

PR #1191, opened 2025-08-26
Fix `spack compiler info` call

douglasjacobsen:
This merge fixes an issue where we pass spack the incorrect spec for compilers. This results in spack trying to re-install compilers, even if they are already installed, making repeat calls to `ramble workspace setup` take longer than needed.

### DIFF
--- a/var/ramble/repos/builtin/package_managers/spack-lightweight/test/spack_runner.py
+++ b/var/ramble/repos/builtin/package_managers/spack-lightweight/test/spack_runner.py
@@ -384,7 +384,7 @@ packages:
                 sr.add_include_file(packages_path)
                 sr.add_include_file(compilers_path)
                 sr.generate_env_file()
-                sr.install_compiler("gcc@12.1.0")
+                sr.install_compiler("gcc@12.1.0", "gcc@12.1.0")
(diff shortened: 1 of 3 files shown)


var/ramble/repos/builtin/package_managers/spack-lightweight/test/spack_runner.py
```diff
@@ -384,7 +384,7 @@ def test_new_compiler_installs(tmpdir, capsys, request):
                 sr.add_include_file(packages_path)
                 sr.add_include_file(compilers_path)
                 sr.generate_env_file()
-                sr.install_compiler("gcc@12.1.0")
+                sr.install_compiler("gcc@12.1.0", "gcc@12.1.0")
```
linsword13: nit: perhaps useful to pass two different args, and assert that the compiler_spec is what's passed into the info call?

linsword13 (approved):

var/ramble/repos/builtin/package_managers/spack-lightweight/test/spack_runner.py
douglasjacobsen: I updated this some. LMK if you want other changes.

PR #1193, opened 2025-08-26
Tweaks to ior foms

linsword13:
* Add a variant to select/deselect per-iteration FOMs. When iteration is set to a larger number, the number of FOMs can be unwieldy.

* Update the per-iter FOM context regex, such that it distinguish between read and write modes.

### DIFF
--- a/var/ramble/repos/builtin/applications/ior/application.py
+++ b/var/ramble/repos/builtin/applications/ior/application.py
@@ -93,6 +93,13 @@ class Ior(ExecutableApplication):
         use_mpi=True,
     )
 
+    variant(
+        "include_iter_foms",
+        default=True,
+        values=[True, False],
+        description="Whether to include per iteration FOMs in analyze",
+    )
+
     # FOMS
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/ior/application.py
```diff
@@ -93,6 +93,13 @@ class Ior(ExecutableApplication):
         use_mpi=True,
     )
 
+    variant(
+        "include_iter_foms",
```
douglasjacobsen: Can you change the name of this variant to be more IOR specific? That way it is less likely to collide with other object's variants that might have a similar name.

douglasjacobsen (changes requested):

var/ramble/repos/builtin/applications/ior/application.py
linsword13: Good call, updated.

douglasjacobsen (approved):

PR #1215, opened 2025-09-18
Refactor experiment result class

douglasjacobsen:
This merge splits the init method of the experiment result class into an init (which doesn't do anything beyond creating the attributes of the result class), and a finalize method (which correctly sets the attributes). This gets around issues where experiments (namely, repeats under parent experiments) were not setup / run and are considered invalid because they didn't have a valid results object.


rfbgo (dismissed):
Thanks for this updated change. This is a nice fix

rfbgo (approved):

PR #1216, opened 2025-09-18
Add anti match to openfoam to detect solver nans

rfbgo:


var/ramble/repos/builtin/base_applications/openfoam/base_application.py
```diff
@@ -421,6 +421,13 @@ class Openfoam(ExecutableApplication):
         file="{experiment_run_dir}/log.simpleFoam",
     )
 
+    success_criteria(
+        "nan",
+        mode="string",
+        anti_match=r".*[+-=\s]nan",
```
linsword13: nit: perhaps slightly more efficient to make the wildcard matching non-greedy? `r".*?[+-=\s]nan"`

rfbgo:
```suggestion
        anti_match=r".*?[+-=\s]nan",
```
Good idea! FWIW if you do it like this with a "suggestion" block I think there is a cool "one click accept"

linsword13 (approved):

PR #1219, opened 2025-09-19
Class-ify environment variable modifications

dapomeroy:
Restructure environment variable modifications as a class to enable correct printing from info command

### DIFF
--- a/lib/ramble/ramble/cmd/common/info.py
+++ b/lib/ramble/ramble/cmd/common/info.py
@@ -343,9 +343,8 @@ def print_single_attribute(obj, attr, verbose=False, pattern="*", format=support
         #     otherwise filter it and print using the format specification
         # Otherwise, print it as a raw string.
         if isinstance(to_print, list):
-            if isinstance(next(iter(internal_attr)), dict):
-                for attr_dict in internal_attr:
-                    to_print = list(attr_dict.keys())
+            if internal_attr and isinstance(next(iter(internal_attr)), dict):
+                to_print = [key for attr_dict in internal_attr for key in attr_dict]
             _print_nonverbose_list_attr(to_print, pattern=pattern, format=format)
         else:
             color.cprint(f"    {str(to_print)}\n")
--- a/lib/ramble/ramble/definitions/variables.py
+++ b/lib/ramble/ramble/definitions/variables.py
@@ -1,0 +1,161 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from typing import List, Optional
+
+import ramble.util.colors as rucolor
+
+
+def _title_color(title: str, n_indent: int = 0):
+    """Set the appropriate color for titles based on indentation"""
+    if n_indent == 0:
+        out_str = rucolor.section_title(f"{title}")
+    elif n_indent == 4:
+        out_str = rucolor.nested_1(f"{title}")
+    elif n_indent == 8:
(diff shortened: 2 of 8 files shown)


lib/ramble/ramble/definitions/variables.py
```diff
@@ -0,0 +1,159 @@
...
+    """Class representing an environment variable modification"""
+
+    all_methods = ["set", "unset", "prepend", "append"]
+
+    def __init__(
+        self,
+        name: str,
+        modification: str,
+        method: str = "set",
+        when: list = [],
```
linsword13: Might be a good idea to avoid using mutable default value: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

lib/ramble/ramble/definitions/variables.py
```diff
@@ -0,0 +1,159 @@
...
+            keyword argument of 'separator' to define the delimiter between values.
+        """
+        self.name = name
+        self.when = when.copy() if when else []
+        self.set = {}
+        self.unset = {}
+        self.prepend = []
+        self.append = []
+
+        self.add_modification(
```
linsword13, line 69: Do we want to be more strict and report error when given `method` is not supported?

lib/ramble/ramble/definitions/variables.py
```diff
@@ -0,0 +1,159 @@
...
+        out_str = rucolor.nested_2(f"{title}")
+    elif n_indent == 12:
+        out_str = rucolor.nested_3(f"{title}")
+    elif n_indent == 16:
+        out_str = rucolor.nested_4(f"{title}")
+
+    return out_str
+
+
+class EnvironmentVariableModification:
```
linsword13: From the name of the class, I have the impression that this represents a single modification of a single env-var. But it also has a method named `add_modification`, so is the intent that multiple modifications can be included for a given instance of this class?

dapomeroy: Each instance is modification of a single env-var, and it may contain multiple modifications of that env-var. I tried to represent the way that the mods worked before. A key thing was that the 'set' method used a dict and only allowed one value for each env-var, which would be overwritten by subsequent calls to the directive: `mod.env_var_modifications[when_set][method][name] = modification`. I wanted to make sure that this class didn't let you create two different 'set' modifications. Do you think there's a better way to do that?

lib/ramble/ramble/definitions/variables.py
dapomeroy: Supported methods are checked in the env_var_modification directive as the first step there, and invalid values raise a DirectiveError. Right now the only place modifications are created is through the directive. Do you think we need a check here too just in case someone adds a call to this method that doesn't come from the directive?

lib/ramble/ramble/definitions/variables.py
dapomeroy: Good catch, thanks!

lib/ramble/ramble/definitions/variables.py
dapomeroy: Fixed the name

lib/ramble/ramble/definitions/variables.py
linsword13: Ah good to know checks are in place from the directive. I think that sounds good.

lib/ramble/ramble/cmd/common/info.py
```diff
@@ -344,8 +344,9 @@ def print_single_attribute(obj, attr, verbose=False, pattern="*", format=support
         # Otherwise, print it as a raw string.
         if isinstance(to_print, list):
             if isinstance(next(iter(internal_attr)), dict):
+                to_print = []
```
linsword13: If I understand correctly, here it may be simplified as:

```
to_print = [key for attr_dict in internal_attr for key in attr_dict]
```

Also not related to the changed code, but wouldn't the above `next` call hit an exception if `internal_attr` is an empty dict?

dapomeroy: Fixed both.

linsword13 (approved):

PR #1226, opened 2025-09-24
Remove ngspice workloads which cannot be run in batchmode

vishalbh:


var/ramble/repos/builtin/applications/ngspice/application.py
```diff
@@ -81,7 +81,16 @@ class Ngspice(ExecutableApplication):
             workloads=[normal_workload],
         )
 
-    for ann_workload in all_workloads:
+    ann_workloads = [
```
linsword13, line 84: Curious, what's the meaning of `ann`? Also for `normal_workloads`, does that mostly mean "workloads that can run in batch mode"? If so, would `batch_workloads` be a better name?

vishalbh: ann here seems to be standing for annotated. iscas85Circuits/85 contains two types of circuit file {name1}.net and {name1}_ann.net. That is where I took name from. Both the workloads are ran in batch mode only.

linsword13 (approved):

PR #1239, opened 2025-09-30
`spack-pip` Package Manager

michaelmckinsey1:
We have a use case where we need to install a spack package ([Caliper](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/caliper/package.py)) and would like to use pip to install the rest of the package dependencies. Caliper cannot be installed via pip, and is instead used by the python application through python bindings (spack builds Caliper with the bindings). `spack-pip` inherits from both the `spack` and `pip` package managers and has it's own `SpackPipRunner` which inherits from the `SpackRunner` and `PipRunner`

This package manager:

1. Creates a single python environment (by `Spack._software_create_env`)
2. Concretizes and installs spack packages
3. Installs pip packages

Right now the pip spec is specified by a class attribute
```
class App(ExecutableApplication):
   pip_spec = "mypkg[optional]"
```

Spack-installed packages are still determined by `package.py` definitions. But I believe this requires
```
    depends_on("python@3.11", type=("build", "run"))
    depends_on("py-setuptools", type="build")
    depends_on("py-wheel", type="build")
    depends_on("py-pip", type=("build", "run"))
```
in the spack package to be able to use pip

### DIFF
--- a/var/ramble/repos/builtin/package_managers/pip/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/pip/package_manager.py
@@ -152,7 +152,7 @@ class Pip(PackageManagerBase):
                         )
                         and pkg not in installed_pkgs
                     ):
-                        logger.die(
+                        logger.warn(
                             f"Package {pkg} is not installed "
                             f"in environment {env_context}, but is "
                             f"required by the {mod_inst.name} modifier "
--- a/var/ramble/repos/builtin/package_managers/spack-pip/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/spack-pip/package_manager.py
@@ -1,0 +1,325 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import os
+
+from ramble.util.hashing import hash_string
+from ramble.util.logger import logger
+from ramble.pkgmankit import *
+from ramble.pkg_man.builtin.spack import Spack
+from ramble.pkg_man.builtin.spack_lightweight import SpackRunner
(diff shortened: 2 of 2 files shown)


var/ramble/repos/builtin/package_managers/pip/package_manager.py
```diff
@@ -152,7 +152,7 @@ def _software_install(self, workspace, app_inst=None):
                         )
                         and pkg not in installed_pkgs
                     ):
-                        logger.die(
+                        logger.warn(
```
michaelmckinsey1, line 155: I need to figure out how a mechanism to designate which packages should be spack installed vs pip installed in the `ramble.yaml` to avoid this. 

Right now, this is the pip installer complaining about the `caliper` package not being installed, which makes sense why it is an error currently.

var/ramble/repos/builtin/package_managers/spack-pip/package_manager.py
```diff
@@ -0,0 +1,325 @@
...
+
+    # Deactivation: nothing for Spack view; fall back to pip deactivate if we used a venv
+    def generate_deactivate_command(self, shell="bash"):
+        cmds = []
+        if not self._spack_view_exists():
+            cmds.extend(PipRunner.generate_deactivate_command(self))
+        cmds.extend(SpackRunner.generate_deactivate_command(self, shell=shell))
+        return cmds
+
+    # TODO: This overrides the pip check which will fail, because we want spack to create the venv, otherwise spack will complain if venv already exists if pip makes it.
```
michaelmckinsey1, line 207: I disabled the pip mgr from creating the python env, but it runs this check a couple of times before spack creates a python environment. This check then fails because there is no python env. It seems unecessary though, as I'm not sure there needs to be an env to create the requirements.txt etc

var/ramble/repos/builtin/package_managers/pip/package_manager.py
douglasjacobsen: This might help. https://ramble.readthedocs.io/en/latest/configuration_files.html#package-manager-specific-packages

If not, let me know.

douglasjacobsen (commented):
Thanks for this! I'll give it a read soon.

michaelmckinsey1:
superseded by #1248

PR #1241, opened 2025-10-01
Evaluate formatted executables in correct order

dapomeroy:
Adds a FormattedExecutableGraph to reorder nested formatted executables so that they are evaluated in the correct order regardless of how they are ordered in the YAML.

### DIFF
--- a/lib/ramble/ramble/graphs.py
+++ b/lib/ramble/ramble/graphs.py
@@ -9,12 +9,15 @@
 import enum
 import graphlib
 import itertools
+import re
 from collections import defaultdict
 
 import ramble.error
+import ramble.expander
 import ramble.util.graph
 from ramble.util.logger import logger
 from ramble.util.naming import NS_SEPARATOR
+from ramble.workspace import namespace
 
 
 class AttributeGraph:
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/graphs.py
```diff
@@ -451,6 +453,42 @@ def _resolve_builtin_node(self, builtin_name):
...
+        # Define all graph nodes
+        for exec_name, exec_def in formatted_execs.items():
+            exec_node = ramble.util.graph.GraphNode(exec_name, attribute=exec_def)
+            super().add_node(exec_node)
+
+        # Search for internal dependencies and define edges
+        for exec_node in self.node_definitions.values():
+            formatted_conf = exec_node.attribute
+
+            expansion_pattern = re.compile(r"\{(\w+)\}")
```
douglasjacobsen: For this, can you use `expander.expansion_str` instead of coding how the expansion string is built again?

douglasjacobsen (changes requested):

lib/ramble/ramble/graphs.py
dapomeroy: Got it. Updated

douglasjacobsen (approved):

PR #1244, opened 2025-10-02
Reorder confidence intervals in stats

douglasjacobsen:
This merge reorders the confidence interval entries in statistical output.


dapomeroy (approved):
LGTM

PR #1248, opened 2025-10-03
Add support for multiple package-manager experiments

douglasjacobsen:
This merge performs a lot of internal tasks to enable experiments to have multiple package managers attached to them. An example implementation of this is provided in the `spack-pip` package manager, though in the future this could be abstracted into a base class to make it easier to implement for additional use cases.

Generally, more of the ownership is pushed into the package manager class for the software spec and compiler spec definitions / formatting.

Tests for the spack-pip package manager are added as well.

### DIFF
--- a/var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
@@ -67,7 +67,7 @@ class ObjectMixin:
         experiment_variants = self.experiment_variants()
         return app_inst.expander.satisfies(when_key, experiment_variants)
 
-    def experiment_variants(self, single_modifier=None):
+    def experiment_variants(self, include_modifier=None):
         """Construct a VariantSet for this experiment.
 
         Apply some merging logic to VariantSet combination, in order to
@@ -86,9 +86,9 @@ class ObjectMixin:
 
         exclude_types = [self_type]
 
-        if single_modifier is not None:
+        if include_modifier is not None:
             exclude_types.append(ObjectTypes.modifiers)
-            new_set.merge_variants(single_modifier.object_variants)
+            new_set.merge_variants(include_modifier.object_variants)
 
         for _, obj in app_inst._objects(exclude_types=exclude_types):
             new_set.merge_variants(obj.object_variants)
--- a/var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
@@ -7,6 +7,7 @@
 # except according to those terms.
 from html import escape
 
+from ramble.repository import ObjectTypes, get_base_class
 from ramble.util import format
 
 
@@ -45,9 +46,54 @@ class ObjectMixin:
             return self.app_inst
         return self
 
-    def satisfy_when(self, when_key):
+    def _get_object_type(self):
+        ApplicationBase = get_base_class("application-base")
+        PackageManagerBase = get_base_class("package-manager-base")
+        WorkflowManagerBase = get_base_class("workflow-manager-base")
+        ModifierBase = get_base_class("modifier-base")
+
+        if isinstance(self, ApplicationBase):
+            return ObjectTypes.applications
+        elif isinstance(self, PackageManagerBase):
+            return ObjectTypes.package_managers
+        elif isinstance(self, WorkflowManagerBase):
+            return ObjectTypes.workflow_managers
+        elif isinstance(self, ModifierBase):
+            return ObjectTypes.modifiers
+        return None
+
+    def satisfy_when(self, when_key, variant_set=None):
         app_inst = self._get_app_inst()
-        return app_inst.expander.satisfies(when_key, app_inst.object_variants)
+        experiment_variants = self.experiment_variants()
+        return app_inst.expander.satisfies(when_key, experiment_variants)
+
+    def experiment_variants(self, single_modifier=None):
+        """Construct a VariantSet for this experiment.
+
+        Apply some merging logic to VariantSet combination, in order to
+        provide scoped variant definitions.
+
+        Args:
+            include_modifier (ModifierBase): A single modifier to merge in to resulting set
+
+        Returns:
+            VariantSet: Merged variants for the experiment.
+        """
+        app_inst = self._get_app_inst()
+
+        self_type = self._get_object_type()
+        new_set = self.object_variants.copy()
+
+        exclude_types = [self_type]
+
+        if single_modifier is not None:
+            exclude_types.append(ObjectTypes.modifiers)
+            new_set.merge_variants(single_modifier.object_variants)
+
+        for _, obj in app_inst._objects(exclude_types=exclude_types):
+            new_set.merge_variants(obj.object_variants)
+
+        return new_set
 
     @property
     def required_variables(self):
--- a/var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
@@ -7,7 +7,7 @@
 # except according to those terms.
 from html import escape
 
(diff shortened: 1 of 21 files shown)


var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
```diff
@@ -742,6 +747,13 @@ def __init__(self, shell="bash", dry_run=False):
         )
         self.concretizer.add_default_arg("concretize")
 
+    def get_spack_python(self):
+        if self.dry_run:
+            return self.bs_python.copy()
```
linsword13: Does command_runner have a `copy` method?

var/ramble/repos/builtin/package_managers/spack-pip/package_manager.py
```diff
@@ -0,0 +1,273 @@
...
+
+from ramble.pkg_man.builtin.pip import Pip
+from ramble.pkg_man.builtin.spack import Spack
+from ramble.pkgmankit import *
+
+
+class SpackPip(PackageManagerBase):
+    """Chained spack + pip package manager
+
+    This package manager uses spack to install a spack environment, then using the python that is installed within this environment (or whichever one is available in your path) will install pip packages into a python virtual environment. Experiments will load both environments, in the order of spack then pip.
```
linsword13, line 19: Do we want something like checking that a python package spec is included in the spack env definition?

douglasjacobsen: Yeah, I can add some tests once we decide if we like this.

FWIW, you *would* have spack package specs in the environment if you only used the `pkg_spec` name and not the split `spack_pkg_spec` and `pip_pkg_spec` names.

var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
douglasjacobsen: Ah, I just copied this from the pip package manager. But i guess you used a different class for `bs_python` there.

var/ramble/repos/builtin/package_managers/spack-pip/package_manager.py
linsword13: Yeah I like the implementation. Do you plan to add tests to the current PR?

For my original question, I am more curious if we want to always ensure that the user actually specifies it such that a python would indeed be installed by spack.

douglasjacobsen: Ah sure. I thought you were asking about tests. I hadn't really planned on adding many tests... but I'm open to it.

I'm torn on the specification of packages. I can see both sides of it, and I think it might not be super easy to implement that they require a prefix. But I can look into it as well.

douglasjacobsen: OK, this should be resolved now.

michaelmckinsey1 (approved):
I have tested the new `spack-pip` package manager in a use case that successfully installs the [Caliper spack package](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/caliper/package.py) with python bindings, and correctly links against a python application built with pip (including its dependencies). This use case is also described in https://github.com/GoogleCloudPlatform/ramble/pull/1239.

I want to note that the importance of the `spack-pip` manager is that a python spack package no longer has to specify all, or any, of it's python dependencies as spack packages, and users can use the existing pip packages, which can be very convenient.

var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
```diff
@@ -45,9 +46,54 @@ def _get_app_inst(self):
             return self.app_inst
         return self
 
-    def satisfy_when(self, when_key):
+    def _get_object_type(self):
+        ApplicationBase = get_base_class("application-base")
+        PackageManagerBase = get_base_class("package-manager-base")
+        WorkflowManagerBase = get_base_class("workflow-manager-base")
+        ModifierBase = get_base_class("modifier-base")
```
linsword13: Would these somewhat hard-coded base class names be an issue for someone trying to override to use their own base classes? (Unless they reuse the same set of names in their overriding classes.)

Perhaps we can instead using the `origin_type` to figure it out?

var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
```diff
@@ -45,9 +46,54 @@ def _get_app_inst(self):
...
+        return app_inst.expander.satisfies(when_key, experiment_variants)
+
+    def experiment_variants(self, single_modifier=None):
+        """Construct a VariantSet for this experiment.
+
+        Apply some merging logic to VariantSet combination, in order to
+        provide scoped variant definitions.
+
+        Args:
+            include_modifier (ModifierBase): A single modifier to merge in to resulting set
```
linsword13, line 72: Do you prefer this name or `single_modifier`? I kinda think this is a better name?

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
```diff
@@ -99,6 +100,10 @@ def runner(self):
         # Turn `runner` into a property for delayed init
         return None
 
+    @property
+    def allow_unprefixed_specs(self):
+        return getattr(self, "_allow_unprefixed_specs", True)
```
linsword13: Perhaps the `_allow_unprefixed_specs` can be set inside `init`, and here it would just be a simple `return self._allow_unprefixed_specs`?

var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
```diff
@@ -45,9 +46,54 @@ def _get_app_inst(self):
...
+            return ObjectTypes.modifiers
+        return None
+
+    def satisfy_when(self, when_key, variant_set=None):
         app_inst = self._get_app_inst()
-        return app_inst.expander.satisfies(when_key, app_inst.object_variants)
+        experiment_variants = self.experiment_variants()
+        return app_inst.expander.satisfies(when_key, experiment_variants)
+
+    def experiment_variants(self, single_modifier=None):
```
linsword13: Curious how much caching can be done here: is it a pure function based on the inputs?

douglasjacobsen: It's kind of a pure function, but also not...

It should be a pure function, but some objects (like the application) call this multiple times while they are getting initialized. So, if we cache the objects too early it makes them wrong.

I added a naive implementation of caching that should pass the test though, so we can see how this works.

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
douglasjacobsen: Good suggestion, I'll update this.

var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
douglasjacobsen: I didn't have a strong preference. I had three names I actually tried, and it's nice that I accidentally left this artifact of one of them. I'll change to this one since you like it more.

var/ramble/repos/builtin/base_classes/object_mixin/base_class.py
douglasjacobsen: I had kind of expected that using the base classes (like this) would be safer than using the `origin_type` but I can definitely swap to the `origin_type` and then if we run into issues convert it back to this.

Basically, it's hard for them to not inherit the base application type but expect it to work. But they could change origin type to something other than application and have it still work.

linsword13 (dismissed):
Great stuff!

linsword13 (dismissed):

linsword13 (dismissed):

linsword13 (approved):

PR #1249, opened 2025-10-03
Search `experiment_run_dir` for `ramble on` executor

linsword13:
This is mostly for convenience, such that any executor templates specified inside the config directory can be invoked without explicitly specifying the full path.

### DIFF
--- a/lib/ramble/ramble/pipeline.py
+++ b/lib/ramble/ramble/pipeline.py
@@ -584,7 +584,19 @@ class ExecutePipeline(Pipeline):
             exec_name = exec_parts[0]
             exec_args = exec_parts[1:]
 
-            executor = Executable(exec_name)
+            executor = which(exec_name)
+            if executor is None:
+                # attempt searching inside the run directory
+                if not os.path.isabs(exec_name):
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/pipeline.py
```diff
@@ -584,7 +584,19 @@ def _execute(self):
             exec_name = exec_parts[0]
             exec_args = exec_parts[1:]
 
-            executor = Executable(exec_name)
+            executor = which(exec_name)
```
douglasjacobsen, line 587: Do you want to prioritze the experiment run dir's path for the executor?

If so, you can pass in the path. i.e.:

```
which(exec_name, path=os.path.join( ... ))
```

douglasjacobsen (commented):
Thanks for this. One question / suggestion.

lib/ramble/ramble/pipeline.py
linsword13: That's a good tip. I thought about it, and my feeling is I want to err on having it with lower precedence than the ones already in $PATH. If the override is desirable, I prefer it being more explicitly by specifying the path.

douglasjacobsen (approved):

PR #1258, opened 2025-10-08
Extend environment_variable directive with append and prepend

douglasjacobsen:
This merge extends the environment variable directive to support appending and prepending environment variables, instead of just setting them.

### DIFF
--- a/var/ramble/repos/builtin/base_classes/application-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/application-base/base_class.py
@@ -500,46 +500,68 @@ class ApplicationBase(ObjectMixin, metaclass=ApplicationMeta):
         for env_var in self.selected_environment_variables.values():
             action = env_var.method
             value = env_var.value
-            new_set = new_env_vars[action]
+            # Pull out base group from the new_env_vars dict, based on the env_var's action.
+            new_group = new_env_vars[action]
 
             if action == "append":
+                # If action is append, we need to pull out the append actions.
+                # Here, the value in the dict is a list (instead of a dict) to allow
+                # for different separators.
+                # We need to find the group (if it exists) with the correct separator,
+                # and add the new env-var to that group.
                 found = False
-                for group in new_set:
+                for group in new_group:
                     if (
                         "var-separator" in group
                         and group["var-separator"] == env_var.separator
                     ):
-                        new_set = group["vars"]
+                        found = True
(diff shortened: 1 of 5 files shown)


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -492,10 +492,40 @@ def set_env_variable_sets(self, env_variable_sets):
...
+            if action == "append":
+                found = False
+                for group in new_set:
+                    if (
+                        "var-separator" in group
+                        and group["var-separator"] == env_var.separator
+                    ):
+                        new_set = group["vars"]
+                        break
+                if not found:
```
dapomeroy, line 522: Should there be a `found = True` inside the loop, otherwise this will execute every time?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -492,10 +492,40 @@ def set_env_variable_sets(self, env_variable_sets):
...
+        new_env_vars = {
+            "set": {},
+            "append": [],
+            "prepend": [],
+        }
         for env_var in self.selected_environment_variables.values():
-            action = "set"
+            action = env_var.method
             value = env_var.value
+            new_set = new_env_vars[action]
```
dapomeroy: It's probably just because I'm not used to this pattern, but I find it fairly hard to follow the logic of this loop when the action dict value is assigned as a variable here. I find `new_set` a bit confusing when we're talking about `set` as one of three actions and this var being either a list or a dict, and not a set.

dapomeroy (changes requested):

var/ramble/repos/builtin/base_classes/application-base/base_class.py
douglasjacobsen: Nice catch, fixed!

var/ramble/repos/builtin/base_classes/application-base/base_class.py
douglasjacobsen: I changed the name to `new_group` and added more descriptive comments to help follow the logic. Let me know if this helps.

dapomeroy: Thanks, that's helpful!

dapomeroy (approved):
LGTM

PR #1264, opened 2025-10-10
Scope success criteria in results

douglasjacobsen:
This merge adds a scoped name to the success criteria in the results output. The primary goal here is to help users understand where a criteria is coming from.

### DIFF
--- a/lib/ramble/ramble/success_criteria.py
+++ b/lib/ramble/ramble/success_criteria.py
@@ -57,13 +57,15 @@ class ScopedCriteriaList:
                 f"Possible scopes are: {self._valid_scopes}"
             )
 
-    def add_criteria(self, scope, name, mode, *args, **kwargs):
+    def add_criteria(self, scope, name, mode, *args, owning_object=None, **kwargs):
         self.validate_scope(scope)
         exists = self.find_criteria(name)
         if exists:
             logger.die(f"Success criteria {name} is not unique.")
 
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/success_criteria.py
```diff
@@ -89,7 +91,8 @@ def passed(self):
 
     def all_criteria(self):
         for scope in self._valid_scopes:
-            yield from self.criteria[scope]
+            for criteria in self.criteria[scope]:
+                yield criteria, scope
```
rfbgo, line 95: I don't actually mind this, but it seems like 90% of the calls here use the "old" interface and do not care about the scope we return. Do you think it's best to have a single condensed interface or two more explicit ones? 

I'm guessing the former, but wanted to check before merging

rfbgo (approved):

lib/ramble/ramble/success_criteria.py
douglasjacobsen: I don't really care much... but I also agree with you on the "most calls use the old interface". I figured adding the scope as an output here was more along the lines of DRY. I could also push the scope down into the criteria object instead, and then make it easier to extract.

rfbgo: I'm not too worried either, just wanted to check in. If it's an intentional choice, then I'm happy

PR #1265, opened 2025-10-10
Clean up duplicate imports

linsword13:


lib/ramble/ramble/appkit.py
```diff
@@ -35,5 +35,6 @@
 
 # Import new logger as tty to preserve old behavior
 from ramble.util.logger import logger
-from ramble.util.logger import logger as tty
+
+tty = logger
```
douglasjacobsen: We should maybe just remove the `tty` definitions. I don't know that we use them anymore, and we should actually be using `logger` instead anyway.

douglasjacobsen (changes requested):

lib/ramble/ramble/appkit.py
linsword13: Great call, updated.

douglasjacobsen (approved):

PR #1271, opened 2025-10-13
Allow ramble to assume application name is the class name

rfbgo:
Gives output like:
```
 $ ramble workspace setup --dry-run
==> Warning: Application Openfoam is missing explicit name.
==> Warning: Application Openfoam is missing explicit name.
==> Streaming details to log:
==>   /Users/robertbird/ramble/var/ramble/workspaces/openfoam/logs/setup.2025-10-13_11.39.17.out
      -- DRY-RUN -- DRY-RUN -- DRY-RUN -- DRY-RUN -- DRY-RUN --
==>   Setting up 1 out of 1 experiments:
==> Experiment #1 (1/1):
==>     name: openfoam.motorbike_20m.generated
==>     root experiment_index: 1
...
```
for app with name removed. This previously would have caused issues at license path generation time


linsword13 (dismissed):

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -200,6 +200,12 @@ def __init__(self, file_path):
         # A dict storing fom values, currently it only stores inmem FOMs
         self._fom_map = {}
 
+        if not self.name:
```
douglasjacobsen: Do you want to put this in the mixin instead of the application base class? That way all objects can have this implicit naming?

rfbgo: I'm honestly not sure, because what I think we (eventually) want is for everything to have a name and I want to minimize the reliance on this behavior not promote it..

douglasjacobsen: I think that's a good point. But I think in that case we should probably just error if name is not set, instead of implicitly setting a name.

rfbgo: We chatted about this offline, and are Ok with the semantic of "warn and continue" as long as it is consistent across objects. PR updated

douglasjacobsen (approved):

PR #1272, opened 2025-10-14
Update gcc version for ngspice application

vishalbh:


var/ramble/repos/builtin/applications/ngspice/application.py
linsword13: You probably want to use `gcc14` here?

vishalbh: yeah, missed this. Updated now.

linsword13 (approved):

PR #1274, opened 2025-10-15
Add language feature to detect conflicting modifier definitions

douglasjacobsen:
This merge adds a new language feature (`modifier_conflict`) which can be used within modifiers to control when a modifier should conflict with other modifier definitions. This aims to provide a better user experience when modifiers conflict with each other.

Tests and documentation are added as well.

### DIFF
--- a/lib/ramble/ramble/language/modifier_language.py
+++ b/lib/ramble/ramble/language/modifier_language.py
@@ -418,10 +418,10 @@ def modifier_conflict(
                 else:
                     conflict_value = MODIFIER_CONFLICT[conflict_type]
             elif isinstance(conflict_type, int):
-                if conflict_type not in MODIFIER_CONFLICT._value2member_map_:
+                try:
+                    conflict_value = MODIFIER_CONFLICT(conflict_type)
+                except ValueError:
                     usage_error = True
-                else:
-                    conflict_value = MODIFIER_CONFLICT._value2member_map_[conflict_type]
             elif isinstance(conflict_type, MODIFIER_CONFLICT):
                 conflict_value = conflict_type
             else:
--- a/lib/ramble/ramble/language/modifier_language.py
+++ b/lib/ramble/ramble/language/modifier_language.py
@@ -389,3 +389,61 @@ def package_manager_requirement(
         )
 
     return _new_package_manager_requirement
+
+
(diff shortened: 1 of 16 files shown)


lib/ramble/ramble/language/modifier_language.py
```diff
@@ -389,3 +389,61 @@ def _new_package_manager_requirement(mod):
...
+        conflict_value = None
+        usage_error = False
+        if conflict_type is not None:
+            if isinstance(conflict_type, str):
+                if conflict_type not in MODIFIER_CONFLICT._member_names_:
+                    usage_error = True
+                else:
+                    conflict_value = MODIFIER_CONFLICT[conflict_type]
+            elif isinstance(conflict_type, int):
+                if conflict_type not in MODIFIER_CONFLICT._value2member_map_:
```
rfbgo: Is this to avoid the value error `MODIFIER_CONFLICT(conflict_type)` would raise? I kind of expected this to `try` instead of using the magic map

lib/ramble/ramble/language/modifier_language.py
```diff
@@ -389,3 +389,61 @@ def _new_package_manager_requirement(mod):
...
+                else:
+                    conflict_value = MODIFIER_CONFLICT._value2member_map_[conflict_type]
+            elif isinstance(conflict_type, MODIFIER_CONFLICT):
+                conflict_value = conflict_type
+            else:
+                usage_error = True
+
+        if usage_error:
+            raise DirectiveError(
+                f"modifier_conflict directive on modifier {mod.name} was given "
```
rfbgo, line 432: Do you think this would be better as a """ block?

lib/ramble/ramble/language/modifier_language.py
douglasjacobsen: Yeah. But I updated it to use the `try` block. Enums are kind of annoying with how they have changed between python versions though...

lib/ramble/ramble/language/modifier_language.py
douglasjacobsen: This actually makes formatting it nicely for the CLI output harder (also, to make it look the same the line is too long and then `ramble style` complains about that).

rfbgo (approved):

PR #1275, opened 2025-10-15
Add pip site-packages path variable

douglasjacobsen:
This merge adds a new variable, `pip_site_packages_path` to the pip package manager to help users refer to the location (since it changes with python version).

### DIFF
--- a/var/ramble/repos/builtin/package_managers/pip/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/pip/package_manager.py
@@ -40,6 +40,12 @@ class Pip(PackageManagerBase):
     archive_pattern(os.path.join("{env_path}", "requirements.txt"))
     archive_pattern(os.path.join("{env_path}", "requirements.lock"))
 
+    package_manager_variable(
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/package_managers/pip/package_manager.py
```diff
@@ -40,6 +40,12 @@ class Pip(PackageManagerBase):
     archive_pattern(os.path.join("{env_path}", "requirements.txt"))
     archive_pattern(os.path.join("{env_path}", "requirements.lock"))
 
+    package_manager_variable(
```
linsword13, line 43: I assume this is mostly used for documentation purpose. But would it be a bit confusing if user tries to set this variable, only realizes later this is only set by the package manager?

douglasjacobsen: Yeah exactly. I wanted people to know that from looking at info of this object, they can rely on this variable begin defined.

linsword13 (approved):

PR #1280, opened 2025-10-16
Add generalized copy / link directive

douglasjacobsen:
This merge adds a new directive named stage_files which creates an executable to stage some files into a target directory, as part of an experiment.

This currently supports staging with: cp, rsync, and ln (symbolic / hard links).

Existing application definitions are updated to use this instead of hardcoded copy / link commands.

### DIFF
--- a/lib/ramble/ramble/language/application_language.py
+++ b/lib/ramble/ramble/language/application_language.py
@@ -443,7 +443,10 @@ def stage_files(
         if when_set not in app.executables:
             app.executables[when_set] = {}
 
-        if exec_name in app.executables[when_set]:
+        if (
+            exec_name in app.executables[when_set]
+            and not app.executables[when_set][exec_name].allow_extension
+        ):
             raise DirectiveError(
                 f"stage_files directive on application {app.name} is creating "
                 f"has name attribute of '{exec_name}' which already exists "
@@ -460,15 +463,18 @@ def stage_files(
         else:  # stage_method == "cp"
             stage_cmd = f"cp -r {src} {dst}"
 
+        template = [stage_cmd]
+
         # Prepend mkdir if dst has a parent directory
         parent_dir = os.path.dirname(dst)
         if parent_dir and parent_dir != ".":
-            template = f"mkdir -p {parent_dir} && {stage_cmd}"
-        else:
-            template = stage_cmd
+            template.insert(0, f"mkdir -p {parent_dir}")
 
-        app.executables[when_set][exec_name] = CommandExecutable(
-            name=exec_name, template=template, **kwargs
-        )
+        if exec_name in app.executables[when_set]:
+            app.executables[when_set][exec_name].add_template(template)
+        else:
+            app.executables[when_set][exec_name] = CommandExecutable(
+                name=exec_name, template=template, allow_extension=True, **kwargs
+            )
 
     return _execute_stage_files
--- a/lib/ramble/ramble/language/application_language.py
+++ b/lib/ramble/ramble/language/application_language.py
@@ -399,3 +399,76 @@ def cleanup(
         }
 
     return _define_cleanup
+
+
+@application_directive("executables")
+def stage_files(
+    src,
+    dst,
+    name=None,
+    when=None,
+    **kwargs,
+):
+    """Adds an executable that stages an input file or directory.
+
+    Defines a new executable that copies or links a file or directory
+    from a source to a destination. This is useful for staging input
+    files that are not managed by the `input_file` directive.
+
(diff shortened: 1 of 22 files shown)


lib/ramble/ramble/language/application_language.py
```diff
@@ -370,3 +370,76 @@ def _define_cleanup(obj):
         }
 
     return _define_cleanup
+
+
+@application_directive("executables")
+def stage_files(
```
linsword13, line 405: This is nice!

Thinking aloud, I wonder if we want to provide an option like `remove_after_run` or something, although that does seem to somewhat overlap with the recently added cleanup directive... WDYT?

lib/ramble/ramble/language/application_language.py
```diff
@@ -370,3 +370,76 @@ def _define_cleanup(obj):
...
+            app.executables[when_set] = {}
+
+        if exec_name in app.executables[when_set]:
+            raise DirectiveError(
+                f"stage_files directive on application {app.name} is creating "
+                f"has name attribute of '{exec_name}' which already exists "
+                "as an executable. Please provide a unique name attribute."
+            )
+
+        # Prepare the core staging command
```
linsword13, line 456: I kinda want it to generate commands that  "pick the best available copy method on my system", perhaps we can make the default to do that? Something along the lines of "if not specified, prefer rsync, and fallback to cp if that's not available"...

douglasjacobsen: Are you OK if the command it generates is based on the software stack where `workspace setup` is run? Or would you rather it tries to detect the appropriate option at runtime?

linsword13: I prefer the latter.

Although now that I think again, your current approach might be cleaner. As long as a user can switch the copy method via config, it seems that's good enough.

A separate thing: I wonder if the `staging_method` should instead be `copy`, `hard-link` and `soft-link`. It seems `rsync` and `cp` might be a bit too platform dependent. (And in most cases, an app-definition writer wouldn't be able to dictate whether to use `rsync` or `cp`.)

lib/ramble/ramble/language/application_language.py
douglasjacobsen: I like this idea, but I think it needs some more work to be implemented nicely.

I had thought we could just inject a `cleanup` directive call (into this directive) to perform that operation, but I think there are weird edge cases that can cause an issue. For example, if `dst` is `'{experiment_run_dir}/.'` it's hard to know what the removal should be. And I'm not even sure we can provide a nice error mechanism for that.

lib/ramble/ramble/language/application_language.py
douglasjacobsen: Do you think application developers should be able to control what stage method is used in general? Or do you think it should entirely be user specified (And if a developer wants to control it, they should write the bash)?

linsword13: I feel it (the decision of whether using rsync or cp) should be made by the user of the workload. The app developer decides whether the files are linked or copied over.

linsword13: Resolving as the code already does what I want. I misread earlier.

var/ramble/repos/builtin/applications/wrfv3/application.py
```diff
@@ -54,36 +54,43 @@ class Wrfv3(ExecutableApplication):
...
+        src="{wrf_path}/run/*",
+        dst="{experiment_run_dir}/.",
+        method="link",
     )
     executable("execute", "wrf.exe", use_mpi=True)
 
     workload(
         "CONUS_2p5km",
-        executables=["cleanup", "copy", "execute"],
+        executables=["cleanup", "stage-input", "link-run-files", "execute"],
```
linsword13: Do we want to support staging multiple `stage_files` entries under a single executable? I am thinking something like:

```
    stage_files(
        src="{input_path}/*",
        dst="{experiment_run_dir}/.",
    )
    stage_files(
        src="{wrf_path}/run/*",
        dst="{experiment_run_dir}/.",
        method="link",
    )

    executables=["cleanup", "stage-files", "execute"]
```

douglasjacobsen: You mean separately than using the name of the created `stage_files` executable?

I think some of the other apps we have have multiple `stage_files` directives, but they do not apply to every workload that stages files. This was mostly the reason behind the names.

That being said, I should probably rename these. I don't think the `method=link` is valid anymore (as I think it's entirely configured through the `config:stage_method` option now).

linsword13 (approved):

PR #1296, opened 2025-10-24
Add command to show index of experiment results

dapomeroy:
Adds a command to show attributes of experiment results, including FOMs, contexts, and template variables, to use in report filtering and grouping.

### DIFF
--- a/lib/ramble/ramble/reports.py
+++ b/lib/ramble/ramble/reports.py
@@ -11,11 +11,12 @@ import datetime
 import os
 import re
 from enum import Enum
-from typing import List
+from typing import Dict, List
 
 import llnl.util.filesystem as fs
 
 import ramble.config
+import ramble.repository
 import ramble.util.path
 from ramble.keywords import keywords
 from ramble.util.file_util import create_symlink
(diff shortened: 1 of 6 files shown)


lib/ramble/ramble/reports.py
```diff
@@ -895,3 +908,55 @@ def make_report(results_df, ws_name, args):
...
+            )
+            for fom in context["foms"]:
+                if fom["origin"] == exp["application_name"]:
+                    foms_and_vars[exp["application_name"]][exp["workload_name"]]["FOMs"].add(
+                        fom["name"]
+                    )
+                else:
+                    if fom["origin"] not in foms_and_vars:
+                        foms_and_vars[fom["origin"]] = {"FOMs": set()}
+                    foms_and_vars[fom["origin"]]["FOMs"].add(fom["name"])
```
douglasjacobsen: It'd probably nice to have some identification in the output table that this is the origin type (or maybe even what the context is) to help users understand how to translate what they see with what is in the results.

douglasjacobsen:
Thanks. This looks good overall. I think it would be nice if the sections you print had a bit more information (like the origin type, and the context) to make it easier for people to make use of this in other filtering methods. It might also be nice to have information about what the sections are intended to mean. Like, some of them are from modifiers, and others are from the package manager / workflow managers.

lib/ramble/ramble/reports.py
dapomeroy: I've updated this to group all of the attributes by object names ("Applications", "Modifiers", etc). Also added logic to extract summary names for repeats, broken out into the Experiment Summary FOMs and the names of all FOM Summary Statistics.

dapomeroy:
After working on this a bit more, I think that 'index' might be a better command name than 'attrs'. I'm happy to change it to a better name if anyone has one.

douglasjacobsen (approved):

PR #1297, opened 2025-10-24
Allow `workspace analyze` to output tabular results

douglasjacobsen:
This merge adds a new configuration section to control tabulating the results of experiments.

Tables are scoped to the block they are defined inside. Currently only csv output is supported, but the results table classes can be extended to support alternative output formats.

Tests and documentation are added for this feature.

### DIFF
--- a/lib/ramble/docs/configuration_files.rst
+++ b/lib/ramble/docs/configuration_files.rst
@@ -40,6 +40,7 @@ Currently, Ramble supports the following configuration sections:
 * :ref:`repos <repos-config>`
 * :ref:`software <software-config>`
 * :ref:`success_criteria <success-criteria-config>`
+* :ref:`tables <tables-config>`
 * :ref:`variables <variables-config>`
 * :ref:`variants <variants-config>`
 
@@ -142,6 +143,7 @@ In the above ``[optional_definitions]`` can include any of:
 * :ref:`internals <internals-config>`
 * :ref:`modifiers <modifiers-config>`
 * :ref:`success_criteria <success-criteria-config>`
+* :ref:`tables <tables-config>`
 * :ref:`variables <variables-config>`
 * :ref:`variants <variants-config>`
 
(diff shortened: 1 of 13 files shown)


lib/ramble/ramble/results_table.py
```diff
@@ -0,0 +1,367 @@
...
+            directory (str): Directory to write tabular data into
+            timestamp (str): Timestamp to apply to table output files
+        """
+        self._to_dataframe()
+        self._group_dataframe()
+        self._sort_dataframe()
+
+        extension = "csv"
+        inner_delim = "."
+        filename = self.name + inner_delim + timestamp + inner_delim + extension
```
linsword13, line 291: If I define two table configs with the same name, in two different experiments, would that cause a conflict?

lib/ramble/ramble/pipeline.py
```diff
@@ -297,6 +302,8 @@ def _complete(self):
             summary_only=self.summary_only,
         )
 
+        self.workspace.dump_tables(self.experiment_set, self.filters)
```
linsword13, line 305: Mostly thinking aloud: do we want an option to enable/disable the table generation?

lib/ramble/ramble/results_table.py
douglasjacobsen: Yes. And currently it would just silently ignore this, mostly because you didn't put anything in the table name to distinguish between them.

You'd end up getting the last one that was processed, FWIW.

lib/ramble/ramble/pipeline.py
douglasjacobsen: I'm fine either way. It's easy enough to add an option, but my goal was mostly that tables are cheap enough that you never really want to disable them if you have them in your config.

linsword13: Sounds good. Yeah if the table generation is not costly (and it doesn't currently clutter the results output), it sounds good to have it enabled.

lib/ramble/ramble/results_table.py
linsword13: I see, I wonder if it'd be useful to detect and warn about such duplicate table names?

lib/ramble/ramble/experiment_set.py
```diff
@@ -573,11 +576,44 @@ def _ingest_experiments(self, die_on_validate_error=True):
...
 
+                workload_names.add(app_inst.expander.workload_name)
+
                 app_inst.set_success_list(final_context.success_criteria)
                 rendered_experiments.add(final_exp_namespace)
                 self.experiments[final_exp_namespace] = app_inst
                 self.experiment_order.append(final_exp_namespace)
 
+        # Generate focused tables for results
+        app_context = self._context[self._contexts["application"]]
```
rfbgo, line 590: Given this is all table specific do you think it would be useful to pull it out into it's own function?

lib/ramble/ramble/experiment_set.py
```diff
@@ -573,11 +576,44 @@ def _ingest_experiments(self, die_on_validate_error=True):
...
+            results_table.add_where(
+                "'{application_name}' == '" + app_context.context_name + "'",
+            )
+
+        wl_context = self._context[self._contexts["workload"]]
+        for table in wl_context.tables:
+            for workload_name in workload_names:
+                results_table = self._workspace.results_tables.add_table_template(table)
+                results_table.add_where(
+                    [
```
rfbgo, line 602: Any reason not to fstring this? I feel like it would be less confusing (but I guess it could be style that is easier to line break?)

lib/ramble/ramble/results_table.py
```diff
@@ -0,0 +1,367 @@
...
+                                        "Non-unique values found " f"for column {self.name}"
+                                    )
+                                value = fom["value"]
+        return value
+
+
+class ResultsTable:
+    """A single results table based on the tables.py schema"""
+
+    _default_group_method = "max"
```
rfbgo, line 124: Is there an easy to understand possible list of values for this somewhere? I'm wondering if we can document the potential values so it's easy to see at a glance

lib/ramble/docs/configuration_files.rst
```diff
@@ -675,6 +677,85 @@ use when determining if they were successful or not. Its format is as follows:
...
+
+.. code-block:: yaml
+
+   columns:
+   - name: "column name template"
+     where:
+     - "list of expressions"
+     - "to filter experiments"
+     - "when building this column"
+     expression: "Ramble style expression for column value"
```
rfbgo: nit for clarity? 

```suggestion
     expression: "Ramble-style expression for column value"
```

lib/ramble/ramble/results_table.py
linsword13: I am mostly concerned about when they differ only by the timestamp, such that there are:

* my_table.ts0.csv
* my_table.ts1.csv

And it's hard to predict which one the `my_table.latest.csv` points to.

douglasjacobsen: We kind of abuse the fact that they have duplicate names though, so we can aggregate things (easily) into a single table if we want to.

I think the latest timestamp will always point to the last one though.

douglasjacobsen: Actually, sorry I think I'm wrong.

If you define two tables that have the same name but different columns, I believe what will happen currently is that you will get a single table with the name, but all of the columns. Rows will have a values of `None` for any columns that are not part of the tables they are included in.

lib/ramble/ramble/experiment_set.py
douglasjacobsen: Done.

lib/ramble/ramble/experiment_set.py
douglasjacobsen: Done.

lib/ramble/docs/configuration_files.rst
douglasjacobsen: Done

lib/ramble/ramble/results_table.py
douglasjacobsen: I added a link to this in our docs, but it's any groupby method available from Pandas (https://pandas.pydata.org/docs/reference/groupby.html#dataframegroupby-computations-descriptive-stats)

rfbgo: Thanks!

lib/ramble/ramble/results_table.py
```diff
@@ -0,0 +1,367 @@
...
+
+        Args:
+            conf_dict (dict): dictionary structured like the column schema in tables.py
+
+        """
+        # Extract column attributes
+        for attr in self._column_attrs:
+            setattr(self, attr, str(conf_dict[attr]) if attr in conf_dict else None)
+
+        if self.expression and self.figure_of_merit:
```
linsword13, line 42: Do we want to ensure one of them is not None?

douglasjacobsen: Done

lib/ramble/ramble/results_table.py
linsword13: Ah I see, makes sense now.

linsword13 (approved):

PR #1304, opened 2025-10-28
Support `env_var_name` arg to auto inject env-var from a variable directive

linsword13:


lib/ramble/ramble/language/application_language.py
```diff
@@ -233,6 +233,7 @@ def workload_variable(
     expandable: bool = True,
     track_used: bool = True,
     when=None,
+    env_var_name=None,
```
douglasjacobsen: Annoying nit; Can we make the name the more verbose `environment_variable_name`?

linsword13: Definitely, updated.

douglasjacobsen (approved):

PR #1308, opened 2025-10-29
Add expander support for string methods

linsword13:


douglasjacobsen (commented):
With this PR, do we need the other `_str` methods? (i.e. does this cover `upper_str`?)

linsword13:
> With this PR, do we need the other `_str` methods? (i.e. does this cover `upper_str`?)

Yes it does. But I am thinking to keep these existing ones around, just so that workspaces that use them don't need to be updated.

douglasjacobsen:
I think given we just added them, it's probably safe to remove. I only used them in table headers when I was adding them.

linsword13:
> I think given we just added them, it's probably safe to remove. I only used them in table headers when I was adding them.

Sounds good, updated.

douglasjacobsen (approved):

PR #1321, opened 2025-11-06
Allows software_spec's to be injected if not present in environments

douglasjacobsen:
This merge adds a new keyword attribute (inject_if_missing) to all software_spec directive calls. This allows objects to quietly inject packages into experiment environments without having to edit the workspace configuration file.

Required compilers are also automatically injected into the environment to be used for the new packages.

### DIFF
--- a/lib/ramble/ramble/software_environments.py
+++ b/lib/ramble/ramble/software_environments.py
@@ -14,6 +14,7 @@ import ramble.util.colors as rucolor
 from ramble.expander import Expander
 from ramble.namespace import namespace
 from ramble.util.logger import logger
+from ramble.util.spec_utils import SoftwareSpec
 
 SUB_INDENT = 2
 
@@ -55,6 +56,7 @@ class SoftwarePackage:
         self.pkg_info = pkg_info
         self._package_type = "Base"
         self._used = False
+        self.injected = False
 
     def mark_used(self):
         """Mark this package a used"""
@@ -107,7 +109,9 @@ class SoftwarePackage:
         indentation = " " * indent
         color = rucolor.level_func(color_level)
 
-        out_str = color(f"{indentation}{self._package_type} package: {self.name}\n")
+        injected_str = "" if not self.injected else " (injected by ramble)"
+
+        out_str = color(f"{indentation}{self._package_type} package: {self.name} {injected_str}\n")
         return out_str
 
     def __str__(self):
@@ -466,14 +470,24 @@ class SoftwareEnvironment:
         Returns:
             (bool): True if environments are equivalent, False otherwise
         """
-        equal = self.name == other.name and len(self._packages) == len(other._packages)
+        equal = self.name == other.name
 
         if not equal:
             return False
 
-        for self_pkg, other_pkg in zip(self._packages, other._packages):
-            if self_pkg != other_pkg:
-                return False
+        self_pkgs = {}
+        other_pkgs = {}
+
+        for self_pkg in self._packages:
+            if not self_pkg.injected:
+                self_pkgs[self_pkg.name] = self_pkg
+
+        for other_pkg in other._packages:
+            if not other_pkg.injected:
+                other_pkgs[other_pkg.name] = other_pkg
+
+        if self_pkgs != other_pkgs:
+            return False
 
         return True
 
@@ -853,6 +867,7 @@ class SoftwareEnvironments:
 
                             if rendered_pkg.compiler:
                                 cur_compiler = rendered_pkg.compiler
+
                     if not added:
                         raise RambleSoftwareEnvironmentError(
                             f"Compiler {pkg.compiler} used, but not "
@@ -920,6 +935,45 @@ class SoftwareEnvironments:
 
         for pkg in environment._packages:
             yield pkg.spec_str(all_packages=self._rendered_packages, compiler=False)
+
+    def add_spec_to_environment(
+        self,
+        environment: SoftwareEnvironment,
+        spec: SoftwareSpec,
+        expander: Expander,
+        package_manager,
+    ):
+        """Add a spec to a given environment
+
+        Creates a new template / rendered package (if needed) from the input spec,
+        and adds to the template and rendered environment as a package in the environment.
+
+        Args:
+            environment (SoftwareEnvironment): Rendered environment to add package to
+            spec (ramble.util.spec_utils.SoftwareSpec): Software spec to add to environment
+            expander (ramble.expander.Expander): Experiment's expander object,
+                                                 to render package / environment with
+            package_manager: Package manager from the experiment
+        """
+
+        pm_name = package_manager.spec_prefix
+
+        if spec.name not in self._package_templates:
+            template_package = TemplatePackage(spec.name, spec.to_dict())
+
+            self._package_templates[spec.name] = template_package
+
+        template_package = self._package_templates[spec.name]
+
+        rendered_package = template_package.render_package(expander, package_manager)
+        rendered_package.injected = True
+
+        if rendered_package.name not in self._rendered_packages:
+            template_package.add_rendered_package(
+                rendered_package, self._rendered_packages, pm_name
+            )
+
+        environment.add_package(rendered_package)
 
     def _check_environment(self, environment):
         """Check an environment for common issues
--- a/lib/ramble/ramble/software_environments.py
+++ b/lib/ramble/ramble/software_environments.py
@@ -470,9 +470,7 @@ class SoftwareEnvironment:
         Returns:
             (bool): True if environments are equivalent, False otherwise
         """
-        equal = self.name == other.name
-
-        if not equal:
+        if not self.name == other.name:
             return False
 
         self_pkgs = {}
--- a/lib/ramble/ramble/test/end_to_end/software_spec_injection.py
+++ b/lib/ramble/ramble/test/end_to_end/software_spec_injection.py
@@ -1,0 +1,261 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import os
+
+import pytest
+
+import ramble.workspace
+from ramble.error import RambleCommandError
+from ramble.main import RambleCommand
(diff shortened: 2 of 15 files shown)


lib/ramble/ramble/test/end_to_end/software_spec_injection.py
```diff
@@ -0,0 +1,197 @@
...
+from ramble.pkg_man.builtin.spack_lightweight import SpackRunner
+from ramble.software_environments import RambleSoftwareEnvironmentError
+from ramble.util.command_runner import RunnerError
+
+pytestmark = pytest.mark.usefixtures("mutable_config", "mutable_mock_workspace_path")
+
+workspace = RambleCommand("workspace")
+
+
+def test_software_spec_injection_works(mock_modifiers, workspace_name):
```
linsword13, line 25: nit: Probably can mark these tests as potentially slow?

lib/ramble/ramble/software_environments.py
```diff
@@ -471,8 +472,23 @@ def __eq__(self, other):
         if not equal:
             return False
 
-        for self_pkg, other_pkg in zip(self._packages, other._packages):
-            if self_pkg != other_pkg:
+        if len(self._packages) != len(other._packages):
```
linsword13: nit: this seems redundant given the above (line 470) already covers this?

I wonder if the following snippet may be slightly more concise:

```
    if self.name != other.name or len(self._packages) != len(other._packages):
        return False

    self_pkgs = {pkg.name: pkg for pkg in self._packages}
    other_pkgs = {pkg.name: pkg for pkg in other._packages}

    return self_pkgs == other_pkgs
```

lib/ramble/ramble/test/end_to_end/software_spec_injection.py
douglasjacobsen: This one shouldn't be slow. `mock_modifiers` just adds the mock repo rather than iterating over all mock modifiers.

lib/ramble/ramble/software_environments.py
douglasjacobsen: Good point. Removed!

lib/ramble/ramble/test/end_to_end/software_spec_injection.py
linsword13: Ah I see. I saw the requirement of spack, and immediately defaulted to thinking that may be slow. SG.

lib/ramble/ramble/software_environments.py
```diff
@@ -471,8 +472,20 @@ def __eq__(self, other):
...
+        self_pkgs = {}
+        other_pkgs = {}
+
+        for self_pkg in self._packages:
+            self_pkgs[self_pkg.name] = self_pkg
+
+        for other_pkg in other._packages:
+            other_pkgs[other_pkg.name] = other_pkg
+
+        for pkg_name in self_pkgs:
```
linsword13: I think there might be a super edgy case (I don't know if in practice that'd ever come up) where:

```
self._packages = [
    "zlib@1",
    "zlib@2",
]
other._packages = [
    "zlib@2",
    "something_else",
]

# Then
self_pkgs = {"zlib": "zlib@2"}
other_pkgs = {"zlib": "zlib@2", "something_else": "something_else"}
```

And even though they are different, this function returns `True`.

In fact, since you already have the two dictionaries, a direct comparison `self_pkgs == other_pkgs` should be sufficient.

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
```diff
@@ -210,6 +215,138 @@ def populate_inventory(
...
+                for specs in obj.software_specs.values():
+                    for spec in specs:
+                        pkg_name = self.package_name_from_spec(spec.pkg_spec)
+
+                        if (
+                            self.app_inst.expander.satisfies(
+                                spec.when,
+                                variant_set=obj.experiment_variants(),
+                            )
+                            and spec.inject_if_missing
```
linsword13: I wonder if it makes sense to do this check first, such that it can short-circuit the more expensive expander call?

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
```diff
@@ -210,6 +215,138 @@ def populate_inventory(
...
+        # Inject any missing, but injected compilers
+        for _, obj in self.app_inst._objects():
+            for comps in obj.compilers.values():
+                for comp in comps:
+                    if (
+                        self.satisfy_when(
+                            comp.when,
+                            variant_set=obj.experiment_variants(),
+                        )
+                        and comp.inject_if_missing
```
linsword13: Same here perhaps to do the cheaper check first

lib/ramble/ramble/software_environments.py
douglasjacobsen: I actually had to change this logic to ignore injected packages. So, review the updated logic and let me know what you think (once I push it).

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
douglasjacobsen: Good call. Fixed!

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
douglasjacobsen: Good call. Fixed!

lib/ramble/ramble/software_environments.py
douglasjacobsen: As mentioned below, I had to update this logic to fix a separate issue. Let me know what you think now.

lib/ramble/ramble/software_environments.py
```diff
@@ -466,13 +470,30 @@ def __eq__(self, other):
...
+                self_pkgs[self_pkg.name] = self_pkg
+
+        for other_pkg in other._packages:
+            if not other_pkg.injected:
+                other_pkgs[other_pkg.name] = other_pkg
+
+        if self_pkgs != other_pkgs:
+            return False
+
+        for pkg_name in self_pkgs:
```
linsword13: I think this for-loop is actually not needed anymore. The dict comparison above should take care of it.

douglasjacobsen: Good point, removed it.

linsword13 (dismissed):

douglasjacobsen:
BTW, I just added a new test (which is why @linsword13 's review was dismissed). This test covers that a package is not injected if it is already defined in the environment.

lib/ramble/ramble/software_environments.py
```diff
@@ -466,14 +470,24 @@ def __eq__(self, other):
         Returns:
             (bool): True if environments are equivalent, False otherwise
         """
-        equal = self.name == other.name and len(self._packages) == len(other._packages)
+        equal = self.name == other.name
 
         if not equal:
             return False
```
rfbgo: nit: is there much benefit to splitting this calculation out? 

```suggestion
        if not self.name == other.name:
            return False
```

var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
```diff
@@ -44,6 +44,16 @@ def __init__(self, file_path):
 
         self._runner = None
 
+    name_regex = re.compile(r"\s*(?P<name>[\w-]+).*")
+
+    def package_name_from_spec(self, spec):
```
rfbgo, line 49: nit: is there some nice way to pull this common code across PMs into a shared location?

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
```diff
@@ -210,6 +215,137 @@ def populate_inventory(
 
         pass
 
+    def define_missing_packages(self, workspace):
+        """Injection of missing packages that are auto-injected by objects
```
rfbgo, line 219: Is this the same but simpler?
```suggestion
        """Auto-inject missing packages by objects
```

rfbgo (commented):
LGTM. A few (slightly annoying) little questions

lib/ramble/ramble/software_environments.py
douglasjacobsen: No, it used to be more complex but we removed the other part of it. I've changed it to what you suggested.

var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
douglasjacobsen: It's not actually common across PMs. Each PM has a (potentially very different) definition of how a package name is split from the rest of a pkg spec.

This method is basically a required attribute of a PM, to say "Given a package spec string, how can I extract the name of the package from it?"

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
douglasjacobsen: Not exactly. Ultimately, we only inject packages that are marked as injectable. We do not arbitrarily inject missing packages from objects though.

var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
rfbgo: I understand that for some it can be different; but pip, env mod, and this one all have identical code though right?

douglasjacobsen: They might currently be the same, because we don't have any examples where they need to be different, but there also isn't a great mechanism of DRY-ing them out while making them a requirement for package managers. Without injecting a base class to inherit from (which I'm not in support of) it's difficult to make these abstract methods (and required as a result) while also unifying the interface for them.

I'd personally prefer that package manager developers are required to think about how to get a package name from a spec for that package manager.

rfbgo: Yea I agree with that, and it's hard to find a nice way to dry it. It would almost be like:

```
 def package_name_from_spec(self, spec):
    return self._name_regex_match() # shared function in parent that has the same body
```

and call some other none default function in the parent

That's why it was a nit.. it feels bad to not have this DRY but it's not clear what a great way to have a sensible helper would be 

We can leave as is, just wanted to check it was a conscious choice

douglasjacobsen: Yeah, sounds good. It was definitely intentional. The primary thing I was looking for was making it required within the package managers.

rfbgo (approved):

PR #1325, opened 2025-11-08
Add iterative variable definition based on variant values

douglasjacobsen:
This merge updates the logic for defining undefined variables based on the values of variants to be iterative.

This allows variables to recursively update the definition of variants, before allowing them to define additional variables.

Tests are added to cover this (arguably complex) case as well.

### DIFF
--- a/var/ramble/repos/builtin/base_classes/application-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/application-base/base_class.py
@@ -621,17 +621,31 @@ class ApplicationBase(ObjectMixin, metaclass=ApplicationMeta):
                         if not var.expandable:
                             self.no_expand_vars.add(var.name)
 
-        # Define missing variables
-        for _, obj in self._objects():
-            for var, val in obj.selected_variables.items():
-                if var not in self.variables:
-                    self.define_variable(var, val.default)
+        self.define_missing_variables()
 
         self.expander.set_no_expand_vars(self.no_expand_vars)
         if experiment_set and experiment_set._workspace:
             self.expander.replacement_paths = (
                 experiment_set._workspace.workspace_paths()
             )
+
+    def define_missing_variables(self):
+        """Iterate over missing variable definitions, and add them until there
+        are no more to add."""
+
+        while True:
+            missing_vars = {}
+            for _, obj in self._objects():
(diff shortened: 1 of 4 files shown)


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -1188,9 +1196,13 @@ def build_modifier_instances(self):
                 mod_inst.expander.add_no_expand_var(var)
 
             # Define any missing modifier variables
-            for var, val in mod_inst.selected_variables.items():
-                if var not in self.variables:
-                    self.define_variable(var, val.default)
+            defined = 1
```
linsword13: This block is repeated in two other places, perhaps DRY it?

douglasjacobsen: Fixed! I also updated the logic to support iterative cross object definitions as well.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -621,18 +621,31 @@ def set_variables_and_variants(self, variables, variants, experiment_set):
...
-                    self.define_variable(var, val.default)
+        self.define_missing_variables()
 
         self.expander.set_no_expand_vars(self.no_expand_vars)
         if experiment_set and experiment_set._workspace:
             self.expander.replacement_paths = (
                 experiment_set._workspace.workspace_paths()
             )
 
+    def define_missing_variables(self):
```
linsword13, line 632: Purely personal preference, I wonder how do you feel about this slightly less nested version? I also added a check using the `missing_vars`, as I am curious if we want first-wins or last-wins semantic?

```
while True:
        missing_vars = {}
        for _, obj in self._objects():
            for var, val in obj.selected_variables.items():
                // Check for its presence in missing_vars, for the "first-defined-wins" semantic
                if var not in self.variables and var not in missing_vars:
                    missing_vars[var] = val.default

        if not missing_vars:
            break

        for var, default_val in missing_vars.items():
            self.define_variable(var, default_val)
```

douglasjacobsen: LGTM. Updated.

linsword13 (approved):

PR #1329, opened 2025-11-11
Add usage filters to executable_modifiers

douglasjacobsen:
This merge allows executable_modifiers defined within modifiers to apply usage filters. Usage filters can change what the rules are for applying an executable_modifier to a given executable. Pre-defined filters include "None" (i.e. no restrictions), "once" (only apply once), "first_mpi" (apply to the first MPI executable), or "all_mpi" (apply to all MPI executables).

Modifiers can define their own additional usage_filters as well, to change the logic.

### DIFF
--- a/var/ramble/repos/builtin/base_classes/modifier-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/modifier-base/base_class.py
@@ -70,6 +70,7 @@ class ModifierBase(ObjectMixin, metaclass=ModifierMeta):
         self.expander = None
         self._usage_mode = None
         self.app_inst = None
+        self._executable_modification_applied = set()
 
         self._mod_regex = re.compile(
             self._mod_prefix_builtin + f"{self.name}{NS_SEPARATOR}"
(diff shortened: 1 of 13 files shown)


var/ramble/repos/builtin/base_classes/modifier-base/base_class.py
```diff
@@ -375,22 +376,135 @@ def applies_to_executable(self, executable):
...
+    def filter_once(self, exec_mod, executable) -> bool:
+        """Usage filter for only allowing an executable modifier to be applied
+        once in an experiment"""
+        return not self.executable_modifier_applied(exec_mod)
+
+    @executable_modifier_usage_filter("first_mpi")
+    def filter_first_mpi(self, exec_mod, executable) -> bool:
+        """Usage filter for only applying executable modifier to the first MPI
+        executable in an experiment"""
+        return executable.mpi and not self.executable_modifier_applied(
```
linsword13, line 411: One thing I am thinking, is whether it makes sense to have a filtering stage against the full list of executables. That way this kind of applying only to the i-th executable filtering becomes easier.

douglasjacobsen: I'm not sure I follow. Can you give me an example of what you are thinking? 

The executable in this command should be the full merged executables (but not a builtin).

linsword13: Yeah I don't have a super concrete proposal. Roughly what I am thinking is that right now we check the applicability against individual exec_node (`for exec_node in exec_graph.walk():`), if instead we do the filtering against all the exec nodes (in the right topological order), i.e., the filtering receives the sorted list of exec nodes, then it becomes easier to express something like "only apply to the last mpi-enabled executable", etc.

douglasjacobsen: I think that's a good idea, but it also kind of requires a decent amount of tangential work. For example, the executable a modifier operates on is actually a copy of the real executable (so you can't check that they are the same, in the topological ordering at least).

Here's an example of how you would write it (something complex like you described) currently: https://github.com/GoogleCloudPlatform/ramble/pull/1329/files#diff-a9ae4a5ab12d873fcd4f398544dbf34a6c1cba06ba097778eb92c86f97c4fd48R684

I think without passing in the graph, a modifier developer would have to extract it from the application instance (which isn't difficult today). But we don't have nice methods for inspecting the graph (or determining where in the graph a given node exists).

I can make a task for adding this kind of stuff, but it feels more like we want graph introspection utilities than we need to do anything about changing the behavior of filters. Do you think that sounds accurate?

linsword13: Thanks for the explanation! Agreed that this kind of utilities should be slated for later (if at all.) Resolving.

linsword13 (approved):

PR #1331, opened 2025-11-12
Add caching to experiment results to speed repeated analysis

douglasjacobsen:
This merge introduces file based caching / serialization of experiment results within each experiment's run directory. These are then read in (if they are not invalid) instead of re-processing the output files for the experiment.

These caches are invalidated if the timestamp on the cache is older than any other file in the experiment directory, or if the hashes on the objects (i.e. application / modifier / etc) changed relative to the last time the cache was written.

Tests are also added for this.

### DIFF
--- a/var/ramble/repos/builtin/base_classes/application-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/application-base/base_class.py
@@ -54,6 +54,7 @@ from ramble.experiment_result import ExperimentResult, ExperimentStatus
 from ramble.language.application_language import ApplicationMeta
 from ramble.language.shared_language import (
     SharedMeta,
+    archive_pattern,
     register_builtin,
     register_phase,
 )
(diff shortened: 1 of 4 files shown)


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -2445,213 +2490,236 @@ def format_context(context_match, context_format):
...
-
-        files, f_defs, inmem_defs = self._analysis_dicts(criteria_list)
-
-        exp_lock = self.experiment_lock
-
-        fom_values = {}
-        # Iterate over files. We already know they exist
-        with lk.ReadTransaction(exp_lock):
-            for file, file_conf in files.items():
+        if not self.result.read_cache(workspace, self):
```
linsword13: nit: would it save some indentation if we exit earlier when the cache is present?

```
if self.result.read_cache(workspace, self):
    self.result.finalize(workspace)
    return

criteria_list = ...
```

linsword13 (commented):
Great stuff, LGTM with one nit comment.

rfbgo (dismissed):

var/ramble/repos/builtin/base_classes/application-base/base_class.py
douglasjacobsen: Done!

rfbgo (approved):

PR #1332, opened 2025-11-13
Several perf improvements

linsword13:
Several perf improvements:

* Untangle to avoid repetitive calls to `expander.satisfies`.
* More efficient `get_node` in `graphs.py`.
* Add caching to `ast.parse` in expander.

### DIFF
--- a/lib/ramble/ramble/expander.py
+++ b/lib/ramble/ramble/expander.py
@@ -25,6 +25,18 @@ from ramble.util.path import substitute_config_variables
 
 import spack.util.naming
 
+_ast_cache: Dict[str, str] = {}
+
+
+def _ast_parse(in_str):
+    """Parse a string into an AST, with caching."""
+    if in_str in _ast_cache:
+        return _ast_cache[in_str]
+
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/expander.py
```diff
@@ -418,6 +418,7 @@ def __init__(self, variables, experiment_set, no_expand_vars=None):
         if no_expand_vars is None:
             no_expand_vars = set()
 
+        self._ast_cache = {}
```
douglasjacobsen: Since we're caching the AST, and not the output, do you think there's a nice way to make this global, rather than local to each individual expander?

Like, maybe a class static attribute or something?

douglasjacobsen (commented):
Thanks, this LGTM. I have one question, but otherwise this is good.

lib/ramble/ramble/expander.py
linsword13: Great idea, added.

douglasjacobsen (approved):

PR #1333, opened 2025-11-13
Improve reports performance

dapomeroy:
Improves performance of `ramble results report`, especially with large datasets.

### DIFF
--- a/lib/ramble/ramble/reports.py
+++ b/lib/ramble/ramble/reports.py
@@ -19,7 +19,7 @@ import ramble.repository
 import ramble.util.path
 from ramble.keywords import keywords
 from ramble.util.file_util import create_symlink
-from ramble.util.foms import BetterDirection, FomType
+from ramble.util.foms import BetterDirection, FomType, SummaryFoms
 from ramble.util.logger import logger
 
 import spack.util.spack_yaml as syaml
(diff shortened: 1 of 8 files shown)


dapomeroy:
The main change here is in how the data is extracted from the results file. Previously, the whole results file was dumped into a Pandas Dataframe and then filtered and queried from there. This worked fine with small datasets, but didn't scale well to results files with ~1M or so rows, lots of FOMs, many modifiers, etc. Now it only extracts what's needed for each chart from the results dict and leaves the rest behind, making subsequent transformations much faster.

lib/ramble/ramble/reports.py
```diff
@@ -124,81 +141,239 @@ def filter_exp_results(experiments: list):
...
+
+        if all_vars:
+            if "All Variables" not in app_dict[exp["workload_name"]]:
+                app_dict[exp["workload_name"]]["All Variables"] = set()
+            for var_name in exp.keys():
+                # Exclude non-variables
+                if var_name in vars_to_ignore:
+                    continue
+                if re.search(dir_regex, var_name):
+                    continue
```
rfbgo: In practice this is always actually ends with, right? I wonder if we should prefer regular python over regex

lib/ramble/ramble/reports.py
```diff
@@ -124,81 +141,239 @@ def filter_exp_results(experiments: list):
...
+                if re.search(path_regex, var_name):
+                    continue
+                app_dict[exp["workload_name"]]["All Variables"].add(var_name)
+
+            for var_name in exp["RAMBLE_VARIABLES"].keys():
+                if var_name in vars_to_ignore:
+                    continue
+                if re.search(dir_regex, var_name):
```
rfbgo: Can we dry these repeated checks by using a function?

lib/ramble/ramble/reports.py
```diff
@@ -124,81 +141,239 @@ def filter_exp_results(experiments: list):
...
+            )
+
+        for context in exp["CONTEXTS"]:
+            if not context["foms"]:
+                continue
+            app_dict[exp["workload_name"]]["Contexts"].add(context["name"])
+            for fom in context["foms"]:
+                if fom["origin"] == exp["application_name"]:
+                    # If it's a repeat summary, add summary FOMs and stat names
+                    if fom["name"] == "Experiment Summary":
```
rfbgo: It would be nice if we dried these magic strings across the code base so they had a "shared fate" and we only have to change them in one place

rfbgo (changes requested):
Overall I really like this, and it's great to see even more stuff being tested!

I suggested a few changes based on readability and magic string usage. Since the business logic is fairly complex here I think it's worth us spending a bit of extra time making sure things are simplified and tidy

lib/ramble/ramble/reports.py
dapomeroy: Good suggestion. Yes, it does look like they are all at the end of the string. I'll switch this to use endswith().

lib/ramble/ramble/reports.py
dapomeroy: Done

lib/ramble/ramble/reports.py
dapomeroy: Converted the 'experiment summary' strings to an enum

rfbgo (approved):

PR #1342, opened 2025-11-20
Allow rendering of chained experiments

douglasjacobsen:
This merge enables the ability to render experiments into the chained experiment list within an experiment set. Previously, rendered experiments could only be added as base experiments.

### DIFF
--- a/lib/ramble/ramble/context.py
+++ b/lib/ramble/ramble/context.py
@@ -19,6 +19,24 @@ class Context:
     (such as application, workload, or experiment) and logic to merge in
     additional contexts by order of precedence."""
 
+    output_mapping = {
+        "variables": namespace.variables,
+        "variants": namespace.variants,
+        "env_variables": namespace.env_var,
+        "internals": namespace.internals,
+        "chained_experiments": namespace.chained_experiments,
+        "modifiers": namespace.modifiers,
+        "template": namespace.template,  # TODO: Make sure this is good
+        "exclude": namespace.exclude,
+        "zips": namespace.zips,
+        "tables": namespace.tables,
+        "tags": namespace.tags,
+        "matrices": namespace.matrices,
+        "n_repeats": namespace.n_repeats,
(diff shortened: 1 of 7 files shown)


lib/ramble/ramble/context.py
```diff
@@ -99,6 +117,28 @@ def merge_context(self, in_context):
         if in_context.tables:
             self.tables.extend(in_context.tables)
 
+    def to_workspace_config(self, application_name, workload_name):
+        experiment_config = {}
+
+        for attr_name, namespace_name in self.output_mapping.items():
+            attr_val = getattr(self, attr_name, None)
+            if attr_val is not None and attr_val:
```
linsword13: Can this be simplified as `if attr_val`?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -617,6 +618,40 @@ def set_variables_and_variants(self, variables, variants, experiment_set):
...
+
+        Args:
+            remove_keys (set): Set of keys to remove from variable definitions
+
+        Returns:
+            (dict): A dictionary of variable, value pairs from this experiment.
+        """
+
+        def _remove_key(key, key_dict):
+            if key in key_dict:
```
linsword13: nit: `key_dict.pop(key, None)` can be used to avoid the membership check.

linsword13 (commented):
Cool stuff! A couple nit comments, but otherwise LGTM.

lib/ramble/ramble/context.py
douglasjacobsen: Thanks, fixed.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
douglasjacobsen: Thanks, this is fixed.

rfbgo (approved):

linsword13 (approved):

PR #1353, opened 2025-12-04
Move definition of required mpi variables to application

douglasjacobsen:
This merge moves the definition of undefined, but required MPI variables to the application base class instead of having it defined when creating an experiment set.

This allows the variables to be defined based on workload / object variables rather than erroring before workload / object variables are defined.

### DIFF
--- a/var/ramble/repos/builtin/base_classes/application-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/application-base/base_class.py
@@ -1585,15 +1585,11 @@ class ApplicationBase(ObjectMixin, metaclass=ApplicationMeta):
                                 "{mpi_command}",
                                 exec_vars,
                             ).strip()
-                            if (
-                                not raw_mpi_cmd
-                                and int(
-                                    self.expander.expand_var_name(
-                                        self.keywords.n_nodes
-                                    )
-                                )
-                                > 1
-                            ):
+                            n_nodes = self.expander.expand_var_name(
+                                self.keywords.n_nodes
+                            )
+                            n_nodes = 1 if not n_nodes else int(n_nodes)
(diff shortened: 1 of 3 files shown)


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -3526,6 +3522,47 @@ def _objects(self, exclude_types=None):
 
     def set_required_variables(self):
         """Set required variables from all objects"""
+
```
linsword13, line 3525: nit: I kinda like that previously it encapsulates these calculations inside a `compute_mpi_vars` function

linsword13 (commented):
LGTM, with one nit comment.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
douglasjacobsen: I moved them here, because they are technically defined because they are required rather than because they are MPI variables.

Also, I kind of want to move away from these being required if possible, to enable more workflows to map easily to Ramble.

linsword13: Yeah I am definitely aligned on the move. I just like that the calculations of these variable values  can be defined in a separate function as opposed to directly put into the `set_required_variables` function. That way it's grouped more tightly and can be easier down-the-line when we want to remove the logic.

douglasjacobsen: I made a sub-method to house the MPI variables. Let me know if that seems sufficient to you.

linsword13 (approved):

PR #1355, opened 2025-12-08
Fallback to string instead of bailout in binary op eval

linsword13:
This is potentially more friendly to "bin-op-alike" expressions, such as "tcp-ip".


linsword13:
The codecov patch miss is due to the code containing an if-block for handling pre Python 3.8 compatibility. (So that block is actually exercised when run with python 3.7.)

douglasjacobsen (approved):

PR #1365, opened 2025-12-30
Allow Ramble to create upstream BigQuery db via the new data command 

rfbgo:


lib/ramble/ramble/schema/metadata.py
```diff
@@ -0,0 +1,21 @@
...
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+metadata_schema_version = 0.1
+metadata_schema = {
+    0.1: {
```
douglasjacobsen, line 11: nit; Can you make the key here (0.1) just refer to the schema version above?

lib/ramble/ramble/schema/fom.py
```diff
@@ -0,0 +1,35 @@
...
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+fom_schema_version = 0.1
+fom_schema = {
+    0.1: {
```
douglasjacobsen, line 11: Same comment about version referral here.

Do you expect that the versions of these various tables will differ? If not, can we unify the version definition? But if they might differ eventually it is fine to live in the separate files.

lib/ramble/ramble/schema/experiment.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2022-2025 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+experiment_schema_version = 0.1
```
douglasjacobsen, line 9: Same comment about version referral to the key below.

lib/ramble/ramble/cmd/zeppelin.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2022-2025 The Ramble Authors
```
douglasjacobsen: Annoying EOY comment... you probably need to rebase this branch and update the license headers so we don't have messed up versions on develop.

douglasjacobsen (changes requested):
Looks good! Thanks for working on this. A few nits, and one thing that probably needs to be addressed.

lib/ramble/ramble/schema/fom.py
rfbgo: I want to intentionally support them differing, so we can communicate what changed

lib/ramble/ramble/schema/metadata.py
rfbgo: We chatted about this offline, but I want to maintain historical backwards compatibility. `metadata_schema` is the map and the version var is the "selector". I don't think the selector should be used in the definition

lib/ramble/ramble/cmd/zeppelin.py
rfbgo: done!

douglasjacobsen (approved):

PR #1367, opened 2026-01-02
Allow ramble to add empty or missing repo objects [AI]

rfbgo:
This PR does two closely related things:

1. It makes it so if you add a repo that does not have all object types, ramble will continue to look there in the future incase they are added (eg via git pull). We retain the behavior that at least one repo type must be valid at initial add time, but it will add all possible object types
2. Lets `ramble repo rm` iterate through scopes to better find and remove things less explicitly

### DIFF
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -243,6 +243,8 @@ Ramble uses `pytest` for its unit tests. The tests can be run using a wrapper co
     ramble unit-test -k gromacs
     ```
 
+Prefer using `-k` if we know which tests we added.
+
 *   **Getting Help:**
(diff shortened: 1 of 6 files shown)


rfbgo:
This is failing `[codecov/patch](https://app.codecov.io/gh/GoogleCloudPlatform/ramble/commit/b20515ea66a09127cd9cfebc6ab6fdaaf643b22c)
codecov/patch — 90.00% of diff hit (target 91.48%)` but in my belief it would be reasonable to merge it anyway since the new code is tested

douglasjacobsen (approved):

PR #1373, opened 2026-01-13
Add a `make_workspace` pytest fixture

linsword13:
This aims to reduce duplications in tests, for creating a test workspace.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -715,7 +715,7 @@ def mock_file_auto_create(monkeypatch):
 
 
 @pytest.fixture
-def make_workspace(workspace_name, mutable_config, mutable_mock_workspace_path):
+def make_workspace_from_config(workspace_name, mutable_config, mutable_mock_workspace_path):
     """Fixture to create a workspace with a specific configuration."""
 
     def _create(config_str=None, name=None, activate=False):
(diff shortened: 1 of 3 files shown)


conftest.py
```diff
@@ -714,6 +714,28 @@ def open_or_create_inmem(path, *args, **kwargs):
     monkeypatch.setattr(builtins, "open", open_or_create_inmem)
 
 
+@pytest.fixture
+def make_workspace(workspace_name, mutable_config, mutable_mock_workspace_path):
```
douglasjacobsen: It might be good if this fixture was `make_workspace_from_config` since you are writing a static YAML file into it.

I could see a complementary `make_workspace_from_args` fixture where it internally calls `workspace manage experiments ...` to setup the configuration file as well.

douglasjacobsen (changes requested):
Thanks! This is a great idea. I have one comment, and then you can decide whether to address gemini's comment or not.

conftest.py
linsword13: Great call out! Updated the name to `make_workspace_from_config`. I do like your suggestion for a `from_args` fixture, but leaving that up for a future PR.

douglasjacobsen (approved):

PR #1382, opened 2026-01-20
Adds build system and project to pyproject.toml file

sarthakag:
Command `pip3 install .` sets up the project environment with the dependencies.


pyproject.toml
```diff
@@ -1,3 +1,30 @@
+[build-system]
+requires = ["setuptools>=61.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ramble"
+version = "0.6.0"
+description = "A multi-platform experimentation framework written in python."
+readme = "README.md"
+requires-python = ">=3.7"
```
linsword13: We do generally still try to support 3.6. Would this config cause issues for users with 3.6?

sarthakag: I took 3.7 requirement from the README, but happy to change it 3.6.

linsword13: The README does state a best-effort support for 3.6. So yeah let's use 3.6 if we can.

pyproject.toml
```diff
@@ -1,3 +1,30 @@
...
+classifiers = [
+    "Programming Language :: Python :: 3",
+]
+dynamic = ["dependencies"]
+
+[tool.setuptools_scm]
+# This enables git-based versioning.
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
```
linsword13, line 19: Another complication is that we have another `requirements-dev.txt` file that's used by Ramble developers. Is this something that can be accommodated by this setup?

sarthakag: This can be added:`optional-dependencies = {dev = {file = ["requirements-dev.txt"]}}` and then be installed via `pip install -e ".[dev]"`. 
However, I don't think it's required currently so will skip it in the PR.

pyproject.toml
```diff
@@ -1,3 +1,48 @@
+[build-system]
+requires = ["setuptools>=61.0", "wheel"]
```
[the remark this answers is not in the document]
sarthakag: Ramble does not update the git tags, so we will skip automatic versioning from git tags.

pyproject.toml
```diff
@@ -1,3 +1,48 @@
...
+version = "0.6.0"
+description = "A multi-platform experimentation framework written in python."
+readme = "README.md"
+requires-python = ">=3.6"
+license = "Apache-2.0 OR MIT"
+license-files = ["LICENSE-APACHE", "LICENSE-MIT"]
+classifiers = [
+    "Programming Language :: Python :: 3",
+]
+dynamic = ["dependencies"]
```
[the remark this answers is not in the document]
sarthakag: Same as above.

pyproject.toml
```diff
@@ -1,3 +1,48 @@
...
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
+
+[tool.setuptools]
+package-dir = { "" = "." }
+include-package-data = true
+
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["lib*", "etc*", "var*", "bin*", "share*"]
```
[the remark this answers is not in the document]
sarthakag: Current directory structure is not pip package-able. This is a workaround

pyproject.toml
```diff
@@ -1,3 +1,48 @@
...
+namespaces = true
+
+[tool.setuptools.package-data]
+"*" = ["*"]
```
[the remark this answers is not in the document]
sarthakag: Removing package-data removes the non python files and directories, which are required for ramble execution.

sarthakag:
Hi @linsword13, 

The PR is ready for review, PTAL and run the CICD if required. Thank you!

pyproject.toml
```diff
@@ -1,3 +1,41 @@
...
+
+[tool.setuptools.data-files]
```
linsword13: Total novice in terms of python packaging, but looking at the setuptools doc they have some strong words around the use of `data-files`: https://screenshot.googleplex.com/7nQcfCuAnPnVakk. Can you help me understand what this is for, and is the warning given by the setuptools doc applicable here?

sarthakag: We actually don't require data-files since we're adding the all the non python files as package data. Removed this block. Thank you!

douglasjacobsen (approved):

PR #1386, opened 2026-01-22
force conversion of data to int to match schema 

rfbgo:
This will give a more reliable error early

### DIFF
--- a/lib/ramble/ramble/test/uploader.py
+++ b/lib/ramble/ramble/test/uploader.py
@@ -67,6 +67,15 @@ def test_data_preparation(request, mock_applications):
         ap._prepare()
         ap._execute()
 
+        # Add default values for n_nodes, processes_per_node, and n_ranks
+        # as the zlib application does not provide them.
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/test/uploader.py
```diff
@@ -67,6 +67,15 @@ def test_data_preparation(request, mock_applications):
         ap._prepare()
         ap._execute()
 
+        # Add default values for n_nodes, processes_per_node, and n_ranks
```
linsword13: nit: can these values get supplied in the `manage experiments` call? Something like:

```
       workspace(
            "manage",
            "experiments",
            app_name,
            "-w",
            wl_name,
            "-p",
            "spack",
            "-v",
            "n_ranks=1",
            "-v",
            "n_nodes=1",
            "-v",
            "processes_per_node=1",
            global_args=global_args,
        )
```

rfbgo: That is a great idea, thankyou!

linsword13 (approved):

PR #1390, opened 2026-01-23
Stage default var values to allow precedence

douglasjacobsen:
This merge changes way missing variables are defined, by staging them into a default_variables dict. This allows more complex / higher precedence definitions to override lower precedence ones (when they come from object definitions).

Config file defined variables still override any object definitions.

### DIFF
--- a/lib/ramble/ramble/test/application.py
+++ b/lib/ramble/ramble/test/application.py
@@ -81,7 +81,7 @@ def test_basic_app(mutable_mock_apps_repo):
     example_input = basic_inst.workloads[_FS]["test_wl"].find_input("input")
     assert example_input is not None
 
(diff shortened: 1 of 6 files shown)


var/ramble/repos/builtin/applications/wrfv4/application.py
```diff
@@ -259,9 +259,9 @@ class Wrfv4(ExecutableApplication):
         "input_path",
         description="Path for workload inputs.",
         workload_defaults={
-            "CONUS_12km": "{CONUS_12km}",
```
linsword13, line 262: Are these left-overs from testing?

Also is there value to add some tests?

linsword13 (approved):

PR #1393, opened 2026-01-27
Add two tutorials for writing application definitions

douglasjacobsen:


douglasjacobsen:
The codecov failure is due to code that is only exercised when building the docs.

linsword13 (approved):

PR #1396, opened 2026-01-29
Fix modeless variable

douglasjacobsen:


lib/ramble/ramble/language/modifier_language.py
dapomeroy, line 304: This can be simplified by replacing L304-320 with `ramble.language.language_helpers.require_condition()` which takes the `mode`, `modes`, and `when` arguments and returns a when list. Example above in variable_modification(). It looks like you had implemented `when` here before I wrote that function and then I overlooked updating this directive to use it. Sorry about that!

dapomeroy (dismissed):

lib/ramble/ramble/language/modifier_language.py
douglasjacobsen: I think that is a slightly different implementation than what I have coded here though, so we should discuss this.

It looks like `require_condition` requires that either a mode (or modes) or a when argument is passed in, and it kind of assumes that you would set a mode through one of those two mechanisms. Granted, this is not always the case, and it looks like as long as you have a single when condition it would still give you a valid when. However, it also looks like the logic in `require_condition` for handing the merging of `mode` and `modes` is incorrect, as it takes these and builds a single when list (something like `when=["mod_mode=mode1", "mod_mode=mode2"]`) which would never be able to be satisfied. In this case we actually more want one when list per mode that was input.

Anyway, what I coded up here is more like if you don't pass in anything, still define the variable just without a mode (so it applies to all modes, instead of a single one). We can talk internally about how we want this to behave though, and then I can update this to follow that.

dapomeroy: OK, sounds good!

dapomeroy (approved):
LGTM

PR #1416, opened 2026-02-11
Improve experiment render performance

douglasjacobsen:
This merge performs several optimizations to improve performance of experiment rendering. It adds a new test (which is arguably slow) that renders a large number of experiments to help evaluate performance improvments.

Additionally, some minor improvements are made to the expander, and the Renderer, but some more significant performance improvements are made to the workspace info command, and the ExperimentSet objects.

Using the unit test in this PR locally on my laptop, the baseline performance is:
149.67s

And the final performance is:
25.38s

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -11,6 +11,7 @@ import itertools
 import os
 import sys
 import tempfile
+from collections import defaultdict
 from typing import Callable, Dict
 
 import llnl.util.tty as tty
@@ -794,8 +795,6 @@ def workspace_info(args):
     color.cprint(rucolor.section_title("Experiments:"))
 
     # Build an index of experiments to avoid re-rendering them in the loops below
-    from collections import defaultdict
-
     experiment_index_map = defaultdict(list)
     for exp_name, app_inst, _ in experiment_set.all_experiments():
         key = (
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -792,26 +792,30 @@ def workspace_info(args):
     all_pipelines = {}
     color.cprint("")
     color.cprint(rucolor.section_title("Experiments:"))
+
+    # Build an index of experiments to avoid re-rendering them in the loops below
+    from collections import defaultdict
```
linsword13: nit: move this import to the top of the file.

lib/ramble/ramble/renderer.py
```diff
@@ -439,30 +469,34 @@ def render_objects(self, render_group, exclude_where=None, ignore_used=True, fat
...
-                object_variables[var] = val
+
+            # Avoid mutating object_variables
+            # Use extra_vars for evaluation
 
             keep_object = True
             if exclude_where:
                 for where in exclude_where:
-                    evaluated = where_expander.expand_var(where)
+                    evaluated = where_expander.expand_var(where, extra_vars=obj)
```
linsword13: Would this subtly change any variable precedence?

linsword13 (commented):
This is great, some minor comments.

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Fixed!

lib/ramble/ramble/renderer.py
douglasjacobsen: I'm not sure. It seems like a weird change, and the tests pass without it, so I removed it.

linsword13 (approved):

PR #1422, opened 2026-02-20
Ensure active and non-existent spack env doesn't error

douglasjacobsen:
This merge fixes an edge case where an externally activated spack environment that no longer exists would prevent ramble from being able to create a new spack environment.

The fix here is to force ramble to ignore external activated environments.

### DIFF
--- a/var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
@@ -132,6 +132,14 @@ class SpackLightweight(PackageManagerBase):
 
         logger.msg("Creating Spack environment")
 
+        # Ignore externally active spack environment
+        if "SPACK_ENV" in os.environ and not os.path.isdir(
+            os.environ["SPACK_ENV"]
+        ):
+            _spack_vars = ["SPACK_ENV", "SPACK_ENV_VIEW"]
+            for var in _spack_vars:
+                os.environ.pop(var, None)
+
         # See if we cached this already, and if so return
         env_path = app_inst.expander.env_path
         if not env_path:


var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
```diff
@@ -132,6 +132,11 @@ def _software_create_env(self, workspace, app_inst=None):
 
         logger.msg("Creating Spack environment")
 
+        # Ignore externally active spack environment
```
linsword13, line 135: Would having an externally active env cause any issues? Mostly want to know if it's better to stop when this happens, or at least have a warning?

douglasjacobsen: It shouldn't really have any issues. Though, it's possible that you could have an external env that has compilers in it that you generally don't want to use.

Basically, loading an external env could change you `PATH` and `LD_LIBRARY_PATH` env-vars in a way that is non-standard for the rest of your config. However, erroring prevents that as a use-case (which could be entirely valid).

TBF, Ramble doesn't error if the externally active environment exists. So, I could actually just trap this on if the env doesn't exist instead of always purging the env-vars if that is preferrable.

A repo for the error is something like:

```
export SPACK_ENV=/tmp/__rmb_fake_dir
ramble workspace setup
```

If `SPACK_ENV` points to a real spack environment, then ramble will work like it should.

linsword13: Nested envs ftw! I kinda like your idea of purging only if it doesn't point to a valid directory.

douglasjacobsen: Done!

linsword13 (approved):

PR #1423, opened 2026-02-20
Fix repeat experiments in `workspace info`

douglasjacobsen:
This merge fixes an issue introduced in 20f1ae9be5997318a82ae054983f7a32e3d6f6dc that prevented child experiments of a repeat base from showing up in `workspace info` output.

Additionally, a test was added to ensure this behavior remains fixed, and two additional keyword variables are added:
 - is_repeat_parent
 - is_repeat_child

Which can be used to filter out parent / children experiments when repeats are included in a workspace.

### DIFF
--- a/lib/ramble/ramble/experiment_set.py
+++ b/lib/ramble/ramble/experiment_set.py
@@ -266,11 +266,13 @@ class ExperimentSet:
 
         experiment_suffix = ""
         # After generating the base experiment, append the index to repeat experiments
+        variables[self.keywords.is_repeat_parent] = repeats.is_repeat_base
+        variables[self.keywords.is_repeat_child] = False
+        variables[self.keywords.repeat_index] = 0
         if repeats.repeat_index:
             experiment_suffix = f".{repeats.repeat_index}"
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/experiment_set.py
```diff
@@ -266,11 +266,13 @@ def _prepare_experiment(
 
         experiment_suffix = ""
         # After generating the base experiment, append the index to repeat experiments
+        variables[self.keywords.is_repeat_parent] = repeats.is_repeat_base
+        variables[self.keywords.is_repeat_child] = False
```
linsword13, line 270: How are these two vars being used?

douglasjacobsen: They are not used internally. But they are defined so users can filter with `--where`

Things like:

```
ramble workspace analyze --where '{is_repeat_parent}'
```

linsword13: I see, so basically serving as a better way than using `repeat_index`.

linsword13 (approved):

PR #1432, opened 2026-02-27
Fixes issue with version variable expansion

dapomeroy:
Fixes issue with version variable not expanding correctly in `info`

#aadd

### DIFF
--- a/lib/ramble/ramble/experiment_set.py
+++ b/lib/ramble/ramble/experiment_set.py
@@ -84,7 +84,7 @@ class ExperimentSet:
         # Set some base variables from the workspace definition.
         self.set_base_var(self.keywords.log_dir, workspace.log_dir)
         self.set_base_var(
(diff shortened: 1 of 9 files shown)


lib/ramble/ramble/experiment_set.py
```diff
@@ -213,7 +213,6 @@ def _setup_experiment_minimal(
         # Define some standard variables before the application is created
         # to ensure variables and variants can be defined correctly.
         variables[self.keywords.workload_template_name] = workload_template_name
-        variables[self.keywords.application_name] = final_app_name
```
douglasjacobsen, line 216: If we remove this variable, is it still accessible to objects / workspaces that use it?

dapomeroy: Per our discussion, decided to create a new variable `application_spec` and have that be the one that goes into namespace variables. `application_name` will continue to be just the name of the app object in the repo

douglasjacobsen (approved):

PR #1434, opened 2026-02-27
Add scoped version variables to objects

douglasjacobsen:
This merge handles several things with respect to versions. Primarily, it add scoped version variables in the form of ``{object_type}::{object_name}::version`` to replace the ``{object_type}_version`` variables. For example ``application::wrf::version``.

Modifiers will also now conflict when two modifiers have the same name, but different version strings.

Additionally, it fixes the following issues:
 - Modifiers were not previously allowed (they would error with a variable definition call)
 - `manage modifiers` wouldn't add versioned modifiers
 - Experiments could define version variables

This also updates documentation, and adds tests.


dapomeroy (approved):
LGTM

PR #1440, opened 2026-03-02
Add `fom-origin-types` to filter FOMs analyze output

linsword13:
The filtering applies to analyze output, but not to tables, since tables already can select FOMs flexibly.

The filtering is done when dumping the output instead of during FOM definitions. Otherwise it may cause issues when FOMs are used as variables in certain cases.

### DIFF
--- a/lib/ramble/ramble/cmd/workspace.py
+++ b/lib/ramble/ramble/cmd/workspace.py
@@ -545,7 +545,9 @@ def workspace_analyze_setup_parser(subparser):
         "--fom-origin-types",
         dest="fom_origin_types",
         nargs="+",
-        help="list of FOM origin types to include in analysis",
+        action="append",
+        help="FOM origin types to include in analysis output. "
+        + "Accepts space-delimited lists and can be specified multiple times.",
         required=False,
     )
 
@@ -582,6 +584,11 @@ def workspace_analyze(args):
         tags=args.filter_tags,
     )
 
+    if args.fom_origin_types:
+        fom_origin_types = [item for sublist in args.fom_origin_types for item in sublist]
+    else:
+        fom_origin_types = None
+
     pipeline_cls = ramble.pipeline.pipeline_class(current_pipeline)
 
     logger.debug("Analyzing workspace")
@@ -592,7 +599,7 @@ def workspace_analyze(args):
         upload=args.upload,
         print_results=args.print_results,
         summary_only=args.summary_only,
-        fom_origin_types=args.fom_origin_types,
+        fom_origin_types=fom_origin_types,
     )
 
     with ws.read_transaction():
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -2763,13 +2767,31 @@ def no_active_workspace():
...
+            filtered_contexts = []
+            for context in r.get("CONTEXTS", []):
+                filtered_foms = []
+                for fom in context.get("foms", []):
+                    if fom.get("origin_type") in fom_origin_types:
+                        filtered_foms.append(fom)
+                if filtered_foms:
+                    context["foms"] = filtered_foms
+                    filtered_contexts.append(context)
+            r["CONTEXTS"] = filtered_contexts
```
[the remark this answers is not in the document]
linsword13: Skipping this suggestion for now. The comprehension code ends up being quite long and IMHO doesn't help with readability much.

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -541,6 +541,14 @@ def workspace_analyze_setup_parser(subparser):
         help="print out the analysis result",
     )
 
+    subparser.add_argument(
+        "--fom-origin-types",
+        dest="fom_origin_types",
+        nargs="+",
+        help="list of FOM origin types to include in analysis",
```
douglasjacobsen: This is a space delimited list, right? If so, can we clarify that in the docs, since we have support for space delimited lists and comma delimited lists in various arguments.

douglasjacobsen (changes requested):

lib/ramble/ramble/cmd/workspace.py
linsword13: Great point, done. I also added the `action="append"` to allow for specifying it multiple times, similar to how `--where` works.

lib/ramble/ramble/test/end_to_end/analyze_fom_output.py
```diff
@@ -134,3 +134,74 @@ def test_analyze_fail_with_no_fom_detected(mock_applications, workspace_name):
     with open(result_file) as f:
         content = f.read()
         assert "Status = FAILED" in content
+
+
+def test_analyze_fom_origin_types_filter(mock_applications, make_workspace_from_config):
```
douglasjacobsen, line 139: Really annoying nit;

Can you add a modifier in this test, and assert that the modifier's FOMs are not in the filtered output?

douglasjacobsen (changes requested):

lib/ramble/ramble/test/end_to_end/analyze_fom_output.py
douglasjacobsen: Actually... maybe this isn't needed with the `foo` filter. Scratch my request.

douglasjacobsen (approved):

PR #1441, opened 2026-03-03
Update tutorials and repo app versions

dapomeroy:
Revise docs and tutorials with newer versions of example software and updated syntax. Update versions of various packages in repo.

#aadd


lib/ramble/ramble/test/end_to_end/manage_software.py
```diff
@@ -43,42 +43,42 @@ def test_manage_software(mutable_config, mutable_mock_workspace_path, workspace_
             content = f.read()
             # Check that wrf has a package, and the package is in an environment
             assert "pkg_spec: wrf" in content
-            assert "- wrfv4" in content
+            assert "- wrfv4-{application::wrf::version}" in content
```
douglasjacobsen, line 46: Does this actually work? I would expect to see `wrf-{...}` instead of `wrfv4-{...}` here.

dapomeroy: I defined the software spec in the wrf app definition with the names 'wrfv3' and 'wrfv4' to avoid software conflict errors, so this is correct. https://github.com/GoogleCloudPlatform/ramble/pull/1441/changes#diff-d046911539ff45784fc1a2fecdcd1c97121169af9dc99c5bce3772e358cc794dL40

Let me know if you want me to look at refactoring conflict detection so that we can unify on a 'wrf' spec.

douglasjacobsen: Ah OK, thanks!

douglasjacobsen (approved):

PR #1442, opened 2026-03-03
Fix filtering of modifier names

douglasjacobsen:
This merge fixes an issue where wildcard characters were added to filter modifiers when calling `add_modifier` in the workspace class.

This is updated to use `fnmatch.filter` instead, as it will not force wildcards on modifier names that shouldn't have them.

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -40,7 +40,6 @@ import ramble.util.path
 import ramble.util.version
 from ramble.mirror import MirrorStats
 from ramble.namespace import namespace
-from ramble.util import object_utils
 from ramble.util.conversions import list_str_to_list
 from ramble.util.logger import logger
 from ramble.util.path import substitute_path_variables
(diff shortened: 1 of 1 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -2440,9 +2439,10 @@ def add_modifier(
             base_section[namespace.modifiers] = syaml.syaml_list()
 
         mod_type = ramble.repository.ObjectTypes.modifiers
+        mod_objects = ramble.repository.all_object_names(object_type=mod_type)
         if isinstance(name_pattern, str):
             spec_parts = name_pattern.partition("@")
-        mod_names = object_utils.filter_by_name([spec_parts[0]], False, mod_type)
+        mod_names = fnmatch.filter(mod_objects, spec_parts[0])
```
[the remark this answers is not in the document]
douglasjacobsen: Thanks! Done.

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -2440,9 +2439,14 @@ def add_modifier(
             base_section[namespace.modifiers] = syaml.syaml_list()
 
         mod_type = ramble.repository.ObjectTypes.modifiers
+        mod_objects = ramble.repository.all_object_names(object_type=mod_type)
         if isinstance(name_pattern, str):
             spec_parts = name_pattern.partition("@")
-        mod_names = object_utils.filter_by_name([spec_parts[0]], False, mod_type)
```
linsword13, line 2445: Are there other users of this utility that suffer from the same issue? Mostly wonder if we want to fix this utility instead?

douglasjacobsen: The other usages of this utility are mostly front-end commands (i.e. `ramble list`) and that usage assumes this behavior is true.

Basically: `ramble list h` is "expected" to currently give all application definitions with an h in their name.

linsword13 (approved):

PR #1445, opened 2026-03-04
Ensure variables that match reserved patterns are removed

douglasjacobsen:
This merge updates the `non_reserved_variables` method to remove any variable that is either reserved, or matches a reserved pattern.


linsword13 (commented):
LGTM. One nit: is there test that can be added to show the scenario the change covers?

douglasjacobsen:
@linsword13 Done. :)

linsword13 (approved):

PR #1455, opened 2026-03-10
Add SQLite uploader [AI] 

rfbgo:


lib/ramble/ramble/uploader.py
```diff
@@ -624,3 +647,150 @@ def perform_upload(self, uri, results):
...
+        sqlite_schema = []
+        for name, props in schema.get("properties", {}).items():
+            sqlite_type = type_map[props["type"]]
+            sqlite_schema.append(f"{name} {sqlite_type}")
+
+        return ", ".join(sqlite_schema)
+
+    def create_tables(self, uri):
+        import os
+        import sqlite3
```
linsword13, line 683: nit: perhaps better to put these imports at module-level instead of on-demand?

lib/ramble/ramble/uploader.py
```diff
@@ -624,3 +647,150 @@ def perform_upload(self, uri, results):
...
+            if cursor.fetchone()[0] == 1:
+                logger.info(f"Table {table_name} already exists.")
+            else:
+                logger.info(f"Creating table {table_name}")
+                sqlite_schema = self._schema_to_sqlite(table_def["schema"][table_def["version"]])
+                cursor.execute(f"CREATE TABLE {table_name} ({sqlite_schema})")
+                tables_created = True
+
+        conn.commit()
+        conn.close()
```
linsword13: Would it be safer to have these cleanup ops inside some try-except-final block, to ensure they get called?

douglasjacobsen (dismissed):

douglasjacobsen (dismissed):

linsword13 (commented):
Sorry forgot to send out my review comments...

lib/ramble/ramble/uploader.py
rfbgo: It's a good question, and I'm not 100% sure. My logic here for `sqlite3` was that I wanted to localize it to where it is needed, on the off chance we are missing the (system level) deps 

I will move `os`, good catch

lib/ramble/ramble/uploader.py
rfbgo: Sure, that's a good idea. I think in practice it is pretty safe already, but your way is even better

lib/ramble/ramble/uploader.py
linsword13: Makes sense. And now I see that we do the same for bigquery. LGTM!

linsword13 (approved):

PR #1460, opened 2026-03-11
Improve expander performance

dapomeroy:
Improves performance of expander by:
* Updating heuristic to avoid parsing most identifiers
* Moving _ast_parse out of try/warn blocks to reduce overhead, including logger.debug calls, for strings that fail to parse as math. Errors during evaluation of math are still logged.


dapomeroy:
Per team discussion, removed instance-level cache from this PR to focus only on the parsing changes. Caching will be considered in a future design.

linsword13 (approved):

douglasjacobsen (approved):

PR #1468, opened 2026-03-23
Convert string formatting into f-strings

linsword13:
The changes are generated using `flynt`:

```
flynt lib/ramble/ramble var/ramble/repos/builtin/base_classes/ && ramble style -f
```

Additionally, add in a couple tests to cover some of the changes. Also remove an unused method in stage.py.

### DIFF
--- a/lib/ramble/ramble/pipeline.py
+++ b/lib/ramble/ramble/pipeline.py
@@ -483,17 +483,17 @@ class MirrorPipeline(Pipeline):
         logger.msg(
             f"Successfully {verb} spack software in {self.workspace.mirror_path}",
             "Archive stats:",
-            "  %-4d already present" % len(self.workspace.software_mirror_stats.present),
-            "  %-4d added" % len(self.workspace.software_mirror_stats.new),
-            "  %-4d failed to fetch." % len(self.workspace.software_mirror_stats.errors),
+            f"  {len(self.workspace.software_mirror_stats.present):<4} already present",
+            f"  {len(self.workspace.software_mirror_stats.new):<4} added",
+            f"  {len(self.workspace.software_mirror_stats.errors):<4} failed to fetch.",
         )
 
         logger.msg(
             f"Successfully {verb} inputs in {self.workspace.mirror_path}",
             "Archive stats:",
-            "  %-4d already present" % len(self.workspace.input_mirror_stats.present),
-            "  %-4d added" % len(self.workspace.input_mirror_stats.new),
-            "  %-4d failed to fetch." % len(self.workspace.input_mirror_stats.errors),
+            f"  {len(self.workspace.input_mirror_stats.present):<4} already present",
+            f"  {len(self.workspace.input_mirror_stats.new):<4} added",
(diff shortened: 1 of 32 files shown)


lib/ramble/ramble/pipeline.py
```diff
@@ -483,17 +483,17 @@ def _complete(self):
         logger.msg(
             f"Successfully {verb} spack software in {self.workspace.mirror_path}",
             "Archive stats:",
-            "  %-4d already present" % len(self.workspace.software_mirror_stats.present),
-            "  %-4d added" % len(self.workspace.software_mirror_stats.new),
-            "  %-4d failed to fetch." % len(self.workspace.software_mirror_stats.errors),
+            f"  {len(self.workspace.software_mirror_stats.present):-4} already present",
+            f"  {len(self.workspace.software_mirror_stats.new):-4} added",
+            f"  {len(self.workspace.software_mirror_stats.errors):-4} failed to fetch.",
```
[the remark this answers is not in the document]
linsword13: That's a great catch, updated.

lib/ramble/ramble/pipeline.py
```diff
@@ -483,17 +483,17 @@ def _complete(self):
...
 
         logger.msg(
             f"Successfully {verb} inputs in {self.workspace.mirror_path}",
             "Archive stats:",
-            "  %-4d already present" % len(self.workspace.input_mirror_stats.present),
-            "  %-4d added" % len(self.workspace.input_mirror_stats.new),
-            "  %-4d failed to fetch." % len(self.workspace.input_mirror_stats.errors),
+            f"  {len(self.workspace.input_mirror_stats.present):-4} already present",
+            f"  {len(self.workspace.input_mirror_stats.new):-4} added",
+            f"  {len(self.workspace.input_mirror_stats.errors):-4} failed to fetch.",
```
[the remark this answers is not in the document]
linsword13: Fixed.

lib/ramble/ramble/stage.py
```diff
@@ -809,113 +807,6 @@ def ensure_access(file):
         logger.die(f"Insufficient permissions for {file}")
 
 
-# TODO (dwj): Need to add checksums for inputs.
```
douglasjacobsen, line 812: This doesn't seem like changing f-strings. I'm kind of surprised this doesn't cause errors, but maybe we don't check downloading in our unit tests?

linsword13: Yeah the logger.msg calls inside it were converted to f-strings. But then it seems this method isn't actually used right now (is this true?) Therefore I went ahead and removed it.

douglasjacobsen: I guess I'm not sure if it's used right now or not. We definitely validate the SHA on inputs. But that never happens if we run a dry-run. Almost all of our unit tests only performs a dry-run.

It's also possible that the fetch strategy handles the SHA check rather than the stage though.

douglasjacobsen (approved):

lib/ramble/ramble/stage.py
linsword13: Discussed offline to ensure it's not being used.

PR #1470, opened 2026-03-25
Remove unused code

linsword13:
Initially I tried using `vulture`, but that didn't work well at all and kept focusing on the `appkit` and other import-only modules. Instead I asked Gemini to come up with an ast parsing script, to filter out functions in the lib that are defined, but not called. After having that list, I grepped to ensure these are indeed not being used (as opposed to being called dynamically using `getattr` or the like.)

One potential danger is if any downstream projects referencing any of these (which are not used by us,) but I considered that probably quite low. And I think it's better that we clean these unused signatures up now rather than later.


lib/ramble/ramble/definitions/families.py
```diff
@@ -51,8 +51,3 @@ def as_str(self, n_indent: int = 0, **kwargs):
             out_str += f"{indentation}{self.family_type}={family}\n"
 
         return out_str
-
-    def add_families(self, families: List[str]):
```
douglasjacobsen, line 55: I would prefer to not remove this code if possible.

lib/ramble/ramble/definitions/versions.py
```diff
@@ -105,11 +105,6 @@ def get_version_num(self):
 
         return self.version_number
 
-    def evaluate_conflicts(self, variant):
```
douglasjacobsen, line 108: @dapomeroy: Is the plan with this code to build it out for conflicts? (I think we have a task to fill this in)

douglasjacobsen (changes requested):

lib/ramble/ramble/definitions/versions.py
dapomeroy: I have a task to fill this in. I don't really have a use case for it yet, so I'm not sure if it's worth doing at the moment.

lib/ramble/ramble/definitions/families.py
linsword13: Definitely, reverted.

lib/ramble/ramble/definitions/versions.py
linsword13: No problem, keeping this in.

douglasjacobsen (approved):

PR #1486, opened 2026-04-02
Update HPL default config for calculator

douglasjacobsen:


var/ramble/repos/builtin/base_applications/hpl/base_application.py
```diff
@@ -269,6 +269,43 @@ class Hpl(ExecutableApplication):
         workload_group="calculator",
     )
 
+    workload_variable(
+        "NPFACTs",
```
linsword13, line 273: I thought the values of these variables, for calculator, comes from the `_calculate_values` function from the base app?

douglasjacobsen: Kind of, but not really. `_calculate_values` only really calculates things like P, Q, N, and NB. The rest are just read as their default values from the standard workload (if they are not set, see https://github.com/GoogleCloudPlatform/ramble/pull/1486/changes#diff-51a2c4995aab35d193f80a7ffcfba4d2d989af042bad40fc77d17421843f7b08L401)

We commonly change these (to the values in this PR) to get the best performance. So, I'm just encoding what we use as defaults in the calculator workload so we don't use the standard workloads settings here.

linsword13: I see, makes sense. Does that mean we can now remove this part where the calculator workload tries to fetch values from the standard? https://github.com/douglasjacobsen/ramble/blob/develop/var/ramble/repos/builtin/base_applications/hpl/base_application.py#L401-L409

douglasjacobsen: No, that part still pulls in other defaults from standard. Standard has the full config definition, while calculator only has part. This portion makes sure all default variables are defiend in the selected workload (which can be useful for other workloads we add in the future as well).


More specifically, the section you linked to pulls in all of these variable: https://github.com/douglasjacobsen/ramble/blob/develop/var/ramble/repos/builtin/base_applications/hpl/base_application.py#L298-L328

linsword13: Got it, thanks.

linsword13 (approved):

PR #1488, opened 2026-04-03
Update color escaping logic to be more maintainable [AI]

douglasjacobsen:
This merge updates the logic used for escaping strings prior to printing out colored output. `@` symbols need to be escaped when they are to be printed as is, by converting them to `@@` entries instead. The color utility is updated to provide wrappers that can only escape strings that are not used for coloring directly (since color markers also include `@`).

### DIFF
--- a/lib/ramble/ramble/cmd/common/info.py
+++ b/lib/ramble/ramble/cmd/common/info.py
@@ -136,7 +136,7 @@ def print_object_header(obj_type, obj):
     singular = ramble.repository.type_definitions[obj_type]["singular"]
     parts = [part[0].upper() + part[1:] for part in singular.split()]
     type_name = color.section_title(" ".join(parts))
-    color.cprint(f"{type_name}: %s\n", obj.name)
+    color.cprint(f"{type_name}: {obj.name}\n")
 
     color.cprint(color.section_title("Description:"))
     if obj.__doc__:
@@ -144,7 +144,7 @@ def print_object_header(obj_type, obj):
         for part in obj.__doc__.split("\n"):
             doc_str += f"    {part}\n"
 
-        color.cprint("%s", doc_str)
+        color.cprint(f"{doc_str}")
 
 
 def print_object_overview(obj):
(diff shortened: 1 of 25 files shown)


lib/ramble/ramble/cmd/common/info.py
```diff
@@ -136,16 +135,16 @@ def print_object_header(obj_type, obj):
     """Print an object header"""
     singular = ramble.repository.type_definitions[obj_type]["singular"]
     parts = [part[0].upper() + part[1:] for part in singular.split()]
-    type_name = rucolor.section_title(" ".join(parts))
-    color.cprint(f"{type_name}: {obj.name}\n")
+    type_name = color.section_title(" ".join(parts))
+    color.cprint(f"{type_name}: %s\n", obj.name)
```
dapomeroy: Is there a reason this uses both f-string and printf style formatting? Similar question for the rest of the PR - Is there a reason to use printf-style formats or can we use f-string instead?

lib/ramble/ramble/test/util/colors.py
```diff
@@ -0,0 +1,53 @@
...
+        ("@*R{text}", "@*R{text}"),
+        ("@*r{text}", "@*r{text}"),
+        # Complex cases
+        ("Error at @r{@*line 10}@. in @*file.py", "Error at @r{@*line 10}@. in @*file.py"),
+        ("Multiple @@ characters", "Multiple @@ characters"),
+        # Missing closing brace: @r is matched by itself
+        ("Missing brace: @r{text", "Missing brace: @r{text"),
+    ],
+)
+def test_auto_escape(input_str, expected):
```
dapomeroy, line 54: I noticed that some of the 'plaintext' color formats have been removed, while others remain. Is the presence of 'foo@bar' in this test an indication that auto escaping now works for software versions? If so, can we remove the explicit 'plaintext' function and pass strings containing version syntax into cprint without it?

dapomeroy (changes requested):

lib/ramble/ramble/cmd/common/info.py
douglasjacobsen: Yeah, Gemini made several comments that the f-string style syntax was not safe for referring to attributes of obj, and using the `%s` style would handle cases where these are None better.

lib/ramble/ramble/test/util/colors.py
douglasjacobsen: Good call. This should be resolved in the most recent push now.

dapomeroy (approved):

PR #1490, opened 2026-04-03
make workspace dir an array

rfbgo:
I wanted to beef up workspace dir support so we can normalize having multiple directory 

To add a directory: ramble config add config:workspace_dirs:/path/to/new/dir
To remove a directory: ramble config remove config:workspace_dirs:/path/to/dir

User added and file order gives precedence for any colliding names

### DIFF
--- a/etc/ramble/defaults/config.yaml
+++ b/etc/ramble/defaults/config.yaml
@@ -21,7 +21,8 @@ config:
   checksum: true
   shell: bash
   input_cache: $ramble/var/ramble/cache
-  workspace_dirs: $ramble/var/ramble/workspaces
+  workspace_dirs:
+    - $ramble/var/ramble/workspaces
   report_dirs: ~/.ramble/reports
   include_phase_dependencies: false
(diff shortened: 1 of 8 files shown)


douglasjacobsen:
I don't think this actually works how I would expect right now. Adding and removing a workspace dir definitely works correctly, except that when you add a new workspace dir `ramble workspace list` only lists workspaces in the last workspace dir you added (and hides any from prior workspace directories). So it's almost more like a stack of workspace dirs where you can only look at the top of the stack.

Some other things that woudl be nice in this though:
1) Users might want the ability to control which dir new workspaces are created in
2) Users might want a way to filter the workspaces they are looking at based on their dir (or maybe `workspace list` should group by which dir they are in)

douglasjacobsen (changes requested):
(I should have put my last comment in this block... but refer to that for what I think should change..)

rfbgo:
I agree with your feedback on selecting where things are created, great point

PR #1497, opened 2026-04-07
Handle impossible when conditions in application definitions [AI]

douglasjacobsen:
This merge adds logic to detect and handle impossible when conditions in Ramble application definitions. It includes:
- Logic to check if a 'when' condition is self-contradictory.
- Warnings when entering an impossible 'with when' context.
- Logic to skip directives with impossible 'when' conditions.
- Logic to only add variables/env vars to workloads if their 'when' conditions are compatible with the workload's 'when' condition.
- A new test file 'lib/ramble/ramble/test/impossible_when.py' to verify the fix.

### DIFF
--- a/lib/ramble/ramble/language/language_base.py
+++ b/lib/ramble/ramble/language/language_base.py
@@ -18,13 +18,33 @@ from collections.abc import Sequence  # novm
 from typing import Any, Callable, Dict, List, Set
 
 import llnl.util.lang
-import llnl.util.tty as tty
 
 import ramble.error
 import ramble.language.language_helpers
 from ramble.error import DirectiveError
+from ramble.util.logger import logger
 
 __all__ = ["DirectiveMeta", "DirectiveError"]
+
+
+def _impossible_when_warning(directive_name, obj_type, obj_name, message, args, kwargs):
+    arg_lines = ["Directive Arguments:"] + [f" - {arg}" for arg in args] if args else []
+    kwarg_lines = (
+        ["Directive Keyword Arguments:"] + [f"  {k} = {v}" for k, v in kwargs.items()]
+        if kwargs
+        else []
+    )
+
+    parts = [
+        f'Directive "{directive_name}"'
+        f'{f" in {obj_type} {obj_name}" if obj_name and obj_type else ""} '
+        "has an impossible when condition:",
+        message,
+        *arg_lines,
+        *kwarg_lines,
+    ]
+
+    logger.warn("\n".join(parts))
 
 
 #: These are variant names used by ramble internally; applications can't use
@@ -50,7 +70,7 @@ def _push_to_context(when_condition: str) -> None:
         DirectiveMeta._when_constraints_from_context
     )
     if impossible:
-        tty.warn(f"Entering an impossible 'when' context: {message}")
+        logger.warn(f"Entering an impossible 'when' context: {message}")
 
 
 def _pop_from_context() -> str:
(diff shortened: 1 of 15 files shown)


lib/ramble/ramble/language/language_base.py
```diff
@@ -43,6 +46,12 @@
 def _push_to_context(when_condition: str) -> None:
     DirectiveMeta._when_constraints_from_context.append(when_condition)
 
+    impossible, message = ramble.language.language_helpers.is_when_impossible(
+        DirectiveMeta._when_constraints_from_context
+    )
+    if impossible:
+        tty.warn(f"Entering an impossible 'when' context: {message}")
```
dapomeroy: Should this be using logger instead of tty? We use logger in shared language

lib/ramble/ramble/language/language_base.py
```diff
@@ -247,6 +256,17 @@ def _wrapper(*args, **_kwargs):
 
                     kwargs["when"] = when_constraints.copy()
 
+                if "when" in kwargs:
+                    impossible, message = ramble.language.language_helpers.is_when_impossible(
+                        kwargs["when"]
+                    )
+                    if impossible:
+                        tty.warn(
```
dapomeroy: logger?

lib/ramble/ramble/language/language_base.py
```diff
@@ -247,6 +256,17 @@ def _wrapper(*args, **_kwargs):
...
+                            f'Directive "{decorated_function.__name__}" has '
```
dapomeroy: I tried this out with the impossible_when mock app, and I find this warning a little hard to understand. I think it'd be better if it had the object name in addition to the directive name (In this case, "impossible_variable"), if it's possible to access that here in DirectiveMeta.

For example, in the mock app, the warning appears like this:
`==> Warning: Directive "workload_variable" has an impossible when condition: variant 'workload_versions' has conflicting values: 'ver1' and 'ver2'`

Referring to this block in the app definition:
```    with when("workload_versions=ver1"):
        workload(
            "test_wl",
            executables=["base_exec", "test_exec"],
        )
       ...

        workload_variable(
            "impossible_variable",
            default="Foo",
            description="This is an impossible variable definition",
            workload="test_wl",
            when="workload_versions=ver2",
        )
```

Since it only says 'workload_variable' in the warning, we don't know what variable definition it's referring to and need to go look for it. If this had a longer list of variables, and it wasn't named 'impossible_variable', it might be harder for the app developer to find and fix the problematic when condition.

dapomeroy (commented):
This looks great! A couple of nits but I think it's exactly what we needed here.

lib/ramble/ramble/language/language_base.py
douglasjacobsen: Thanks, that's a good suggestion. How does this look?

```
==> Warning: Directive "workload_variable" in application impossible-when has an impossible when condition:
variant 'workload_versions' has conflicting values: 'ver1' and 'ver2'
Directive Arguments:
 - impossible_variable
Directive Keyword Arguments:
  default = Foo
  description = This is an impossible variable definition
  workload = test_wl
  when = ['workload_versions=ver1', 'workload_versions=ver2']

  ```

lib/ramble/ramble/language/language_base.py
douglasjacobsen: Good catch, updated!

lib/ramble/ramble/language/language_base.py
douglasjacobsen: Good catch, updated!

lib/ramble/ramble/language/language_base.py
dapomeroy: That's great!

dapomeroy (approved):
LGTM

PR #1505, opened 2026-04-14
Fix up a couple tests affected by class var change

linsword13:


rfbgo:
Thanks for catching that! Any idea why they didn't start failing?

linsword13:
> Thanks for catching that! Any idea why they didn't start failing?

Yeah I noticed this in the unit tests trigger that runs upon code merge: https://github.com/GoogleCloudPlatform/ramble/runs/71129657051. I think it didn't get caught because one of the changes got merged earlier and the inmem_fom one didn't get that new change when running its unit tests in the PR.

rfbgo (approved):

PR #1514, opened 2026-04-20
Add a pre-post-capture modifier

douglasjacobsen:
This merge adds a new modifier that can be used to inject command captures before and after portions of each experiment.

### DIFF
--- a/var/ramble/repos/builtin/modifiers/pre-post-capture/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/pre-post-capture/modifier.py
@@ -1,0 +1,163 @@
+# Copyright 2022-2026 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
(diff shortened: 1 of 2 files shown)


var/ramble/repos/builtin/modifiers/pre-post-capture/modifier.py
```diff
@@ -0,0 +1,162 @@
...
+    tags("debugging")
+
+    mode("common", description="Mode for collecting common debugging captures")
+
+    default_mode("common")
+
+    modifier_variable(
+        "pre_post_capture_exec_pattern",
+        default="all_executables",
+        description="Refex pattern to control which executables to capture. If "
```
linsword13: nit: "Regex"? Also might be worth calling out that "all_executables" is a special keyword?

douglasjacobsen: Should be resolved now. Good catch. :)

linsword13 (approved):

PR #1516, opened 2026-04-21
Add a collection daemon modifier

douglasjacobsen:
This merge adds a modifier that starts a daemon to periodically collect the output of some given commands, and store the output in a file.

### DIFF
--- a/var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_start.sh.tpl
+++ b/var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_start.sh.tpl
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+TEST_NAME="$1"
+CMD="$2"
+SLEEP="$3"
+cd {experiment_run_dir}
+
+COLLECTION_DIR=$(dirname {collection_daemon_log_prefix})
+
+NODE_IDX=$(awk -v h="$(hostname)" '$0 == h { print NR; exit }' {hostfile})
+NODE_IDX=$(printf "%03d" $NODE_IDX)
+
+DAEMON_PID=$$
+END_FILE=$COLLECTION_DIR/.end_collection
+PID_FILE=$COLLECTION_DIR/pid-values.node-$NODE_IDX
+
+echo "$DAEMON_PID" >> $PID_FILE
+
+while  [ ! -f $END_FILE ]; do
+  bash -c "$CMD >> {collection_daemon_log_prefix}.daemon-$TEST_NAME.node-$NODE_IDX 2>&1"
+  sleep $SLEEP
+done
--- a/var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_kill_all_daemons.sh.tpl
+++ b/var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_kill_all_daemons.sh.tpl
@@ -1,0 +1,17 @@
+#!/bin/bash
+cd {experiment_run_dir}
+COLLECTION_DIR=$(dirname {collection_daemon_log_prefix})
+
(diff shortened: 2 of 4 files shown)


var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_kill.sh.tpl
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+touch $(dirname {collection_daemon_log_prefix})/.end_collection
+pdsh -R ssh -w {hostlist} "pkill -f col*dae*start.sh"
```
linsword13: Would it be safer to spell out "collection_daemon_start.sh"?

var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_start.sh.tpl
```diff
@@ -0,0 +1,17 @@
...
+SLEEP="$3"
+cd {experiment_run_dir}
+
+NODE_IDX=$(awk -v h="$(hostname)" '$0 == h { print NR; exit }' {hostfile})
+NODE_IDX=$(printf "%03d" $NODE_IDX)
+
+END_FILE=$(dirname {collection_daemon_log_prefix})/.end_collection
+
+while  [ ! -f $END_FILE ]; do
+  bash -c "$CMD >> {collection_daemon_log_prefix}.daemon-$TEST_NAME.node-$NODE_IDX 2>&1"
```
linsword13, line 20: Is the log_prefix a directory?

var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_kill.sh.tpl
douglasjacobsen: Pkill actually compains if you do that. It says "Any pattern with more than 15 characters will have 0 matches"

```
The process name used for matching is limited to the 15 characters present in the output of /proc/pid/stat. Use the -f option to match against the complete command line, /proc/pid/cmdline.
```
From https://linux.die.net/man/1/pkill

We could probably write a helper that we could execute under `pdsh` like we do when we start the daemon to kill all of the PIDs directly though.

But for `pkill` I'm not really sure what the best course is.

var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_start.sh.tpl
douglasjacobsen: No, the log prefix is a full path to a prefix on the log file. But I'm grouping the output files all into the same directory (i.e. `$(dirname {collection_prefix})`) so they are organized and I know where to grab them in the future.

var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_kill.sh.tpl
linsword13: Oh, I didn't know about that character limit, TIL!

Thinking about it, perhaps using PID is safer? The hypothetical scenario is if we have concurrent experiments running on the same node (and they all use this modifier.) Then `pkill` might unintentionally terminates the collection of the other experiments?

var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_start.sh.tpl
linsword13: I see, makes sense!

var/ramble/repos/builtin/modifiers/collection-daemon/collection_daemon_kill.sh.tpl
douglasjacobsen: OK, PID based termination should be added now.

linsword13 (approved):

PR #1517, opened 2026-04-21
Allow Workspace Dir to be an array

rfbgo:
This PR converts workspace dir to be an array, while maintaining backwards compatibility

It also adds explciit args to allow users to target workspace commands at the  dir of their choosing 

(note: when testing this if you have a none-array in your workspace-dir config already it should be changed to an array format)


douglasjacobsen:
I can't get the `--parent-dir` filter to work...

```
└─▪ ramble workspace list --parent-dir $ramble/var/ramble/workspaces/
==> Error: Directory '/var/ramble/workspaces/' is not in configured workspace_dirs
```

It seems like it doesn't expand the paths when comparing to ensure they match even if they are relative in either the config or the CLI, and then absolute in the other.

Also, when I use `ramble workspace list` with multiple workspace dirs, I don't see information about which workspace is in which workspace dir. It might be nice to have something like:

```
Workspaces from dir: <workspace_dir1>
workspace1    workspace2   ....

Workspaces from dir: <workspace_dir2>
otherworkspace1    otherworkspace2
```

Then I would want to probably be able to copy / paste the workspace dir from the top to a `--parent-dir` argument to filter the workspaces.

rfbgo:
Thanks for the comments, I will look into it. I think the first one is simply you are letting your env expand an undefined variable. This works for me:

```
 $ ramble workspace list --parent-dir \$ramble/var/ramble/workspaces
```

I can add the listing by workspace, right now it's a combined view under the assumption that you can just activate it and get the path that way

rfbgo:
How do you feel about:

```
ramble workspace list --by-parent
~/fakerepo:
    hostname

~/ramble/var/ramble/workspaces:
    clean_hostname  hostname    
```

Or do you prefer it's the default behavior?

douglasjacobsen:
I think I'd prefer it's the default behavior. It feels like it's useful even if you have a single workspace directory, just to know where the workpsaces are located.

rfbgo:
> I think I'd prefer it's the default behavior. It feels like it's useful even if you have a single workspace directory, just to know where the workpsaces are located.

Sounds good, done!

douglasjacobsen (approved):

PR #1526, opened 2026-04-24
Add support for scoped variants [AI]

dapomeroy:
Add support for application-scoped variants, which can be used to parameterize aspects of the application definition

### DIFF
--- a/lib/ramble/docs/configuration_files.rst
+++ b/lib/ramble/docs/configuration_files.rst
@@ -850,6 +850,7 @@ The variants config section is used to customize variants to the experiment crea
 These can include application defined variants, or higher level Ramble provided
 variants. Currently, the only supported variants is ``package_manager`` which
 allows an experiment to define the package manager it should use.
+
 The format of this section is as follows:
 
 .. code-block:: yaml
--- a/lib/ramble/docs/dev_guides/application_dev_guide.rst
+++ b/lib/ramble/docs/dev_guides/application_dev_guide.rst
@@ -420,6 +420,57 @@ In this case, the variable ``gpu_flag`` will be defined, and will have
 a value of ``-g`` or an empty string, depending on the value of the
 ``gpu`` variant.
 
+Variants as Expansion Variables
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Additionally, variants can be used as an expansion variable with the syntax
+``{{object_type}::variant::{variant_name}}``. For example, a variant named ``openmp``
+defined within an application can be expanded from 
+``{application::variant::openmp}``.
+
+Boolean variants will evaluate as follows:
+
+* ``+<variant_name>`` if the variant value is True.
+* ``~<variant_name>`` if the variant value is False.
+
+For example, if the ``openmp`` variant is set to ``True``, it will evaluate to
+``+openmp``. If it is set to ``False``, it will evaluate to ``~openmp``.
+
+Multi-value variants evaluate to a ``key=<value>...`` format. For example, to add
+support for GROMACS's `build_type` variant in Ramble, it could be defined as:
+
+.. code-block:: python
+
+    variant(
+        "build_type",
+        default="Release",
+        description="The build type to build",
+        values=[
+            "Debug",
+            "Release",
+            "RelWithDebInfo",
+            "MinSizeRel",
+            "Reference",
+            "RelWithAssert",
+            "Profile",
+        ],
+    )
+
+These variants can then be used to parameterize a software spec:
+
+.. code-block:: python
+
+    with when("package_manager_family=spack"):
+        software_spec(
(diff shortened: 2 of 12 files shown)


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -653,9 +653,22 @@ def define_missing_variables(self):
...
+            self.define_variable(
+                f"{self.origin_type}::variant::{name}", variant.as_definition()
+            )
+
+        for name, variant in self.object_variants.default_variants.items():
+            if name not in self.object_variants.experiment_variants:
+                self.define_variable(
+                    f"{self.origin_type}::variant::{name}",
+                    variant.as_definition(),
+                )
```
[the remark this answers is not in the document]
douglasjacobsen: I agree with this. I think you could delete this block, and then below remove the `exclude_types` argument to `self.objects` and it should function the same.

lib/ramble/docs/dev_guides/application_dev_guide.rst
```diff
@@ -420,10 +420,61 @@ In this case, the variable ``gpu_flag`` will be defined, and will have
...
+Which might evaluate to:
+
+.. code-block:: console
+
+    gromacs@2026.1 +openmp build_type=Release
 
 ^^^^^^^^^^^^^^^^^^^^
 License Names
-^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^
```
douglasjacobsen: This change seems like the style stuff might get mad, but also like it's unrelated?

lib/ramble/docs/configuration_files.rst
```diff
@@ -846,19 +846,21 @@ For more information on variable expansion rules, see:
 Variants Section
 ----------------
 
-The variants config section is used to customize variants to the experiment creation.
-These can include application defined variants, or higher level Ramble provided
-variants. Currently, the only supported variants is ``package_manager`` which
-allows an experiment to define the package manager it should use.
+The variants config section is used to customize variants for experiment creation.
+These can include application-defined variants, or higher-level Ramble provided
+variants (such as ``package_manager``, which controls the package manager an
```
douglasjacobsen: Similar changes are also in the systems and platforms PR. We need to figure out the best way to merge these.

lib/ramble/docs/workspace_config.rst
```diff
@@ -470,25 +470,50 @@ Versions can also be parameterized as a variable:
...
+-----------------
+Variant Expansion
+-----------------
+
+Variants can be expanded like variables into a Spack-like syntax by using the syntax ``{{object_type}::variant::{variant_name}``. For example, a boolean variant with a value of ``True`` formats to ``+bool``, whereas ``False`` formats to ``~bool``. A value-based variant formats to ``key=value``.
+
+Variant Expansion Example
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Suppose multiple applications in a workspace use the variant ``openmp`` (boolean) to parameterize their software specs. We can define it under the workspace ``variants:`` section:
```
douglasjacobsen: Make sure to call out that this is specific to spack.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -671,17 +658,36 @@ def define_missing_variables(self):
                 str(obj.selected_version),
             )
 
-            for when_key, var_list in obj.object_variables.items():
-                keep = False
-                for var in var_list:
-                    if var.name not in self.variables:
-                        keep = True
+            # Define variant variables for Spack-like syntax expansion
+            for (
```
douglasjacobsen, line 662: We might want to make sure the variables are definded the same way the variant (version) variables are defined for other objects.

Mostly, modifiers concern me. I feel like we either don't define modifiers here, or we only define modifier variables inside the modifier.

douglasjacobsen (changes requested):

lib/ramble/docs/configuration_files.rst
dapomeroy: Just looked at your change in the systems and platforms PR, and it looks good to me. I'll just go ahead and remove the chage from this one to avoid conflict.

lib/ramble/docs/workspace_config.rst
dapomeroy: Updated for clarity

var/ramble/repos/builtin/base_classes/application-base/base_class.py
dapomeroy: Version variant variables are defined in the same loop using the same method. Unless there's more to it than the 'define_variable()' call, I think they're defined the same way.

douglasjacobsen: Ah yeah, OK. The logic I was looking for happens in the experiment_variants helper.

douglasjacobsen (approved):

PR #1535, opened 2026-04-28
add version 2025 to lammps to strongly imply support 

rfbgo:
I manually verified the foms. It seems like anything >=2022 uses the same consistent format

<=2021 likely needs different foms

### DIFF
--- a/var/ramble/repos/builtin/applications/lammps/application.py
+++ b/var/ramble/repos/builtin/applications/lammps/application.py
@@ -76,6 +76,7 @@ class Lammps(ExecutableApplication):
         workload_group="all_workloads",
     )
 
+    version("20250722", "Version 20250722 of LAMMPS")
     version("20220623.4", "Version 20220623.4 of LAMMPS", preferred=True)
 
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/lammps/application.py
```diff
@@ -76,6 +76,7 @@ class Lammps(ExecutableApplication):
         workload_group="all_workloads",
     )
 
+    version("20250722", "Version 20250722 of LAMMPS")
```
linsword13, line 79: Do we want this to be the preferred version?

rfbgo: My opinion is "not yet", since we've not done much validation on it.  WDYT?

(My effort here was around a fear the FOMs change, but it turned out that was not true)

linsword13: Sure, makes sense.

linsword13 (approved):

PR #1540, opened 2026-05-01
Add perf modifier

rfbgo:
This likely needs additional foms added for platforms, but is a good starting place for users

### DIFF
--- a/var/ramble/repos/builtin/modifiers/perf/modifier.py
+++ b/var/ramble/repos/builtin/modifiers/perf/modifier.py
@@ -62,9 +62,17 @@ class Perf(BasicModifier):
         modes=["stat", "record"],
     )
 
+    modifier_variable(
+        "use_sudo",
+        default="True",
+        description="Use 'sudo' when executing the 'perf' command",
+        modes=["stat", "record"],
+    )
+
     executable_modifier("apply_perf")
 
     def apply_perf(self, exe_name, exe, app_inst=None):
+        assert False, "apply_perf was called"
         pre_cmds = []
         post_cmds = []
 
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/modifiers/perf/modifier.py
```diff
@@ -0,0 +1,199 @@
...
+            + hostname_cmd
+            + ".out"
+        )
+
+        pre_cmds.append(
+            CommandExecutable(
+                f"start-perf-{exe_name}",
+                template=[
+                    f'perf_log="{log_path}"',
+                    f'sudo {perf_cmd} > "$perf_log" 2>&1 &',
```
douglasjacobsen: Can we make the `sudo` portion configurable? It seems likely that most people wouldn't have sudo permissions, and this would just break for them.

var/ramble/repos/builtin/modifiers/perf/modifier.py
```diff
@@ -0,0 +1,199 @@
...
+                redirect="",
+                output_capture="",
+            )
+        )
+
+        post_cmds.append(
+            CommandExecutable(
+                f"stop-perf-{exe_name}",
+                template=[
+                    'sudo kill -INT "$perf_pid"',
```
douglasjacobsen: Same sudo comment

douglasjacobsen (changes requested):

var/ramble/repos/builtin/modifiers/perf/modifier.py
rfbgo: Great call, fixed

var/ramble/repos/builtin/modifiers/perf/modifier.py
rfbgo: Great call, fixed

var/ramble/repos/builtin/modifiers/perf/modifier.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+
+    modifier_variable(
+        "perf_subcommand",
+        default="stat",
+        description="Internal variable for the perf subcommand",
+        modes=["stat", "record"],
+    )
+
+    modifier_variable(
```
douglasjacobsen: This feels like it's better suited as a variant than a variable.

Maybe if we make `use_sudo` a variant, then `sudo_prefix` can be a variable that is either empty (when ~use_sudo) or `sudo ` (when +use_sudo). But that also lets users update the sudo prefix to be whatever they want.

Alternatively, it being a variant and having the logic to apply the prefix exist in the method like it does in this PR feels like it simplifies the logic you have (since we already handle the "if yes, on, true" stuff).

douglasjacobsen (changes requested):

var/ramble/repos/builtin/modifiers/perf/modifier.py
rfbgo: yea you're totally right, that's way better. I'll do it now

douglasjacobsen (approved):

PR #1541, opened 2026-05-01
Add system and platform object support

douglasjacobsen:
This merge adds support for variants to control the underlying system and platform selection. These are intended to help standardize the selection of a cluster and a compute node within a cluster across Ramble experiments.

Language features are added to help manage auxiliary software files, as well as variables whose definition come from the execution of some command.

Tests and documentation are also added.


lib/ramble/ramble/definitions/variables.py
```diff
@@ -87,6 +95,87 @@ def copy(self):
...
+                        shlex.split(cur_command), stdin=command_ps[-1], stdout=subprocess.PIPE
+                    )
+                else:
+                    new_p = subprocess.Popen(shlex.split(cur_command), stdout=subprocess.PIPE)
+                command_ps.append(new_p)
+
+            for idx in range(len(command_ps) - 1):
+                command_ps[idx].close()
+
+            result = command_ps[-1].communicate()[0].decode()
```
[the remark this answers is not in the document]
douglasjacobsen: Thanks! Accepted.

lib/ramble/ramble/language/system_language.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+
+    def _execute_available_platforms(obj):
+        platform_list = getattr(obj, "system_available_platforms", [])
+
+        if isinstance(platforms, list):
+            platform_list.extend(platforms)
+        else:
+            platform_list.append(platforms)
+        obj.system_available_platforms = sorted(set(platforms))
```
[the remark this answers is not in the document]
douglasjacobsen: Accepted.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -437,6 +441,122 @@ def _set_package_manager(self):
...
+
+            for variant in ["platform", "package_manager", "workflow_manager"]:
+                if (
+                    self.experiment_variants(allow_caching=False).value(
+                        variant
+                    )
+                    is None
+                ):
+                    default_value = getattr(
+                        self.system, f"default_{variant}", None
```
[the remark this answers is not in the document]
douglasjacobsen: Accepted

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -437,6 +441,122 @@ def _set_package_manager(self):
...
+            plat_var = "user-managed"
+
+        if plat_var is not None:
+            plat_name, _, maybe_plat_ver = plat_var.partition("@")
+
+            if self.system and self.system.system_available_platforms:
+                if plat_name not in self.system.system_available_platforms:
+                    logger.die(
+                        f"Platform {plat_name} is not available in system {self.system.name}. "
+                        f"Available platforms are: {', '.join(self.system.available_platforms)}"
```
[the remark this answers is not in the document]
douglasjacobsen: Accepted

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
```diff
@@ -344,6 +345,126 @@ def define_missing_packages(self, workspace):
...
+
+                        if not os.path.isabs(dest_path):
+                            dest_path = os.path.join(env_path, dest_path)
+
+                        os.makedirs(os.path.dirname(dest_path), exist_ok=True)
+
+                        with tempfile.TemporaryDirectory(
+                            delete=False
+                        ) as tmpdir:
+                            dest_name = os.path.basename(dest_path)
```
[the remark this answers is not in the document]
douglasjacobsen: Accepted

lib/ramble/ramble/definitions/variables.py
```diff
@@ -87,6 +95,89 @@ def copy(self):
...
+                        stdout=subprocess.PIPE,
+                    )
+                else:
+                    new_p = subprocess.Popen(shlex.split(cur_command), stdout=subprocess.PIPE)
+                command_ps.append(new_p)
+
+            for idx in range(len(command_ps) - 1):
+                command_ps[idx].stdout.close()
+
+            result = command_ps[-1].communicate()[0].decode("utf-8")
```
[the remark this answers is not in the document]
douglasjacobsen: Accepted

var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
```diff
@@ -1185,65 +1230,85 @@ def _env_file_dict(self):
...
+                new_hash = ramble.util.hashing.hash_file(file)
+                old_hash = ramble.util.hashing.hash_file(env_file)
+                if new_hash != old_hash:
+                    migrate = True
+            else:
+                migrate = True
+
+            if migrate:
+                break
+
```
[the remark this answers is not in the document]
douglasjacobsen: This isn't true. We use `rmtree` to delete an existing env path before copying files into the env.

lib/ramble/docs/platforms.rst
```diff
@@ -0,0 +1,72 @@
...
+   https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+   <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+   option. This file may not be copied, modified, or distributed
+   except according to those terms.
+
+
+.. _platform-control:
+
+=========
+Platforms
```
linsword13, line 13: Annoying question: do you envision say `g4-standard-96` and `g4-standard-192` to be completely separate platforms, or would it have value to have a hierarchy to represent `g4`?

lib/ramble/ramble/language/platform_language.py
```diff
@@ -0,0 +1,59 @@
...
+class PlatformMeta(ramble.language.shared_language.SharedMeta):
+    _directive_names = set()
+    _directives_to_be_executed = []
+
+
+platform_directive = PlatformMeta.directive
+
+
+@platform_directive("class_families")
+def family(name, **kwargs):
```
linsword13: What's the difference between this and `platform_family`?

lib/ramble/docs/platforms.rst
```diff
@@ -0,0 +1,72 @@
...
+Platforms are controlled through a config option in the 
+:ref:`variants configuration section <variants-config>`. The following shows an
+example of controlling this
+
+.. code-block:: yaml
+
+  variants:
+    platform: <platform_name>
+
+The default platform is ``user-managed`` which simply requires the user to
```
linsword13, line 39: Should the default be something like `None`, to avoid causing validation errors for existing workspaces?

lib/ramble/docs/platforms.rst
```diff
@@ -0,0 +1,72 @@
...
+ * user-managed
+
+^^^^^^^^^^^^^^^^^^^^^
+User Managed Platform
+^^^^^^^^^^^^^^^^^^^^^
+
+The selection of ``user-managed`` for system will require users to specify the
+following variables:
+
+ * ``platform_accelerators`` - The number of accelerators on each node of this platform
```
linsword13: Naming is indeed tough here... My preference is to use `accelerators_per_node` for this, and maybe `accelerator_threads_per_node` for the runtime config. But admittedly the latter is still quite confusing.

Separately, do you think we should consider the scenario when there are different types of accelerators on  a given platform?

lib/ramble/docs/platforms.rst
douglasjacobsen: So, there are lots of weird things you could do here. And I think any of them are probably viable, but my gut reaction is that we would have a base platform for `g4` and then specific implemenation of it that change the `cores_per_node` variable.

That being said, I imagine that `h4d` would have a `memory_configuration` variant that could be set to `highmem` or `standard` and another variant for whether local ssd is used or not.

Beyond that, users (within their workspace) are free to edit the `cores_per_node` variable to change how it works. The values in the system / platform classes are really only intended to be defaults and a standardized set of names that users should be able to rely on.

lib/ramble/ramble/language/platform_language.py
douglasjacobsen: They do the same thing. And probably should be collapsed. But I think we have the `family` and `<obj>_family` directives to allow people to choose which syntax they prefer. The `<obj>_family` one is probably better though.

lib/ramble/docs/platforms.rst
douglasjacobsen: I think I've already handled existing workspaces with keeping `user-managed` the default. Previously when we allowed `None` for package managers, it ended up causing more issues.

lib/ramble/docs/platforms.rst
douglasjacobsen: What about `accelerators_on_node` (platform defined) and `accelerators_per_node` (user defined)?

I don't think I've ever seen a single platform with multiple accelerators in it. I think it would ultimately boil down to "do we want an experiment to run one two different accelerators on the same platform at the same time?" If that's not true, then I think what we have (with variants and stuff) is probably fine as it's easy to swap between the acceleartor type per experiment.

douglasjacobsen: These should be updated to `max_` prefixed names now.

lib/ramble/ramble/definitions/variables.py
```diff
@@ -87,6 +95,103 @@ def copy(self):
...
+        result = self.dry_run_value
+
+        if not workspace.dry_run:
+            split_command = shlex.split(expanded_command)
+            cur_command = ""
+            command_chains = []
+            command_ps = []
+            for cmd_part in split_command:
+                if cmd_part == "|":
+                    command_chains.append(cur_command)
```
rfbgo, line 161: If we "consume" `cur_command` here is it safe to not reset it? Right now it feels like we may add stubs of the same commands multiple times?

lib/ramble/ramble/definitions/variables.py
```diff
@@ -87,6 +95,103 @@ def copy(self):
...
+                        )
+                    else:
+                        new_p = subprocess.Popen(shlex.split(cur_command), stdout=subprocess.PIPE)
+                    command_ps.append(new_p)
+
+                for idx in range(len(command_ps) - 1):
+                    command_ps[idx].stdout.close()
+
+                result = command_ps[-1].communicate()[0].decode("utf-8").strip()
+            except FileNotFoundError:
```
rfbgo, line 186: In what conditions is this expected to happen? My naive read feels like this might be useful to emit a debug, but the way you wrote it makes me think there is likely some case where it is expected?

var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
```diff
@@ -1182,65 +1221,87 @@ def _env_file_dict(self):
...
+        Returns:
+            (bool): Whether the env was migrated or not
+        """
+        migrate = False
 
-        # Write spack.yaml to environment before concretizing, and its hash
-        with open(os.path.join(self.env_path, "spack.yaml"), "w+") as f:
-            syaml.dump_config(env_file, f, default_flow_style=False)
+        if not ws_env_path or not stage_env_path:
+            return
```
rfbgo: nit: Is  this actually `return False`?

```suggestion
            return False
```

lib/ramble/ramble/cmd/style.py
```diff
@@ -109,6 +109,18 @@ def is_object(f):
         "F403": [r"^from ramble.wmkit import \*$"],
         **common_object_exemptions,
     },
+    rf"platform.py|{base_class_file}$": {
+        # Allow 'from ramble.modkit import *' in workflow_managers,
```
rfbgo: Is this right? Should maybe be platform?

lib/ramble/ramble/cmd/style.py
```diff
@@ -109,6 +109,18 @@ def is_object(f):
...
+        # but no other wildcards
+        "F403": [r"^from ramble.platkit import \*$"],
+        **common_object_exemptions,
+    },
+    rf"system.py|{base_class_file}$": {
+        # Allow 'from ramble.modkit import *' in workflow_managers,
```
rfbgo: Is this right? Should maybe be system?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1274,6 +1283,7 @@ def workspace_manage_experiments_setup_parser(subparser):
         help="name of generated experiment",
     )
 
+    # TODO: Deprecate / remove in favor of the -V option
```
rfbgo: Can we add some deprecation warning for this via the unit tests so we remember to clean it next release?

lib/ramble/ramble/definitions/variables.py
douglasjacobsen: Good call. Updated

lib/ramble/ramble/definitions/variables.py
douglasjacobsen: Yeah, this is so you can run the command variables on systems where the commands don't exist. For example, if you are running something like:

`sinfo -p {slurm_platform}` on your laptop where you definitely don't have slurm, this allows the setup to continue without just hard-erroring.

var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
douglasjacobsen: Yeah, you're right. Thanks!

lib/ramble/ramble/cmd/style.py
douglasjacobsen: Yeah, fixed. Thanks!

lib/ramble/ramble/cmd/style.py
douglasjacobsen: Yeah, fixed. Thanks!

lib/ramble/docs/dev_guides/system_and_platform_definition_tutorial.rst
```diff
@@ -0,0 +1,572 @@
...
+on your platform, and they are both GPUs.
+
+You can use the ``variable`` directive to define variables within the platform
+definition for each of these quantities. For example, defining the amount of
+memory per node can be done using:
+
+.. code-block:: python
+
+   variable(
+     "memory_per_node",
```
linsword13: Under the new convention, is this `max_memory_per_node`?

lib/ramble/ramble/language/shared_language.py
```diff
@@ -1147,6 +1147,77 @@ def _mark_required_var(obj):
...
+            "src_path": src_path,
+            "dest_path": dest_path,
+            "consumed": False,
+        }
+
+    return _execute_auxiliary_software_file
+
+
+@shared_directive("command_variables")
+def command_variable(
```
linsword13, line 1178: Thinking aloud, I wonder since the distinction seems to be on "how" the value is obtained, would it make sense to instead have an attribute to `object_variable` (something like `evaluate=True`), as opposed to having a new directive?

douglasjacobsen: It can definitely be something defined by the standard variable directive. I made it separate beacuse the logic is slightly different (in that we always try to evaluate the required command, and sometimes we use the dry_run_value instead).

That felt different enough that it should be a separate directive, but I'm OK either way.

linsword13: Does something like `default="$(...)"` work for this purpose? I guess it has the distinction of getting evaluated at setup time, or execution time though...

douglasjacobsen: No, using `$(...)` wouldn't work here. This is something that is evaluated at setup-time, and needs to be extracted by python (rather than letting bash evaluate it at experiment time).

lib/ramble/docs/dev_guides/system_and_platform_definition_tutorial.rst
douglasjacobsen: Good catch. Fixed!

lib/ramble/ramble/language/shared_language.py
linsword13: Got it, makes sense.

linsword13:
> ## Ramble Performance Test Metrics
> Results produced with commit: [8539528](https://github.com/GoogleCloudPlatform/ramble/commit/8539528987ec27b0fa2b7d05e9f8b39d80796708)
> 
> Test Name	Outcome	Duration (s)	Last 5 Avg (s)
> test_analyze_large_file	passed	3.1263	3.1238
> test_large_template_expansion	passed	2.0550	2.0985
> test_many_experiments	passed	87.3625	32.9883
> test_many_objects_defaults	passed	36.4297	30.3430
> test_matrix_filter_perf	passed	1.7066	1.6178

Looks like `test_many_experiments` is being hit quite hard. Do you know what's causing that?

douglasjacobsen:
Not really, but the logic for applying objects is more complex. We have a follow up task to see if we can improve the performance though.

linsword13 (approved):

PR #1543, opened 2026-05-04
Add `ruff check` into `ramble style`

linsword13:
Even though `ruff` is not at v1 yet, it has proven to be quite effective in catching genuine issues that are missed by `flake8`. It's also quite performant and does not add much to the total runtime of style-check.

### DIFF
--- a/requirements-pinned.txt
+++ b/requirements-pinned.txt
@@ -6,3 +6,4 @@ mypy==1.20.1;python_version>="3.9"
 coverage==7.13.5;python_version>="3.9"
 codecov-cli==11.2.5;python_version>="3.9"
 line_profiler==5.0.2;python_version>="3.9"
+ruff==0.15.12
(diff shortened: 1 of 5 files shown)


requirements-pinned.txt
```diff
@@ -6,3 +6,4 @@ mypy==1.20.1;python_version>="3.9"
 coverage==7.13.5;python_version>="3.9"
 codecov-cli==11.2.5;python_version>="3.9"
 line_profiler==5.0.2;python_version>="3.9"
+ruff==0.15.12
```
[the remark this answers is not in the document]
linsword13: This is fine, since ruff supports these older Python versions

douglasjacobsen, line 9: We should probably add an un-pinned `ruff` to the developer requirements too.

douglasjacobsen (changes requested):

requirements-pinned.txt
linsword13: Good call, added.

douglasjacobsen (approved):

PR #1552, opened 2026-05-08
Add conflict() directive [AI]

dapomeroy:
Adds a `conflict()` directive for objects to indicate a condition with a known conflict

### DIFF
--- a/lib/ramble/ramble/language/shared_language.py
+++ b/lib/ramble/ramble/language/shared_language.py
@@ -798,6 +798,37 @@ def register_validator(
     return _define_validator
 
 
+@shared_directive("conflicts", init_value=[])
+def conflict(
+    conflict_spec: str,
+    when: Optional[Union[str, List[str]]] = None,
+    msg: Optional[str] = None,
+    **kwargs,
+):
+    """Defines a conflict for this object.
+
+    Args:
+        conflict_spec: The trigger condition (e.g., `+variant`, `compiler=gcc`)
+        when: Optional conditional under which the conflict occurs
+        msg: Optional custom error/warning message
+    """
+
+    def _execute_conflicts(obj):
+        when_list = ramble.language.language_helpers.build_when_list(
+            when, obj, conflict_spec, "conflicts"
+        )
+
+        obj.conflicts.append(
+            {
+                "conflict_spec": conflict_spec,
+                "when": when_list,
+                "message": msg,
+            }
+        )
+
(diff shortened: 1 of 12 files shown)


lib/ramble/ramble/language/shared_language.py
```diff
@@ -798,6 +798,37 @@ def _define_validator(obj):
...
+        when: Optional conditional under which the conflict occurs
+        msg: Optional custom error/warning message
+    """
+
+    def _execute_conflicts(obj):
+        when_list = ramble.language.language_helpers.build_when_list(
+            when, obj, conflict_spec, "conflicts"
+        )
+
+        obj.conflicts.append(
```
douglasjacobsen: Is there a reason why this doesn't follow the same pattern of our other directives that allow when? Specifically having something like `when_key` and then a list of conflicts inside that?

douglasjacobsen (commented):
Thanks! One question, otherwise this has merge conflicts that need to be resolved.

One small nit too, it might be good for the directive to be `conflict` rather than `conflicts`. It also should generally not have the same name as the internal dictionary, or they will conflict (since we create methods that match the directive names, and the attribute won't work right).

lib/ramble/ramble/language/shared_language.py
dapomeroy: We have a number of other directives that have a list of objects without keying a dict on `when` (e.g., software_spec). It didn't seem strictly necessary, but it wouldn't hurt to add it. Do you have a preference? 

Right now it loops over the whole list of conflicts and evaluates the conflict spec first and when second. If I add the when_key, it'll loop over that first and evaluate when, and then loop over the conflict spec. I'm having a hard time imagining what scenarios one will be more efficient than the other.

dapomeroy:
Resolved merge conflicts, updated directive name to 'conflict()'

lib/ramble/ramble/language/shared_language.py
douglasjacobsen: I feel like it's likely more performnant to have the when key on the outside. I also think this is more of the general pattern, and the ones that don't have the when key on the outside are an exception where it wouldn't quite work the same way. But we can always make some tasks to clean those up eventually.

dapomeroy: I've updated it to use the when key

douglasjacobsen (approved):

PR #1559, opened 2026-05-13
Add directives for editing or patching files [AI]

douglasjacobsen:
This merge adds two new directives, one for performing portable editing of a file, and another for patching files. This allows ramble objects to define manipulations on files in the experiments using a portable syntax that can be applied across operating systems.

Additionally, the edit directive allows users to define a python function to edit the resulting file which will then be executed as part of the experiment.

### DIFF
--- a/lib/ramble/ramble/language/shared_language.py
+++ b/lib/ramble/ramble/language/shared_language.py
@@ -40,6 +40,7 @@ inherit from the SharedMeta class.
 class SharedMeta(ramble.language.language_base.DirectiveMeta):
     _directive_names = set()
     _directives_to_be_executed = []
+    _directive_init_values = {"custom_edit_functions": {}}
 
 
 shared_directive = SharedMeta.directive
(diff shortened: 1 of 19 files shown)


lib/ramble/ramble/language/shared_language.py
```diff
@@ -40,11 +40,180 @@ class Gromacs(ExecutableApplication):
...
+        import inspect
+        import shlex
+
+        from ramble.util.executable import CommandExecutable
+        from ramble.util.file_editor import HELPER_SCRIPT_NAME
+
+        template = [
+            f'python "{{workspace_shared}}/utilities/'
+            f'{{application_name}}.{{workload_name}}.{HELPER_SCRIPT_NAME}" '
+            f"--mode regex --file {shlex.quote(file_path)}"
```
rfbgo: the util naming scheme seems subtly different between here (no dirs) and elsewhere (with dirs):

```
        template = [
            f'python "{{workspace_shared}}/utilities/{{application_name}}/'
            f'{{workload_name}}/{HELPER_SCRIPT_NAME}" '
            f"--mode patch --file {shlex.quote(file_path)} --patch-file {shlex.quote(patch_file)}"
        ]
```

Is there a reason for this? My general feeling is that this path generation should be in a function to help make this clean and repeatable instead of inline

lib/ramble/ramble/workspace/workspace.py
```diff
@@ -805,9 +818,58 @@ def write_templates(self):
...
+
+                    script_content = ramble.util.file_editor.get_file_editor_script(
+                        custom_functions=custom_functions
+                    )
+
+                    script_name = (
+                        f"{app_name}.{workload_name}.{ramble.util.file_editor.HELPER_SCRIPT_NAME}"
+                    )
+                    script_path = os.path.join(self.shared_utilities_dir, script_name)
+                    with open(script_path, "w+", encoding="utf-8") as f:
```
rfbgo: Does this write still happen for apps that don't use a custom function? My naive read is that this script is highly likely to be the same for different apps/workloads that don't do a custom hook. If so it might be nice a (perhaps future) enhancement to point to a shared/core file instead of scattering scripts around

lib/ramble/ramble/language/shared_language.py
```diff
@@ -40,11 +40,180 @@ class Gromacs(ExecutableApplication):
...
+        ]
+        if match:
+            template[0] += f" --match {shlex.quote(match)}"
+        if replace is not None:
```
rfbgo, line 155: Maybe I missed it, but can this parsing lead to a no-op/undervalidated requests? 

It seems like we can go get through this code path without actually specifying an action / replace and I don't see where we enforce that a selection was made

Similarly do we want to add guarding against things like both match and replace being specified?

douglasjacobsen: I believe this is resolved now. There is a check that you are perfoming at least one action with the edit directive when it is defined.

lib/ramble/ramble/workspace/workspace.py
douglasjacobsen: This should be fixed now. I changed it so the custom functions are written to the experiment directories and then there is a single utility written to the shared directory (which can consume these custom functions). I can't decide how I feel about the name of the directory it creates though.

The custom functions are written to `{experiment_run_dir}/utilities` and the shared utility is written to `$workspace_root/shared/utilities`. Let me know if you have opinions about these...

lib/ramble/ramble/language/shared_language.py
douglasjacobsen: This should be resolved now.

rfbgo (approved):

PR #1565, opened 2026-05-16
Fix variable precedence application

douglasjacobsen:
This merge fixes an issue where object precedence is incorrectly applied with system and platform classes.

### DIFF
--- a/lib/ramble/ramble/test/end_to_end/object_precedence.py
+++ b/lib/ramble/ramble/test/end_to_end/object_precedence.py
@@ -12,14 +12,11 @@ import pytest
 
 import ramble.workspace
 from ramble.main import RambleCommand
+from ramble.util.command_runner import RunnerError
 
 pytestmark = pytest.mark.usefixtures(
     "mutable_config",
     "mutable_mock_workspace_path",
-    "mutable_mock_apps_repo",
-    "mutable_mock_mods_repo",
-    "mutable_mock_pkg_mans_repo",
-    "mutable_mock_wms_repo",
 )
 
 workspace = RambleCommand("workspace")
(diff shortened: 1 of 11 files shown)


lib/ramble/ramble/test/end_to_end/object_precedence.py
```diff
@@ -80,3 +86,47 @@ def test_object_precedence_variables(workspace_name, test_args, disabled_value,
...
+        assert os.path.isfile(exec_file)
+        with open(exec_file) as f:
+            data = f.read()
+            # Verify mpi_command from workflow manager is not set
+            assert "srun" not in data
+            # Verify mpi_command from system is set, since that
+            # is higher precedence than workflow managers.
+            assert "mpirun" in data
+            assert "-t sys-variant1-foo" in data
+    except RunnerError as e:
```
linsword13: Given that it's a dry-run, is this skipping needed?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -823,26 +823,43 @@ def define_missing_variables(self, workspace):
         """Iterate over missing variable definitions, and add them until there
         are no more to add."""
 
-        default_variables = {}
+        # Track which precedence level defined each variable.
```
linsword13, line 826: These precedence settings are clever. I do wonder if it'd be more readable to define an explicit enum for the different precedence levels?

Another nitpick: the name `app_precedence` is more like `default_value_precedence`?

lib/ramble/ramble/test/end_to_end/object_precedence.py
douglasjacobsen: Nope! Good catch, fixed.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
douglasjacobsen: I think defining an explict enum in this case might make it a bit more brittle. Mostly because this automatically defines the precedence order based on the order objects are returned from the objects function.

If we define an enum, then we need to make sure that these two are syncronized. Additionally, there are annoying edge cases (like if two modifiers define the same variable, how do we handle the precedence ordering).

In this case, each object gets a unique precedence value (and modifiers are given precedence in order of YAML definition).

All of that is to say that I think it being dynamic is a bit more useful in general. However, I think we could push the precedence definition into the `def objects()` method, and have it return these values in case we want to use them in other places too.

Good call on the var name though. I changed that.

linsword13: Sounds good, I was mistaken. Yes the dynamic nature makes infeasible to use an enum.

linsword13 (approved):

PR #1569, opened 2026-05-19
Optimize variant caching and workspace info generation [AI]

douglasjacobsen:
This merge addresses performance regressions introduced by the addition of system, platform, and package manager objects to every experiment instance.

- Lazy Phase and Software Metadata Aggregation: Guarded pipeline phase aggregation and software environment rendering in `workspace info` with conditional checks so they are only computed when explicitly requested.

- Variant Set Caching: Re-enabled default variant set caching in `experiment_variants()` during experiment setup and added a robust cache invalidation mechanism (`clear_variant_cache()`) that triggers whenever the object hierarchy changes (e.g., when setting system, platform, modifier objects).

- Variant Copy Optimization: Optimized `VariantSet.copy()` to avoid redundant deep copies of `Variant` objects, which are treated as immutable in this context.

These optimizations yield a ~28% reduction in the mean execution time of the `test_many_experiments` benchmark.

### DIFF
--- a/var/ramble/repos/builtin/base_classes/application-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/application-base/base_class.py
@@ -386,9 +386,7 @@ class ApplicationBase(ObjectMixin, metaclass=ApplicationMeta):
         all_workloads_names = set()
         found = False
         for when_set, workloads in self.workloads.items():
-            if self.expander.satisfies(
-                when_set, self.experiment_variants(allow_caching=False)
-            ):
+            if self.expander.satisfies(when_set, self.experiment_variants()):
                 for workload_name, workload in workloads.items():
                     if workload_name in all_workloads_names:
                         logger.die(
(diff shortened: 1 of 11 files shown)


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -529,11 +519,13 @@ def _set_platform(self):
                     plat_name, plat_type
                 ).copy()
                 self.platform.set_application(self)
+                self.clear_variant_cache()
                 if maybe_plat_ver:
                     self.platform.set_version(
                         version_number=maybe_plat_ver,
                         description=f"{plat_name} {maybe_plat_ver}",
                     )
+                    self.clear_variant_cache()
```
linsword13: Mostly wondering if these clear_cache calls should be embedded in the respective methods (`platform.set_application`, `platform.set_version`, etc.)?

rfbgo: Great timing Lin, I was thinking something similar. I happen to also be reviewing this PR right now, and I was trying to figure out if things like `build_modifier_instances` should have the cache cleared inside the function / if we are safely doing it "around" the call today

douglasjacobsen: This sould be improved now. I went through and cleaned up the clearing of the variant cache a lot more aggressively than was done previously.

rfbgo (approved):
Thanks for the changes!

PR #1570, opened 2026-05-19
Clean up references to experiment_set and workspace

rfbgo:


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -35,7 +35,6 @@
 import ramble.success_criteria
 import ramble.util.class_attributes
 import ramble.util.colors as rucolor
-import ramble.util.directives
```
rfbgo: This was removed by ruff in `ramble style -a -f`, but is that correct? We might need to tweak the ruff config if not

linsword13: Yes it's correct. If you rebase this change should go away. (This happens because the style PR check only looks at changed files, so some ruff violations in un-changed files were not included.)

linsword13 (dismissed):

var/ramble/repos/builtin/base_classes/application-base/base_class.py
rfbgo: Awesome, thanks for confirming!

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -1493,7 +1510,7 @@ def build_modifier_instances(self, workspace):
                 mod_inst.expander.add_no_expand_var(var)
 
         # Define any missing modifier variables
-        self.define_missing_variables(workspace)
+        self.define_missing_variables()
```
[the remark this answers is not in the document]
rfbgo: @douglasjacobsen any opinion on this?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -817,9 +816,11 @@ def register_missing_command_variable(self, var):
         """
         self._missing_command_variables[var.name] = var
 
-    def define_missing_variables(self, workspace):
+    def define_missing_variables(self, workspace=None):
```
douglasjacobsen: Is there a way we can remove the workspace argument here? Or do we need it for something?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -961,11 +962,13 @@ def set_chained_experiments(self, chained_experiments):
         if chained_experiments:
             self.chained_experiments = chained_experiments.copy()
 
-    def set_modifiers(self, modifiers, workspace):
+    def set_modifiers(self, modifiers, workspace=None):
```
douglasjacobsen: Same argument comment here. Maybe these should error or something if the workspace reference isn't set?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -1007,11 +1010,15 @@ def experiment_log_file(self, logs_dir):
             os.path.join(logs_dir, self.expander.experiment_namespace) + ".out"
         )
 
-    def get_pipeline_phases(self, pipeline, workspace, phase_filters=None):
+    def get_pipeline_phases(
+        self, pipeline, workspace=None, phase_filters=None
```
douglasjacobsen: Same comment here about the argument.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -1058,7 +1065,7 @@ def print_vars(self, header="", vars_to_print=None, indent=""):
             expanded = self.expander.expand_var(expansion_var)
             color.cprint(f"{indent}  {var} = {val} ==> {expanded}")
 
-    def build_used_variables(self, workspace):
+    def build_used_variables(self, workspace=None):
```
douglasjacobsen: Same argument comment here

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -1170,8 +1179,10 @@ def print_chain_order(self, indent=""):
             color.cprint(f"{indent}- {exp}")
 
     # Phase execution helpers
-    def run_phase(self, pipeline, phase, workspace):
+    def run_phase(self, pipeline, phase, workspace=None):
```
douglasjacobsen: Same argument comment here.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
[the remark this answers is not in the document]
douglasjacobsen: I think the review is only saying this becuase the argument still exists. I wonder if it'll say it if you remove the workspace argument entirely though.

douglasjacobsen (changes requested):
I left some comments, but I think generally we should remove the workspace arguments from these methods instead of just making the defaults None.

I also disagree with what Gemini said and thing that the changes should be safe (which is also evidenced with our tests...)

var/ramble/repos/builtin/base_classes/application-base/base_class.py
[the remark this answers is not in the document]
rfbgo: Yea I was thinking the same, I'll try it out

douglasjacobsen (approved):

PR #1582, opened 2026-05-22
Warn for duplicate YAML blocks

rfbgo:
This can cause some fairly subtle odd behavior since the duplication can hide things you think you asked for

Gemini is confident this is a low overhead change and should help users avoid confusing issues

Produces a warning like:

```
==> Warning: Duplicate key "config" detected in ./my_workspace/configs/ramble.yaml at line 25. This will overwrite the previous value.
```

### DIFF
--- a/lib/ramble/ramble/config.py
+++ b/lib/ramble/ramble/config.py
@@ -126,11 +126,13 @@ all_schemas = copy.deepcopy(section_schemas)
 all_schemas.update(dict.fromkeys(ramble.schema.workspace.keys, ramble.schema.workspace.schema))
 
 
-class RambleLineLoader(syaml.OrderedLineLoader):
+class _RambleLineLoader(syaml.OrderedLineLoader):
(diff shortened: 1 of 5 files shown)


lib/ramble/ramble/config.py
```diff
@@ -125,6 +125,38 @@
...
+                )
+            seen_keys.add(key)
+
+        super(syaml.OrderedLineLoader, self).construct_mapping(node, maptyp, deep=deep)
+
+
+def load_config(*args, **kwargs):
+    """Load YAML using Ramble's custom Loader that warns on duplicate keys."""
+    kwargs["Loader"] = RambleLineLoader
+    return syaml.load(*args, **kwargs)
```
rfbgo, line 170: FWIW this can actually just be `yaml.load` if we want to skip using syaml  for some reason

douglasjacobsen (approved):

PR #1586, opened 2026-05-26
Implement Filter Groups [AI]

dapomeroy:
This feature allows users to define named groups of filters (where and exclude_where) globally (in ~/.ramble/filter_groups.yaml) or per-workspace (in workspace ramble.yaml). These groups can be applied to commands using -fg/--filter-group and -efg/--exclude-filter-group with support for logical expressions (e.g., 'G1 and not G2'). Filter groups can also be persistently activated during workspace activation.

TAG=agy
CONV=9ca1006f-e4ba-44fd-80f3-89f837545cc7

### DIFF
--- a/etc/ramble/defaults/filter_groups.yaml
+++ b/etc/ramble/defaults/filter_groups.yaml
@@ -10,18 +10,15 @@ filter_groups:
   success:
     where:
     - '"{experiment_status}" == "SUCCESS"'
-  notsuccess:
-    where:
-    - '"{experiment_status}" != "SUCCESS"'
   failed:
     where:
     - '"{experiment_status}" == "FAILED"'
-  notfailed:
-    where:
-    - '"{experiment_status}" != "FAILED"'
   repeat_base:
     where:
-    - "{repeat_index} == 0"
+    - "{is_repeat_parent} == True"
   repeats_only:
     where:
-    - "{repeat_index} != 0"
+    - "{is_repeat_child} == True"
+  first_repeat:
+    where:
+    - "{repeat_index} == 1"
--- a/etc/ramble/defaults/filter_groups.yaml
+++ b/etc/ramble/defaults/filter_groups.yaml
@@ -1,0 +1,27 @@
+# Copyright 2022-2026 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+filter_groups:
+  success:
+    where:
+    - '"{experiment_status}" == "SUCCESS"'
+  notsuccess:
+    where:
+    - '"{experiment_status}" != "SUCCESS"'
+  failed:
+    where:
+    - '"{experiment_status}" == "FAILED"'
+  notfailed:
+    where:
+    - '"{experiment_status}" != "FAILED"'
+  repeat_base:
+    where:
+    - "{repeat_index} == 0"
+  repeats_only:
+    where:
+    - "{repeat_index} != 0"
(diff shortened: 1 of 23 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1830,6 +1847,138 @@ def workspace_manage_modifiers(args):
...
+    )
+
+    remove_parser = actions.add_parser("remove", aliases=["rm"], help="remove a filter group")
+    remove_parser.add_argument("-n", "--name", required=True, help="name of filter group")
+
+    actions.add_parser("list", help="list defined filter groups")
+    actions.add_parser("blame", help="show defined filter groups with sources")
+
+
+def workspace_manage_filter_groups(args):
```
douglasjacobsen, line 1890: After using this, it feels like the `workspace manage` should have a `--scope` option (similar to how manage modifiers works). But it should allow scopes outside of the workspace to be printed as well.

Mostly, it would be nice `ramble workspace manage filter-groups list` would print *all* of the filter groups (and their associated scope).

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen, line 760: Workspace info feels like it should also support filter-groups, given it has where and exclude where.

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -137,6 +138,13 @@ def workspace_activate_setup_parser(subparser):
         metavar="dir",
         help="parent directory for the named workspace to activate",
     )
+    subparser.add_argument(
+        "-fg",
```
douglasjacobsen: I think the shorter arguments for both filter-group and exclude-filter-group should have a double dash. I think single dash is, in general within Ramble, reserved for instances where there is only a single character after it.

douglasjacobsen (changes requested):
A few changes before I can test it more throughly.

Also, I ran into the following issue, which feels like it should work..

I ran `ramble workspace setup --dry-run` and then immediately tried to run the following (and got this output)

```
└─▪ ramble -D . workspace setup --dry-run -fg "failed and not success"
==> Streaming details to log:
==>   .../logs/setup.2026-06-02_13.37.41.out
      -- DRY-RUN -- DRY-RUN -- DRY-RUN -- DRY-RUN -- DRY-RUN --
==> Error: When evaluating ( ( (("{experiment_status}" == "FAILED")) ) and not ( (("{experiment_status}" == "SUCCESS")) ) ), evaluate_predicate returned a non-boolean string: "( ( (("SETUP" == "FAILED")) ) and not ( (("SETUP" == "SUCCESS")) ) )"
```

lib/ramble/ramble/cmd/workspace.py
dapomeroy: Added

lib/ramble/ramble/cmd/workspace.py
dapomeroy: Fixed

lib/ramble/ramble/cmd/workspace.py
dapomeroy: Fixed this to print all scopes by default and allow scoping in `workspace manage`. Addressed an issue where `workspace` was not displaying as a valid scope in `--help` when a workspace is active.

etc/ramble/defaults/filter_groups.yaml
```diff
@@ -0,0 +1,27 @@
...
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+filter_groups:
+  success:
+    where:
+    - '"{experiment_status}" == "SUCCESS"'
+  notsuccess:
```
douglasjacobsen: I don't think we need the not versions here, since we can not the base versions.

i.e.

"not success" and "notsuccess" are the same thing, but the latter is not a standard definition for other filter groups.

douglasjacobsen (changes requested):
Thanks! This is great so far. One small nit, but otherwise it's much improved. :)

etc/ramble/defaults/filter_groups.yaml
dapomeroy: Done

douglasjacobsen:
@dapomeroy there is a merge conflict now. But I also just wanted to make a note of our conversation from yesterday regarding the rebease_base_only and repeat_only filter groups. It'd be good to make sure that the repeat_base_only one does *not* include non-repeat experiments.

douglasjacobsen (approved):

PR #1588, opened 2026-05-28
add `simplify-names` flags for reports [AI]

rfbgo:
This helps in instances where x axis labels can be very long 

Also fixes a bug that was stopping pandas-where getting propagated in some cases


dapomeroy (approved):
This looks great, thanks for improving this!

PR #1589, opened 2026-05-28
Make slurm workflow manager call execute_experiment

rfbgo:
Placeholder for discussion of changing slurm workflow manager to use `execute_experiment`

```
...
Processing phase make_experiments (3/6):  50%|====================>                    | Elapsed (s): 0.00==> Warning: `sbatch` is missing in the given `batch_submit` command
==> Warning: In experiment hostname.local.test_default:
  `execute_experiment` contains `#SBATCH` directives, which will be ignored.
  Custom headers should be added to `extra_sbatch_headers` in the workspace configuration instead.
```

### DIFF
--- a/var/ramble/repos/builtin/base_classes/application-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/application-base/base_class.py
@@ -3006,15 +3006,18 @@ class ApplicationBase(ObjectMixin, metaclass=ApplicationMeta):
                 )
                 fs.mkdirp(os.path.dirname(expand_path))
 
+                rendered_content = self.expander.expand_var(
+                    template_conf["contents"], extra_vars=exec_vars
+                )
+
                 with open(expand_path, "w+", encoding="utf-8") as f:
-                    f.write(
-                        self.expander.expand_var(
-                            template_conf["contents"], extra_vars=exec_vars
-                        )
-                    )
+                    f.write(rendered_content)
                 os.chmod(expand_path, _DEFAULT_CONTENT_PERM)
 
             self._render_object_templates(exec_vars)
+
+            if self.workflow_manager:
+                self.workflow_manager.validate_experiment()
 
             experiment_script = workspace.experiments_script
             experiment_script.write(
--- a/var/ramble/repos/builtin/base_classes/application-base/base_class.py
+++ b/var/ramble/repos/builtin/base_classes/application-base/base_class.py
@@ -3016,9 +3016,6 @@ class ApplicationBase(ObjectMixin, metaclass=ApplicationMeta):
 
             self._render_object_templates(exec_vars)
 
-            if self.workflow_manager:
-                self.workflow_manager.validate_experiment()
-
             experiment_script = workspace.experiments_script
             experiment_script.write(
                 self.expander.expand_var("{batch_submit}\n")
--- a/var/ramble/repos/builtin/workflow_managers/slurm/slurm_experiment_sbatch.tpl
+++ b/var/ramble/repos/builtin/workflow_managers/slurm/slurm_experiment_sbatch.tpl
@@ -7,4 +7,4 @@ cd {experiment_run_dir}
 
 {workflow_hostfile_cmd}
 
-. {execute_experiment}
+. "{execute_experiment}"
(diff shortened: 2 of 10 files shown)


var/ramble/repos/builtin/workflow_managers/slurm/slurm_experiment_sbatch.tpl
```diff
@@ -7,4 +7,4 @@ cd {experiment_run_dir}
 
 {workflow_hostfile_cmd}
 
-{command}
+. "{execute_experiment}"
```
linsword13, line 10: This is a great idea!

One thing I wonder, is if (say Slurm) user has expectations that custom sbatch headers they added into execute_experiment template should work, and maybe surprised that they don't?

rfbgo: Yea great call out, this is definitely a very real and valid concern. Do you think it is sufficient to detect it and warn the user? or is it so concerning that we should not pursue this approach?

I guess it's also a risk they do this now into `execute_experiment` and not realize it is potentially unused

linsword13: I think this is the right approach, but a warning might be useful.

We also have an existing warning (via the `workflow_banner`), that's likely made obsolete with the current change.

rfbgo: I think I've now done this, ptal

linsword13: Thanks, LGTM. There are some unit test failures.

rfbgo: Thanks! Should be fixed now. There was actually a really fun unit test failure on an existing unrealted test because I added `validate_experiment` to the workflow maangers

https://github.com/GoogleCloudPlatform/ramble/pull/1589/changes#diff-40de26018453eabc2925665bb0de0937bc8a1d2f9ab2eef6fb737fae935efb1aL83

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -3006,16 +3006,19 @@ def _make_experiments(self, workspace, app_inst=None):
...
-                            template_conf["contents"], extra_vars=exec_vars
-                        )
-                    )
+                    f.write(rendered_content)
                 os.chmod(expand_path, _DEFAULT_CONTENT_PERM)
 
             self._render_object_templates(exec_vars)
 
+            if self.workflow_manager:
+                self.workflow_manager.validate_experiment()
```
linsword13: Actually instead of having this additional method call, would it work if adding a validator via the `register_validator` hook?

rfbgo: Hmm good question. I'm not sure. I did it this way to make the interface closer to what apps do: https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/base_classes/application-base/base_class.py#L1679 

What do you think the validator way would look like?

linsword13: Yeah so the object validator (added via the `register_validator`) is checked inside the app's `validate_experiment` method (in `_check_object_validators`). I think the validation function would like quite similar to your current code. The benefit I see it is that we have a more unified approach for adding such validations.

rfbgo: Oh I think see what you mean, you mean have the workflow managers call `_check_object_validators` and rely on a registered validator in the slurm wm rather than a direct validaiton impl? 

I can certainly look into that

linsword13: I think as long as you have a validator added via `register_validator`, then the validation will be called without explicitly invoking the `_check_object_validators`. This should apply to all objects.

rfbgo: This was a great call, thanks for pointing it out. Done and the code is nicer now

linsword13 (approved):

PR #1590, opened 2026-05-28
Add bootstrapped / required utilities objects [AI]

douglasjacobsen:
This merge adds utilities as a new object type. Utilities represent external dependencies that experiments (or other objects) can require. These can include things like spack (package managers) or pdsh. A language, documentation, and tests are added to help developers create new utilities.

### DIFF
--- a/lib/ramble/docs/dev_guides/utility_dev_guide.rst
+++ b/lib/ramble/docs/dev_guides/utility_dev_guide.rst
@@ -1,0 +1,162 @@
+.. Copyright 2022-2026 The Ramble Authors
+
+   Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+   https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+   <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+   option. This file may not be copied, modified, or distributed
+   except according to those terms.
+
+.. _utility-dev-guide:
+
+=======================
+Utility Developer Guide
+=======================
+
+This guide details how to create and manage utility definitions in Ramble.
+
+What is a Utility?
+==================
+
+Utilities in Ramble are external dependencies or tools that are required by experiments or applications to function correctly. This could include package managers (like Spack), system monitoring tools, or job execution frameworks.
+
+Utilities are defined as Python classes that inherit from `UtilityBase`.
+
+Creating a New Utility
+======================
+
+To create a new utility, you must define a class that inherits from `UtilityBase`. This definition must be placed in a file named `utility.py` within a directory named after your utility, inside a valid repository's `utilities/` folder.
+
+For example, a utility named `my_util` would be located at `var/ramble/repos/builtin/utilities/my_util/utility.py`.
+
+Basic Structure
+---------------
+
+.. code-block:: python
+
+    from ramble.utility import UtilityBase
+    from ramble.language.utility_language import *
+
+    class MyUtil(UtilityBase):
+        """MyUtil provides system information capabilities."""
+
+        name = "my_util"
+        maintainers("your_username")
+
+        bootstrappable(True)
+
+        variable(
+            name="path",
+            default="system",
+            description="Path to MyUtil",
+            scoped=True,
+        )
+
+        provides_executable(
+            "my_util_bin",
+            version_cmd="my_util_bin --version",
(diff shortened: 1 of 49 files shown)


lib/ramble/ramble/cmd/workspace.py
```diff
@@ -1042,6 +1077,65 @@ def workspace_info(args):
...
+            for experiments, _workload_context in ws.all_workloads(workloads):
+                for _exp_contents, _experiment_context in ws.all_experiments(experiments):
+                    app_ctx = _application_context.context_name
+                    wl_ctx = _workload_context.context_name
+                    exp_ctx = _experiment_context.context_name
+                    full_exp_name = f"{app_ctx}.{wl_ctx}.{exp_ctx}"
+                    app_inst = experiment_set.get_experiment(full_exp_name)
+
+                    objects_to_check = [app_inst]
+                    if hasattr(app_inst, "package_manager") and app_inst.package_manager:
```
linsword13: I feel that these `if`s can be replaced by iterating over the `_objects` iterator

var/ramble/repos/builtin/base_classes/utility-base/base_class.py
```diff
@@ -0,0 +1,297 @@
...
+        min_version=None,
+        max_version=None,
+        env=None,
+        origin_name=None,
+        origin_type=None,
+    ):
+        """Check if the provided executables are available and satisfy version constraints.
+        Uses the provided environment or the system environment if None.
+        """
+        import os
```
linsword13: This is already imported at the top-level

var/ramble/repos/builtin/base_classes/utility-base/base_class.py
```diff
@@ -0,0 +1,297 @@
...
+        import re
+        import shutil
+        import subprocess
+
+        from spack.version import Version
```
linsword13: In another PR, @dapomeroy mentioned that we want to avoid taking a dependency on Spack's version module. Looping him in as a FYI.

lib/ramble/docs/dev_guides/utility_dev_guide.rst
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+        env_prepend("PATH", "{utility::my_util::path}/bin")
+
+Utility Directives
+==================
+
+Utilities have access to a specific set of declarative directives to define their behavior.
+
+``bootstrappable``
```
linsword13, line 67: Random thought: I wonder if we can rely on the utility has certain method defined (such as `install`) to infer this. Classes that don't have the method defined is not bootstrappable, etc.

var/ramble/repos/builtin/base_classes/utility-base/base_class.py
douglasjacobsen: I don't understand why agy does this.... thanks! fixed.

lib/ramble/ramble/cmd/workspace.py
douglasjacobsen: Good catch, fixed!

lib/ramble/docs/dev_guides/utility_dev_guide.rst
douglasjacobsen: I think this would be cool. So if there is a way to do this that you see, let me know. But simply relying on the install method I don't think is good enough. The install method is only intended as an override, in case installation isn't a simple "fetch". If you look at the spack object, the install method isn't used (and arguably can probably be deleted from that object definition).

It's more like if install isn't defined, and we don't have a valid way to fetch it. But I figured it was easier to just allow objects to say "this should be in the image" rather than "you can install this".

var/ramble/repos/builtin/base_classes/utility-base/base_class.py
douglasjacobsen: Good call, I updated it to use our versions class instead.

lib/ramble/docs/dev_guides/utility_dev_guide.rst
linsword13: I see, nope the current implementation sounds fine then.

linsword13:
LGTM, reminder that there's a merge conflict.

douglasjacobsen:
Merge conflict should be resolved now.

linsword13 (dismissed):

linsword13 (approved):

PR #1591, opened 2026-05-29
Better support for setting dry-run for workspace cmds

linsword13:
It's often not sufficient to use `ws.dry_run = True`, given that invocations like `workspace("concretize")` do not expose the workspace handle used. To address this, this PR adds in an environment variable to configure the `dry_run` property of a workspace.

Example repro where previously not properly setting dry_run causes test failures under certain conditions (in this case, when `sinfo` is available):

```
$ mkdir -p /tmp/mock_bin
$ echo -e '#!/bin/bash\necho "hello"' > /tmp/mock_bin/sinfo
$ chmod +x /tmp/mock_bin/sinfo
$ PATH="/tmp/mock_bin:$PATH" ramble unit-test -k system_platform_general
```

### DIFF
--- a/lib/ramble/ramble/workspace/workspace.py
+++ b/lib/ramble/ramble/workspace/workspace.py
@@ -442,11 +442,11 @@ class Workspace:
     inventory_file_name = "ramble_inventory.json"
     hash_file_name = "workspace_hash.sha256"
 
-    def __init__(self, root, dry_run=False, read_default_template=True):
+    def __init__(self, root, dry_run=None, read_default_template=True):
         logger.debug(f"In workspace init. Root = {root}")
         self.root = ramble.util.path.canonicalize_path(root)
         self.txlock = lk.Lock(self._transaction_lock_path)
-        self.dry_run = dry_run
+        self._dry_run = dry_run
         self.repeat_success_strict = True
 
         self.read_default_template = read_default_template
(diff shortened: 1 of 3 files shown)


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -442,11 +442,11 @@ class Workspace:
     inventory_file_name = "ramble_inventory.json"
     hash_file_name = "workspace_hash.sha256"
 
-    def __init__(self, root, dry_run=False, read_default_template=True):
+    def __init__(self, root, dry_run=None, read_default_template=True):
```
douglasjacobsen, line 445: I'm curious why we pass None here instead of removing the argument?

It feels like there are two interfaces that could make it a bit more annoying to trace what is happening.

douglasjacobsen (commented):
I don't have any major concerns with this, but I do have one question. Generally, I think I'm curious why we want an env-var instead of making everything consistently accept the `--dry-run` argument.

linsword13:
> I don't have any major concerns with this, but I do have one question. Generally, I think I'm curious why we want an env-var instead of making everything consistently accept the `--dry-run` argument.

Indeed, what you proposed is better. I was using env-var mostly to avoid needing to add this argument to every workspace subcommands. Opened up https://github.com/GoogleCloudPlatform/ramble/pull/1599 to go with the proposed approach.

PR #1594, opened 2026-06-01
improve import error printing

rfbgo:


var/ramble/repos/builtin/package_managers/pip/package_manager.py
linsword13, line 1: These changes look great, but are they intended as a different PR?

rfbgo: Yes, thanks, good catch. I think gemini did a `-a`..

linsword13 (approved):

PR #1597, opened 2026-06-02
add initial openfold3

rfbgo:
Add pip version of openfold3 from https://github.com/aqlaboratory/openfold-3/tree/main/openfold3

Adds multiple inputs from the main repo 

| Experiment Name | Modality | Test Query | Inference Time (s) | Avg. pLDDT (%) | pTM | ipTM | GPDE (Å) |
| :--- | :--- | :--- | :---: | :---: | :---: | :---: | :---: |
| `ubiquitin_monomer` | Monomer | Ubiquitin | **19.78s** | **80.60%** | **0.71** | **0.65** | **0.85** |
| `homomer_complex` | Homomer | Leucine Zipper | **20.35s** | **80.60%** | **0.71** | **0.65** | **0.85** |
| `dna_ptm` | DNA PTM | ptm-DNA | **23.88s** | **91.53%** | **0.89** | **0.87** | **0.50** |
| `protein_ligand` | Protein-Ligand | mcl1 | **24.91s** | **76.60%** | **0.39** | **0.24** | **2.84** |
| `multimer_complex` | Multimer | 7cnx | **30.71s** | **91.53%** | **0.89** | **0.87** | **0.50**

### DIFF
--- a/var/ramble/repos/builtin/applications/openfold3/application.py
+++ b/var/ramble/repos/builtin/applications/openfold3/application.py
@@ -1,0 +1,294 @@
+# Copyright 2022-2026 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.appkit import *
+
+
+class Openfold3(ExecutableApplication):
+    """OpenFold3 is a PyTorch reproduction of AlphaFold 3 and AlphaFold 2,
+    highly optimized for fast protein structure prediction.
+    """
+
+    name = "openfold3"
+
+    tags("bioinformatics", "machine-learning", "protein-folding", "hpc")
+
+    maintainers("rfbgo")
+
+    # Define pip software requirement
+    with when("package_manager_family=pip"):
+        software_spec(
+            "torch",
+            pkg_spec="torch",
+        )
+        software_spec(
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/openfold3/application.py
```diff
@@ -0,0 +1,294 @@
...
+                            )
+                        elif "gp" in best_conf:
+                            f_out.write(
+                                f"Predicted Distance Error (GPDE) = {best_conf.get('gp', 0.0)}\n"
+                            )
+
+                logger.debug(
+                    f"Successfully wrote formatted timings and quality metrics to {timing_out_path}"
+                )
+                break
```
linsword13, line 292: Does this mean we don't need to iterate through all the timing_files?

linsword13 (approved):

var/ramble/repos/builtin/applications/openfold3/application.py
rfbgo: Yea I find this a bit odd too (I argued with gemini a bit about it), but I think it's actually ~correct

The semantic is more like "the data lives in one of the output files, and we aren't such which one". 

I think one can definitely argue this is "overly defensive" but the idea is something like "if this file didn't cause an exception we're fine and can exit". This is why we have to iterate over `conf_files` too 

My thinking was we can nail this down once we have a more "real" use case for the app, but I'm happy to simplify it now too 

A lot of the complexity comes in because this can "secretly" be a batch of runs. One option would be to strongly discourage that when using ramble

linsword13: I see, thanks for the detailed explanation! LGTM

linsword13 (approved):

PR #1601, opened 2026-06-04
add create command [AI]

rfbgo:


lib/ramble/ramble/cmd/create.py
```diff
@@ -0,0 +1,244 @@
...
+        help="the type of object definition to create",
+    )
+    subparser.add_argument(
+        "name",
+        nargs="?",
+        help="the name of the new definition (e.g., my-app)",
+    )
+    subparser.add_argument(
+        "-r",
+        "--repo",
```
douglasjacobsen, line 46: I wonder how hard it would be to support something like:

```
ramble create builtin.app.foo
```

And have the repo be picked (and the type) automatically from  that.

lib/ramble/ramble/cmd/create.py
```diff
@@ -0,0 +1,244 @@
...
+    # Fallback to generic classes if no matches found
+    if not matched_bases:
+        matched_bases = ["object"]
+
+    return matched_bases
+
+
+def run_interactive_wizard():
+    """Runs the step-by-step interactive wizard."""
+    print("\n" + "=" * 40)
```
douglasjacobsen, line 131: Does it make sense (I'm not sure...) for these to be logger calls instead of direct prints?

lib/ramble/ramble/test/cmd/create.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+def test_create_help():
+    """Verify create --help options."""
+    with pytest.raises(SystemExit):
+        create_cmd("--help")
+    assert create_cmd.returncode in (None, 0)
+
+
+def test_create_application_and_modifier(mutable_config, tmpdir):
```
douglasjacobsen: I wonder if the create tests can either be parameterized, or just collapsed into a single one?

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -50,15 +50,21 @@ def set_required_variables(self, app_inst=None):
...
         """
         for _, obj in app_inst.objects():
-            for pkgname in obj.required_packages:
-                app_inst.keywords.update_keys(
-                    {
-                        f"{pkgname}_path": {
-                            "type": ramble.keywords.key_type.required,
-                            "level": ramble.keywords.output_level.variable,
+            for pkgname, config in obj.required_packages.items():
+                if app_inst.expander.satisfies(
```
douglasjacobsen: I don't think this belongs in this PR, but this might be where you were running into this issue.

douglasjacobsen (changes requested):
Thanks, this is great. I have a few comments, but otherwise it looks good.

lib/ramble/ramble/cmd/create.py
rfbgo: Personally, I do not think we should do that and should rely on args over the scoping/"magic incantation" way

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
rfbgo: I agree, good catch!

lib/ramble/ramble/cmd/create.py
rfbgo: My take is no, since the logger is writing to a file, and this is a tty interaction. I also don't think we would capture the users responses in the logger (at least without extra effort), so logging it would not be so helpful

lib/ramble/ramble/test/cmd/create.py
rfbgo: Updated them to be paramterized

lib/ramble/ramble/cmd/create.py
douglasjacobsen: I'm OK with this for now. But it feels like, if we could do both that would be best. The args can feel a bit much when you want to specify everything, but having the magic incantation can also leave users confused about what they are seeing.

lib/ramble/ramble/cmd/create.py
douglasjacobsen: OK, thanks

douglasjacobsen (approved):

PR #1604, opened 2026-06-05
Improve performance of variant evaluation

dapomeroy:
Adds caching of variant set expansions to eliminate most expander calls in this method

### DIFF
--- a/lib/ramble/ramble/variants.py
+++ b/lib/ramble/ramble/variants.py
@@ -37,6 +37,11 @@ class VariantSet:
         self.experiment_variants = {}
         self.version_variants = {}
         self._set_cache = None
+        self._has_template_variants = False
+
+    def _invalidate_cache(self):
+        self._set_cache = None
+        self._has_template_variants = False
 
     def __str__(self):
         if not hasattr(self, "_str_indent"):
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/variants.py
```diff
@@ -323,12 +328,15 @@ def _expanded_set(self, expander: Optional[Expander] = None) -> set:
...
         """
 
-        if expander is None:
+        if expander is None or not self._has_template_variants:
             return self._set_cache
 
         expanded_set = set()
         for variant in self._set_cache:
-            expanded_set.add(expander.expand_var(variant))
+            if "{" in variant:
```
linsword13, line 336: Do we expect to support more complicated expressions for variants? For instance can variant expression be math expressions, or use any of the supported functions?

dapomeroy: If it's inside curly braces. When we send it to the expander, it gets parsed using `ast.parse(in_str, mode="eval")`. In `eval` mode, any string that contains an `=` is going to be treated as an assignment and will fail parsing with a syntax error. All variant strings are formatted as `<variant>=<value>` unless they're bool variants like `+<variant>`, so everything except bool variants is guaranteed to be returned as a string from the expander unless it contains curly braces. I can't think of any case where there would be complex expressions in a bool variant name that aren't expansion variables... can you?

linsword13 (approved):

PR #1605, opened 2026-06-05
convert open to always use uft8

rfbgo:
gemini code review seems pretty obsessed with this, so we should decide if we want this to be our standard moving forward..


linsword13 (changes requested):
I think this makes sense, good change!

A couple comments:

* Looks like there's ruff rule that enforces this as well: https://docs.astral.sh/ruff/rules/unspecified-encoding/. Perhaps we can add that to our ruff rule set to catch future cases?

* I am a bit unsure if we want to modify the vendor deps (the llnl utility in this case.)

rfbgo:
> I think this makes sense, good change!
> 
> A couple comments:
> 
> * Looks like there's ruff rule that enforces this as well: https://docs.astral.sh/ruff/rules/unspecified-encoding/. Perhaps we can add that to our ruff rule set to catch future cases?
> * I am a bit unsure if we want to modify the vendor deps (the llnl utility in this case.)

Thanks for spotting the vendored change. I didn't notice gemini sneaks that in during the iterations 

I will look into the ruff changes, but agreed it looks simple enough and worth while

linsword13 (approved):

PR #1609, opened 2026-06-09
make user managed path respect when

rfbgo:


var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
```diff
@@ -50,15 +50,21 @@ def set_required_variables(self, app_inst=None):
...
         """
         for _, obj in app_inst.objects():
-            for pkgname in obj.required_packages:
-                app_inst.keywords.update_keys(
-                    {
-                        f"{pkgname}_path": {
-                            "type": ramble.keywords.key_type.required,
-                            "level": ramble.keywords.output_level.variable,
+            for pkgname, config in obj.required_packages.items():
+                if app_inst.expander.satisfies(
```
douglasjacobsen, line 54: Are there similar changes that are needed in the other user-managed objects? If there are, I wonder if this means we should be unifying the user-managed objects a bit more than we currently do?

douglasjacobsen (commented):
This is a great change. Thanks. I have one question though.

var/ramble/repos/builtin/package_managers/user-managed/package_manager.py
rfbgo: oh that is possible.. this is a bug fix for something that was directly biting me but it could definitely have broader implications

douglasjacobsen: Looking at the other user-managed objects, none of them have similar logic in them. So, I think for now this is all we really need.

douglasjacobsen (changes requested):
Since you said you encountered this as a bug, can you add a test to make sure we don't hit this again? I'm curious what the failure scenario was (but regardless I think this is a good change).

rfbgo:
> Since you said you encountered this as a bug, can you add a test to make sure we don't hit this again? I'm curious what the failure scenario was (but regardless I think this is a good change).

The thing I hit specifically was a bit of an odd case, but this is a more general fix. It was something like:

```
    with when("package_manager_family=spack"):
        required_package("some_package")
```

and then the user package was asserting that `some_package_path` to be set despite it not being conditionally met by `with when("package_manager_family=user_managed")`

There is a small chance that is "intended" in some cases, since it helps the user know they must define it at setup time but it's not strictly what is expressed in the app def. WDYT? 

Happy to add a test if you agree with the fix in general 

edit: thinking about this a bit more I think this change is probably correct, but also it's likely most `required_package` should be outside of pm specific whens?

douglasjacobsen:
Oh, yeah actually the case you added was specifically to help prevent issues when using user-managed package managers.

Basically, if "some_package" is marked as required (even if by spack) then something in the application definition typically requires that install path be defined (not just the binary in the spack environment).

So, changing user-managed to not require that path is likely going to break something. For example:

WRF requires the `wrf` package:
https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/applications/wrf/application.py#L46

But this is required because it's used as `wrf_path` here:
https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/applications/wrf/application.py#L61

So, it being added under the spack package manager family when clause is a bit misleading because it's more generally required that `wrf_path` is defined (but it just happens to be auto-defined by the `wrf` package definition).

rfbgo:
Yea that's what I mean, but I think you can argue what we should say it something like:

```
    with when("package_manager_family=spack"):
        define_compiler("gcc14", pkg_spec="gcc@14.2.0")

        software_spec(
            "intel-mpi",
            pkg_spec="intel-oneapi-mpi@2021.17.2",
            compiler="gcc14",
        )

        software_spec(
                "wrfv4-{application::wrf::version}",
                pkg_spec="wrf@{application::wrf::version} build_type=dm+sm compile_type=em_real nesting=basic ~pnetcdf",
                compiler="gcc14",
       )

       #required_package("wrf") # Not here inside the pm=spack
    required_package("wrf") # here or in a different when if we think its true for pm=user-managed
```

douglasjacobsen:
Yeah, or even something like:

```
with when("package_manager_family=spack"):
  required_package("wrf")
  
required_variable("wrf_path")
```

Assuming the check for required variables happens *after* the package manager defines the path.

I think before this PR can be merged we'd need to have those things fixed as well, or else this breaks existing object definitions / assumptions about how they function.

PR #1623, opened 2026-06-15
Remove unused utils and corresponding tests

linsword13:


douglasjacobsen (approved):

lib/ramble/ramble/util/web.py
```diff
@@ -289,218 +263,6 @@ def remove_url(url, recursive=False):
     # Don't even try for other URL schemes.
 
 
-def _iter_s3_contents(contents, prefix):
```
douglasjacobsen, line 292: I don't think we should remove the s3 functionality actually (after reviewing this). External users could easily be using this, even though we are not testing it.

douglasjacobsen (changes requested):

douglasjacobsen (approved):

PR #1633, opened 2026-06-22
Define variants for repeats

douglasjacobsen:
This merge adds definitions for three new variants related to repeats.
 - is_repeat_parent
 - is_repeat_child
 - repeat_index

These can then be used by objects to conditionally apply definitions based on if the experiment is a repeat or not.


dapomeroy:
This looks good as soon as the Gemini review comment is resolved

rfbgo (approved):

PR #1639, opened 2026-06-25
Fix potential deadlock in tty.log writer_daemon [AI]

douglasjacobsen:
When `use_fds` falls back to `False` (which occurs when pytest replaces standard file descriptors with its  StringIO  streams for capture via  capsys ),  sys.stdout  was not being correctly closed upon context exit. This caused the pipe to leak and the main process to wait indefinitely on the daemon to exit.


linsword13:
Thanks for the fix. Personally I prefer the wholesale update from https://github.com/GoogleCloudPlatform/ramble/pull/1644 over this patch.

PR #1642, opened 2026-06-25
Add (exclude_)where args to workload directive [AI]

douglasjacobsen:
This merge adds support for filtering experiments at the workload directive level. This can be used to silently reduce the possible experiment definitions, without creating errors.

Tests and documentation are added as well.


lib/ramble/ramble/language/application_language.py
```diff
@@ -58,6 +58,8 @@ def workload(
     input=None,
     inputs=None,
     tags=None,
+    where=None,
+    exclude_where=None,
```
rfbgo, line 63: Are we worried about the change in this arg order? This can technically be breaking for some people right?  I think it should be rare, but if we know anyone who is likely to get broken we could provide backwards compat and depreciate warning?

douglasjacobsen: I don't know anyone who would currently break by this. Most people haven't adopted `when` yet (as it's not in main or a tag).

Technically, it is breaking, as when happens after the where arguments now. I only did it this way to make the interface consistent (as when is last in all of the rest of the interfaces). But I have pushed a changed to place the where args after when to avoid breaking other people's interfaces.

rfbgo (approved):

PR #1644, opened 2026-06-26
Update vendored llnl dependency from upstream spack

douglasjacobsen:
Pulled from Spack commit: a534e6456a11aae3f3306a964eb60a9fb2c4ee91


douglasjacobsen:
The codecov information is because we ignore the lines in this PR, so the test covers 0% of the lines as a result. But they are actually all tested...

linsword13:
> The codecov information is because we ignore the lines in this PR, so the test covers 0% of the lines as a result. But they are actually all tested...

Hm interesting, if we ignore the lines then I'd assume codecov should ignore that too...

Oh I think it's complaining about the one-line change in config.py.

linsword13:
Love the update, glad that it seems to be a straightforward drop-in. One question: do we want to pull from a tagged release instead of from the head?

rfbgo:
I like this solution a lot too.

You probably already know this, but this verison of `llnl`  supports >= python 3.8, since 3.7 does not have typing.Literal, so we would need to move our min version from 3.6 to 3.8

rfbgo:
It looks like if we go back to the tagged commit in v1.2.0 (or earlier) we would be OK with python 3.7 (and hopefully 3.6):

https://github.com/spack/spack/blob/v1.2.0/bin/spack#L30
https://github.com/spack/spack/blob/v1.2.0/lib/spack/spack/llnl/util/filesystem.py#L24

In v1.2.1 they started doing literal stuff but we might be able to take their vendored workaround:
https://github.com/spack/spack/blob/v1.2.1/lib/spack/spack/llnl/util/filesystem.py#L47

linsword13:
> I like this solution a lot too.
> 
> You probably already know this, but this verison of `llnl` supports >= python 3.8, since 3.7 does not have typing.Literal, so we would need to move our min version from 3.6 to 3.8

Oh that's a good point. I think ideally we still want to preserve our support for 3.6.

rfbgo:
I don't really understand what is going on with the failed github action, give it is mad about a file you don't touch and is deleted upstream (gh-pages-deploy.yaml). You might need to rebase

rfbgo (approved):

PR #1647, opened 2026-06-30
Add when argument to archive_pattern directive

douglasjacobsen:


lib/ramble/ramble/test/when.py
```diff
@@ -1015,6 +1015,124 @@ def test_register_builtin_when(workspace_name, include_builtin, builtin_found):
...
+
+        import ramble.filters
+
+        filters = ramble.filters.Filters()
+        for _, app_inst, _ in experiment_set.filtered_experiments(filters):
+            exp_dir = app_inst.expander.expand_var("{experiment_run_dir}")
+            os.makedirs(exp_dir, exist_ok=True)
+
+            # Write expected patterns
+            mod_expected = expected_pattern.replace("test_pattern_when", "test_pattern_when_mod")
```
dapomeroy: These expected/unexpected variations for mod/pm get constructed inside loops, both here and in verify, but are static strings. I think we could simplify by moving them outside of the loops, construct them once, then store as a single pair of lists. Something like:

```
       expected_patterns = [expected_pattern]
       unexpected_patterns = [unexpected_pattern]
       for pattern_variation in ["when_mod", "when_pm"]:
         expected_patterns.append(expected_pattern.replace("when", pattern_variation))
         unexpected_patterns.append(unexpected_pattern.replace("when", pattern_variation))
```

and then iterate over the lists
```
       for pat in expected_patterns:
       ...
       for pat in unexpected_patterns:
```

dapomeroy (changes requested):
This is great. One simplification in the test and resolve merge conflicts and I think it's ready

lib/ramble/ramble/test/when.py
douglasjacobsen: Thanks, this should be resolved now.

dapomeroy (approved):
LGTM

PR #1650, opened 2026-07-01
Fix typo in wrf application

dapomeroy:
Remove duplicate `success_criteria` block

### DIFF
--- a/var/ramble/repos/builtin/applications/wrf/application.py
+++ b/var/ramble/repos/builtin/applications/wrf/application.py
@@ -418,13 +418,6 @@ class Wrf(ExecutableApplication):
         fom_type=FomType.MEASURE,
     )
 
-    success_criteria(
-        "Complete",
-        mode="string",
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/wrf/application.py
```diff
@@ -418,13 +418,6 @@ class Wrf(ExecutableApplication):
         fom_type=FomType.MEASURE,
     )
 
-    success_criteria(
```
linsword13, line 421: Is this success_criteria identical for both V3 and V4? If so, would it make sense to remove the ones in the two `when` blocks, and keep this one instead?

dapomeroy: It is not identical. The ones in the `when` blocks are correct. This one is the v3 block and should not have been duplicated outside the `when`.

linsword13: Got it, thanks.

linsword13 (approved):

PR #1653, opened 2026-07-06
Bug: Fix success_criteria to track when clause precedence

tteggelit:
Currently, `success_criteria` allows for scoping via `when` clauses, but definitions of these criteria are not properly tracked in the `success_criteria` dict. In cases where multiple `when` clauses are satisfied and define the same criteria name, this leads to the last defined criteria winning. This fix properly tracks satisfying `when` conditions selecting the highest precedence criteria.

In the case where multiple `when` conditions are satisfied at the same precedence, ramble will now throw an error. Because of this, the WRF application was also fixed as it defined both an unscoped "Complete" criteria and one for specific WRF versions. The versioned definitions are mutually exclusive, but the unscoped criteria would always be satisfied resulting in a conflict between it and the satisfied versioned criteria.

### DIFF
--- a/lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
+++ b/lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
@@ -10,11 +10,19 @@ import os
 
 import pytest
 
+import ramble.workspace
+from ramble.error import RambleCommandError
 from ramble.main import RambleCommand
 
 # everything here uses the mock_workspace_path
-pytestmark = pytest.mark.usefixtures("mutable_config", "mutable_mock_workspace_path")
+pytestmark = pytest.mark.usefixtures(
+    "mutable_config",
+    "mutable_mock_workspace_path",
+    "mutable_mock_apps_repo",
+    "workspace_deactivate",
+)
 
+config = RambleCommand("config")
 workspace = RambleCommand("workspace")
 ramble_on = RambleCommand("on")
 
@@ -24,6 +32,7 @@ def test_success_criteria_precedence(mock_applications, make_workspace_from_conf
     Tests that YAML success_criteria takes precedence over
     object-defined criteria
     """
+    # TODO: Update once success_criteria has manage experiments command
     test_config = """
 ramble:
   variables:
@@ -61,33 +70,28 @@ ramble:
         assert "config::experiment::test_success = PASSED" in data
 
 
-def test_success_criteria_mutex_versions(mock_applications, make_workspace_from_config):
+def test_success_criteria_mutex_versions(workspace_name):
     """
     Tests that same named success_criteria defined in mutually exclusive
     when clauses are handled appropriately.
     """
-    test_config = """
-ramble:
-  variables:
-    processes_per_node: '1'
-    n_threads: '1'
-  applications:
-    success-criteria-conflicts@1.0.0:
-      workloads:
-        version_wl:
-          experiments:
-            pass-experiment:
-              variables:
-                n_nodes: 1
-  software:
-    packages: {}
-    environments: {}
-"""
-    ws, ws_name = make_workspace_from_config(test_config)
+    global_args = ["-w", workspace_name]
 
-    workspace("setup", global_args=["-w", ws_name])
-    ramble_on(global_args=["-w", ws_name])
-    workspace("analyze", "-f", "text", "json", "yaml", global_args=["-w", ws_name])
+    with ramble.workspace.create(workspace_name) as ws:
+        workspace(
+            "manage",
+            "experiments",
+            "success-criteria-conflicts@1.0.0",
+            "--wf",
+            "version_wl",
+            "--default-variable-value",
+            "1",
+            global_args=global_args,
+        )
+
+    workspace("setup", global_args=global_args)
+    ramble_on(global_args=global_args)
+    workspace("analyze", "-f", "text", "json", "yaml", global_args=global_args)
 
     for ext in ["txt", "json", "yaml"]:
         assert os.path.exists(os.path.join(ws.results_dir, f"results.latest.{ext}"))
@@ -99,33 +103,25 @@ ramble:
         assert "application::success-criteria-conflicts::test_version = PASSED" in data
 
 
-def test_success_criteria_multiple_satisfies(mock_applications, make_workspace_from_config):
+def test_success_criteria_multiple_satisfies(workspace_name):
     """
     Tests that an object, Application in this case, defining the same named
     success_criteria in multiple satisfying when conditions, will error out.
     """
-    test_config = """
-ramble:
-  variables:
-    processes_per_node: '1'
-    n_threads: '1'
-  applications:
-    success-criteria-conflicts:
-      variants:
-        force_pass: True
-      workloads:
-        success_str_wl:
-          experiments:
-            pass-experiment:
-              variables:
-                n_nodes: 1
-  software:
-    packages: {}
-    environments: {}
-"""
-    ws, ws_name = make_workspace_from_config(test_config)
+    global_args = ["-w", workspace_name]
 
-    from ramble.error import RambleCommandError
+    with ramble.workspace.create(workspace_name):
+        workspace(
+            "manage",
+            "experiments",
+            "success-criteria-conflicts",
+            "--wf",
+            "success_str_wl",
+            "--default-variable-value",
+            "1",
+            global_args=global_args,
+        )
+    config("add", "variants:force_pass:true", global_args=global_args)
 
     expected_err = (
         r"Success criteria '.*' in object '.*' is defined "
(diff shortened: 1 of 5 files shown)


var/ramble/repos/builtin/applications/wrf/application.py
linsword13, line 421: Is this what's done in https://github.com/GoogleCloudPlatform/ramble/pull/1650? Do you mind pulling in the latest develop, which should already cover this?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -3679,15 +3635,34 @@ def analysis_dicts(self, criteria_list):
...
+                                f"  1) {sorted(existing_when_set)}\n"
+                                f"  2) {sorted(when_set)}"
+                            )
+                        obj_satisfied_criteria[name] = (when_set, conf)
+
+                for criteria, (_, conf) in obj_satisfied_criteria.items():
+                    if criteria in resolved_criteria:
+                        continue
+
+                    resolved_criteria[criteria] = prec
```
linsword13: How is this `prec` used?

linsword13 (commented):
Would it be useful to add in some unit tests for the new/correct behavior?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
tteggelit: It's not so I took the Gemini advice and converted to a set.

var/ramble/repos/builtin/applications/wrf/application.py
tteggelit: It is and done!

tteggelit:
> Would it be useful to add in some unit tests for the new/correct behavior?

I added some unit tests to cover various precedence situations.

tteggelit:
@linsword13 I think with this latest push, it should be ready for a re-review.

tteggelit:
@linsword13 Implemented the Gemini code review suggestions, so I'll need one more re-review kicked off.

tteggelit:
@linsword13 I think this is now ready for final review.

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
```diff
@@ -0,0 +1,209 @@
...
+                n_nodes: 1
+              success_criteria:
+              - name: test_success
+                mode: string
+                match: 'SUCCESS'
+  software:
+    packages: {}
+    environments: {}
+"""
+    with ramble.workspace.create(workspace_name) as ws:
```
linsword13: nit: There's a pytest fixture `make_workspace_from_config` that may simplify some of these setup.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -3679,15 +3679,39 @@ def analysis_dicts(self, criteria_list):
...
                         variant_set=obj_inst.experiment_variants(),
                     ):
                         continue
 
+                    for name, conf in criteria_dict.items():
+                        if name in obj_satisfied_criteria:
+                            existing_when_set = obj_satisfied_criteria[name][0]
+                            logger.die(
+                                f"Success criteria '{name}' in object '{obj_inst.name}' is defined multiple times "
+                                f"under conflicting satisfied 'when' conditions:\n"
```
linsword13, line 3704: Just curious: does that mean this will not tolerate duplicate when?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -3719,12 +3743,13 @@ def analysis_dicts(self, criteria_list):
                             owning_object=obj_inst,
                         )
 
-        criteria_list.add_criteria(
-            scope="object_definitions",
-            name="_application_function",
-            mode="application_function",
-            owning_object=self,
-        )
+        if "_application_function" not in resolved_criteria:
```
linsword13, line 3746: Is this guard needed?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
tteggelit: Without the specificity functionality we talked about, if there are multiple satisfying when conditions at the same precedence level, it will trigger an error. This seems to be the current way of handling this. Allowing 'when' specificity would change this and only trigger an error if there were multiple satisfies at the same specificity.

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
tteggelit: The unit test I cribbed off of did it this way, but I can update it to use `make_worksapce_from_config` if that is the preferred method now.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
tteggelit: There may be no current code path that defines an `_application_function` scoped `success_criteria` currently, but I added this in defense in case there is. I don't believe `add_criteria()` does any kind of deduplication so if one already existed for some reason, there would be duplicate criteria added then.

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
tteggelit: OK. Just committed update to switch to `make_workspace_from_config`

linsword13 (dismissed):

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
```diff
@@ -0,0 +1,174 @@
...
+ramble_on = RambleCommand("on")
+
+
+def test_success_criteria_precedence(mock_applications, make_workspace_from_config):
+    """
+    Tests that YAML success_criteria takes precedence over
+    object-defined criteria
+    """
+    test_config = """
+ramble:
```
douglasjacobsen, line 37: Can you create a bug / task for adding a front-end command for managing success criteria?

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
```diff
@@ -0,0 +1,174 @@
...
+        assert "config::experiment::test_success = PASSED" in data
+
+
+def test_success_criteria_mutex_versions(mock_applications, make_workspace_from_config):
+    """
+    Tests that same named success_criteria defined in mutually exclusive
+    when clauses are handled appropriately.
+    """
+    test_config = """
+ramble:
```
douglasjacobsen: Can you replace the explicit YAML with calls to `workspace manage experiments`?

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
```diff
@@ -0,0 +1,174 @@
...
+        assert "application::success-criteria-conflicts::test_version = PASSED" in data
+
+
+def test_success_criteria_multiple_satisfies(mock_applications, make_workspace_from_config):
+    """
+    Tests that an object, Application in this case, defining the same named
+    success_criteria in multiple satisfying when conditions, will error out.
+    """
+    test_config = """
+ramble:
```
douglasjacobsen: Same replace question here

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
```diff
@@ -0,0 +1,174 @@
...
+        workspace("setup", global_args=["-w", ws_name])
+
+
+def test_success_criteria_inheritance(mock_applications, make_workspace_from_config):
+    """
+    Tests that same named success_criteria defined in multiple objects (using
+    inheritance) is handled appropriately.
+    """
+    test_config = """
+ramble:
```
douglasjacobsen: Same replace question here.

douglasjacobsen (changes requested):

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
tteggelit: For sure, but can you help me understand what you mean a little bit more what you mean?

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
tteggelit: Can you point me at example tests that do it this way? The ones I found and cribbed off of for these used this YAML approach.

douglasjacobsen: Sure, this is the main one I would point out: https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/test/end_to_end/known_applications.py#L33-L95

It is used to test *every* app definition in our repo.

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
douglasjacobsen: Sure, we have a set of front-end commands (`ramble workspace manage`) that can be used to manipulate various aspects of a workspace's config. Currently, you can't do this with success criteria (i.e. there isn't a command you can use to edit `ramble:applications:foo:workload:success_criteria`). So, it'd be good to track this as a gap and make sure we try to implement it at some point (not now).

lib/ramble/ramble/test/success_criteria_functionality/success_precedence.py
tteggelit: @douglasjacobsen I've updated the 3 tests that can use `manage experiments`. The 4th will have to wait until there is a `success_criteria` command added to `manage experiments`. I'll resolve the other 2 comments and leave this one open until you've had a chance to look over the changes.

douglasjacobsen (approved):

PR #1657, opened 2026-07-14
Validate Scaling for the OSU Microbenchmarks

nhanford:
Currently, the `application.py` for the OSU Microbenchmarks does not validate workloads that must be run on 2 and only 2 ranks. This modification will add a validator to ensure that these workloads are not attempted for invalid rank scaling. For example, attempting to run 10,000 ranks of `osu_bibw` should fail instead of attempting to run.


nhanford:
I have confirmed Lawrence Livermore National Security, LLC has a Contributor License Agreement in place with Google/Alphabet per our General Counsel. We do not, however, use Google Groups or Google Apps for Business. Is this now a hard requirement for PRs?

douglasjacobsen:
@nhanford I dot not believe so. I triggered a re-run of the CLA check though, so lets see what happens.

rfbgo:
It looks like Doug's recheck never came back, so I just forced it and it came back as a failure

```
2026-07-20 08:57:06 | Re-run Requested | failure
```

@douglasjacobsen there is a process for us to request an override.. it does look like we're allowed to accept code from LLNL

douglasjacobsen (approved):

PR #1664, opened 2026-07-16
Fix-up and minor improvements

linsword13:
These are things Gemini found, and I've vetted them for correctness.


dapomeroy (approved):
LGTM

PR #1665, opened 2026-07-17
Add Slurm license support in workflow manager

linsword13:
The default behavior works by:

1. Check for the presence of matching license in the config
2. If license info is found, assume the app instance requires license, and use the license name to specify `{license_name}:{n_ranks}` for the sbatch license header
3. This can be overridden via the `slurm_licenses` workflow manager variable

This is the "client" side, in practice, this expects the Slurm cluster to configure the license limit properly.

### DIFF
--- a/var/ramble/repos/builtin/workflow_managers/slurm/test/slurm.py
+++ b/var/ramble/repos/builtin/workflow_managers/slurm/test/slurm.py
@@ -314,3 +314,82 @@ ramble:
             filename,
         )
         assert os.path.isfile(archived_path)
+
+
+def test_slurm_workflow_licenses(
+    mutable_mock_apps_repo, make_workspace_from_config
+):
+    test_config = """
+ramble:
+  variants:
+    workflow_manager: slurm
+  variables:
+    processes_per_node: 5
+    n_nodes: 2
+  applications:
+    basic:
+      workloads:
+        test_wl:
+          experiments:
+            test_no_lic_file: {}
+            test_with_lic_file: {}
+            test_lic_override:
+              variables:
+                slurm_licenses: 'custom_lic:99'
+            test_lic_disable:
+              variables:
+                slurm_licenses: ''
+"""
+
+    ws, _ = make_workspace_from_config(test_config)
+    workspace("setup", "--dry-run", global_args=["-D", ws.root])
+
+    path_no_lic = os.path.join(
+        ws.experiment_dir, "basic", "test_wl", "test_no_lic_file"
+    )
+    with open(
+        os.path.join(path_no_lic, "slurm_experiment_sbatch"), encoding="utf-8"
+    ) as f:
+        content = f.read()
+        assert "--licenses" not in content
+
+    test_licenses = """
+licenses:
+  basic:
+    set:
+      TEST_VAR: test_val
+"""
+    license_path = os.path.join(ws.config_dir, "licenses.yaml")
+    with open(license_path, "w+", encoding="utf-8") as f:
+        f.write(test_licenses)
+
+    workspace("setup", "--dry-run", global_args=["-D", ws.root])
+
+    path_with_lic = os.path.join(
+        ws.experiment_dir, "basic", "test_wl", "test_with_lic_file"
+    )
+    with open(
+        os.path.join(path_with_lic, "slurm_experiment_sbatch"),
+        encoding="utf-8",
+    ) as f:
+        content = f.read()
+        assert "#SBATCH --licenses=basic:10" in content
+
+    path_override = os.path.join(
+        ws.experiment_dir, "basic", "test_wl", "test_lic_override"
+    )
+    with open(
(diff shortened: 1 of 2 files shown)


var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -202,6 +209,26 @@ def template_render_vars(self):
...
+            license_conf = config.get("licenses")
+            if license_conf:
+                matched_license = next(
+                    (
+                        lic
+                        for lic in app_inst.license_names
+                        if lic in license_conf
+                    ),
+                    None,
+                )
```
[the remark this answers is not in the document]
linsword13: `license_names` at least includes the application name itself. So it's guaranteed to exist as a list.

douglasjacobsen:
I don't have any issues with this, but there is a merge conflict that needs to be resolved.

linsword13:
> I don't have any issues with this, but there is a merge conflict that needs to be resolved.

Ah thanks for the reminder. Resolved the conflict.

var/ramble/repos/builtin/workflow_managers/slurm/test/slurm.py
```diff
@@ -314,3 +314,82 @@ def test_slurm_archive_patterns(make_workspace_from_config):
             filename,
         )
         assert os.path.isfile(archived_path)
+
+
+def test_slurm_workflow_licenses(
+    mutable_mock_apps_repo, make_workspace_from_config
+):
+    test_config = """
+ramble:
```
douglasjacobsen, line 323: nit; Can we replace this with a `manage experiments` config?

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -204,6 +211,26 @@ def template_render_vars(self):
...
+                if matched_license:
+                    n_ranks = expander.expand_var("{n_ranks}")
+                    slurm_licenses = f"{matched_license}:{n_ranks}"
+
+        if slurm_licenses:
+            pragmas.append(f"#SBATCH --licenses={slurm_licenses}")
```
douglasjacobsen, line 232: What is the behavior if the backend portion of this isn't implemented? Is this just ignored, or do we get an error that the license isn't found?

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -204,6 +211,26 @@ def template_render_vars(self):
...
+                matched_license = next(
+                    (
+                        lic
+                        for lic in app_inst.license_names
+                        if lic in license_conf
+                    ),
+                    None,
+                )
+                if matched_license:
+                    n_ranks = expander.expand_var("{n_ranks}")
```
douglasjacobsen, line 228: I think we need a way to override the math for the licenses. Some apps are not just `{n_ranks}` and have more complex formulas invlolving `{n_threads}` as well.

I think in general, this is a cool idea. But I think how we implement the license token bit probably needs some design work to make it easy to do within application definitions.

douglasjacobsen (changes requested):

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Yeah so my thinking is that this is the default behavior. If an app needs specific math on license count, then it can supply a custom `slurm_licenses` variable.

I've thought about having some kind of app-level directive to describe the license count calculation, but I am hesitant as such license logic may tie more to the kind of license one has, instead of inherent to the app itself.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: We will actually get an error. To skip the error, one needs to specify `slurm_licenses: ''`

douglasjacobsen: OK, in that case maybe it would be nice to have a variant (that defaults to false) to enable this behavior? It'd be unfortunate if licensed apps just started to not work because we don't have licensed defined in the slurm cluster.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
douglasjacobsen: OK thanks, that sounds reasonable.

var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
linsword13: Thinking more about this, I feel that in its current form this isn't that useful in practice. This requires more up-front design work to have something more helpful. Closing out this PR for now.

PR #1668, opened 2026-07-20
Add a multichase application

linsword13:
The corresponding spack package has been merged: https://github.com/spack/spack-packages/pull/5543.
    
As an example, to generate a study of memory hierarchy latency plot:
    
1. Create a workspace with the following config:
    
    ```
    ramble:
      variants:
        package_manager: spack
        workflow_manager: slurm
      variables:
        n_nodes: 1
        processes_per_node: 1
        slurm_partition: h4dhighmemlssd
        extra_sbatch_headers: |
          #SBATCH --nodelist=hpcperf-h4dnodeset-0
      applications:
        multichase:
          workloads:
            latency:
              experiments:
                size_{array_size}:
                  variables:
                    array_size: [4k, 8k, 16k, 32k, 64k, 128k, 256k, 512k, 1m, 2m, 4m, 8m, 16m, 32m, 64m, 128m, 256m, 512m, 1g]
      software:
        packages:
          gcc15:
            pkg_spec: gcc@15.2.0
          multichase:
            pkg_spec: multichase@master target=zen5
            compiler: gcc15
        environments:
          multichase:
            packages:
              - multichase
    
    ```
    
2. Run the experiments and generate the json results
    
    ```
    ramble workspace setup
    ramble on
    ramble workspace analyze -f json
    ```
    
3. Generate the report and the plot
    
    ```
    ramble results report --multi-line "Multichase Best Latency" "Array Size" --logx --logy
    ```
    
The generated plot is attached.
multi_line_Multichase-Best-Latency_vs_Array-Size_all-series

### DIFF
--- a/var/ramble/repos/builtin/applications/multichase/application.py
+++ b/var/ramble/repos/builtin/applications/multichase/application.py
@@ -1,0 +1,118 @@
+# Copyright 2022-2026 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.appkit import *
+
+_LATENCY_REGEX = r"^\s*(?P<latency>[0-9\.]+)\s*$"
+
+
+class Multichase(ExecutableApplication):
+    """Multichase pointer chaser benchmark.
+
+    Includes a latency workload that benchmarks latency for a given array size.
+    """
+
+    name = "multichase"
+
+    maintainers("linsword")
+
+    tags("benchmark", "memory", "latency", "bandwidth")
+
+    with when("package_manager_family=spack"):
+        define_compiler("gcc", pkg_spec="gcc")
+        software_spec(
+            "multichase",
+            pkg_spec="multichase",
+            compiler="gcc",
+        )
+        required_package("multichase")
+
+    executable(
+        "execute",
+        "{multichase_bin} -m {array_size} {additional_args}",
+        use_mpi=False,
+    )
+
+    # Workloads
+    workload("latency", executable="execute")
+
+    # Workload variables for latency
+    workload_variable(
+        "multichase_bin",
+        default="multichase",
+        description="Path or binary name for multichase",
+        workloads=["latency"],
+    )
+    workload_variable(
+        "array_size",
+        default="1g",
+        description="Array size for pointer chasing",
+        workloads=["latency"],
+    )
+    workload_variable(
+        "stride",
+        default="256",
+        description="Stride size in bytes",
+        workloads=["latency"],
+    )
+    workload_variable(
+        "threads",
+        default="1",
+        description="Number of threads",
+        workloads=["latency"],
+    )
+    workload_variable(
+        "samples",
+        default="5",
+        description="Number of 0.5 second samples",
+        workloads=["latency"],
+    )
+    workload_variable(
+        "additional_args",
+        default="-s {stride} -t {threads} -n {samples}",
+        description="Additional arguments for multichase",
+        workloads=["latency"],
+    )
+
+    register_validator(
+        name="single_node",
+        predicate="{n_nodes} == 1",
+        message=(
+            "The multichase application is intended to be used on a single "
+            "node, but is configured with n_nodes = {n_nodes}"
+        ),
+        fail_on_invalid=False,
+    )
+
+    with when("workload_name=latency"):
+        figure_of_merit(
+            "Multichase Best Latency",
+            fom_regex=_LATENCY_REGEX,
+            group_name="latency",
+            units="ns",
+        )
+        figure_of_merit(
+            "Array Size",
+            fom_map_key="array_size_bytes",
+            units="bytes",
+        )
+
+    def _prepare_analysis(self, workspace, app_inst=None):
+        def _parse_size_bytes(size_str):
+            s = str(size_str).strip().lower()
+            units = {"k": 1024, "m": 1024**2, "g": 1024**3, "t": 1024**4}
+            if s and s[-1] in units:
+                return int(float(s[:-1]) * units[s[-1]])
+            return int(float(s))
+
+        array_size_str = self.expander.expand_var("{array_size}")
+        try:
+            array_size_bytes = _parse_size_bytes(array_size_str)
+            self.add_inmem_fom_value("array_size_bytes", array_size_bytes)
+        except ValueError:
+            pass


var/ramble/repos/builtin/applications/multichase/application.py
```diff
@@ -0,0 +1,222 @@
...
+
+            fig, ax = plt.subplots(figsize=(10, 6))
+            ax.plot(
+                sizes_b,
+                latencies,
+                marker="o",
+                color="b",
+                linewidth=2,
+            )
+            ax.set_xscale("log", base=2)
```
rfbgo: Is it possible to generate this plot with reports instead? If we keep `matplotlib` maybe we should use the headless backend? https://matplotlib.org/stable/users/explain/figure/backends.html#selecting-a-backend

var/ramble/repos/builtin/applications/multichase/run_multichase.sh.tpl
```diff
@@ -0,0 +1,23 @@
...
+        g|G) echo $((num * 1024 * 1024 * 1024)) ;;
+        t|T) echo $((num * 1024 * 1024 * 1024 * 1024)) ;;
+        *) echo "$val" ;;
+    esac
+}
+
+cur=$(parse_bytes "{min_array_size}")
+max=$(parse_bytes "{max_array_size}")
+
+while [ "$cur" -le "$max" ]; do
```
rfbgo: Do we need some guards/sanity check to stop this being an infinite loop (eg min_array_size=0 will be dangerous)

var/ramble/repos/builtin/applications/multichase/run_multichase.sh.tpl
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
```
rfbgo: This might be more of a general question, but do you think we should be running bash in strict mode for app tpls? Something like `set -euo pipefail` to defend against unset vars?

var/ramble/repos/builtin/applications/multichase/run_multichase.sh.tpl
```diff
@@ -0,0 +1,23 @@
...
+parse_bytes() {
+    val="$1"
+    unit="${val: -1}"
+    num="${val%[kKmMgGtT]}"
+    case "$unit" in
+        k|K) echo $((num * 1024)) ;;
+        m|M) echo $((num * 1024 * 1024)) ;;
+        g|G) echo $((num * 1024 * 1024 * 1024)) ;;
+        t|T) echo $((num * 1024 * 1024 * 1024 * 1024)) ;;
```
rfbgo: nit: I was surprised to see we don't print num here, would it be useful?

rfbgo (commented):
LGTM. Few nits and questions

var/ramble/repos/builtin/applications/multichase/run_multichase.sh.tpl
rfbgo: Just to capture the offline discussion we had offline: should we be calling the app in a loop here, or relying on ramble to generate the different configurations at the workspace level

linsword13: Ack. Will convert this into a draft PR until the change is done.

var/ramble/repos/builtin/applications/multichase/application.py
linsword13: Resolving since this has been changed.

var/ramble/repos/builtin/applications/multichase/run_multichase.sh.tpl
linsword13: Resolving since this has been changed.

var/ramble/repos/builtin/applications/multichase/run_multichase.sh.tpl
linsword13: Resolving since this has been changed.

var/ramble/repos/builtin/applications/multichase/run_multichase.sh.tpl
linsword13: Thanks for the suggestion. I've updated this to have the application logic be more "atomic". The PR description includes an example of how to perform a hierarchy latency study and use report to generate the plot.

var/ramble/repos/builtin/applications/multichase/run_multichase.sh.tpl
linsword13: Resolving as this has been changed.

var/ramble/repos/builtin/applications/multichase/application.py
```diff
@@ -0,0 +1,118 @@
...
+            units="bytes",
+        )
+
+    def _prepare_analysis(self, workspace, app_inst=None):
+        def _parse_size_bytes(size_str):
+            s = str(size_str).strip().lower()
+            units = {"k": 1024, "m": 1024**2, "g": 1024**3, "t": 1024**4}
+            if s and s[-1] in units:
+                return int(float(s[:-1]) * units[s[-1]])
+            return int(float(s))
```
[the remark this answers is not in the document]
linsword13: This is not an issue since this single-character format is the only one supported by the multichase app itself.

It could be a nice extension, but for now skipping it.

rfbgo:
App def looks good to me. @dapomeroy might want to review the reports changes so not merging just yet

lib/ramble/ramble/reports.py
```diff
@@ -523,7 +523,19 @@ def extract_data(experiments: List[dict], foms: List[str], variables: List[str],
                                 elif var in _ADDITIONAL_VARS:
                                     continue
                                 else:
-                                    logger.debug(f"{var} not found in the results data. Skipping.")
+                                    fom_found = False
+                                    for exp_ctx in exp.get("CONTEXTS", []):
+                                        for exp_fom in exp_ctx.get("foms", []):
+                                            if exp_fom.get("name") == var:
+                                                exp_data[var] = exp_fom.get("value")
```
dapomeroy: I'd like to understand the use case here a little better. I'm trying to figure out if there is a scenario where a value that is found in `exp_ctx` would conflict with the parent `context` from L489 that it's being saved under.

linsword13: So in my case the variable `array_size` could be passing in things like `1m`, `4k`, etc., and in my `_prepare_analysis` I convert these into unit of bytes, and the converted value is stored as a FOM. And I want to use this FOM as my x-axis.

linsword13: Synced with David offline on this. David will open a separate PR for the report side of change. I've now backed out the report related changes and only include in the multichase app.

dapomeroy (approved):
LGTM. Updates to `ramble result report` to enable FOMs as scaling vars, as demonstrated with this app, are in PR #1688

rfbgo (approved):

PR #1669, opened 2026-07-20
Allow archive untarred workspace

linsword13:
When uploading to a GCS bucket, sometimes uploading an uncompressed directory is faster than uploading tar file, since the former can upload large objects in parallel.

### DIFF
--- a/lib/ramble/ramble/util/web.py
+++ b/lib/ramble/ramble/util/web.py
@@ -6,6 +6,7 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
+import concurrent.futures
 import errno
 import os
 import os.path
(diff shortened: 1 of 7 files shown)


lib/ramble/ramble/util/web.py
```diff
@@ -167,6 +168,65 @@ def push_to_url(local_file_path, remote_path, keep_original=True, extra_args=Non
...
 
 
+def push_dir_to_url(local_dir_path, remote_dir_path, max_workers=None):
+    """Upload an entire directory tree recursively to a remote URL in parallel."""
+    if max_workers is None:
+        max_workers = ramble.config.get("config:upload_threads")
+
+    remote_url = url_util.parse(remote_dir_path)
+
+    if remote_url.scheme == "gs":
```
rfbgo, line 200: Should we make this check more robust so if someone asks for a scheme we don't support we can bail out early and avoid gathering files in the loop below?

lib/ramble/ramble/util/web.py
```diff
@@ -167,6 +168,65 @@ def push_to_url(local_file_path, remote_path, keep_original=True, extra_args=Non
...
+        from google.cloud.storage import transfer_manager
+
+        gcs_bucket = gcs_util.GCSBucket(remote_url)
```
rfbgo, line 203: nit/fun: I asked gemini if we have to change things to add s3 support and it seemed willing to.. not sure if we want that or not but it would be nice to support as many backends as possible

lib/ramble/ramble/util/web.py
linsword13: Right now this `if` block serves as a fast-path for the `gs` protocol, and for others, the fallback is to invoke `push_to_url` concurrently for the files. So it still supports other protocols like `s3` (via `push_to_url`), just that it invokes the `gs` batch uploader for more efficient handling.

lib/ramble/ramble/util/web.py
linsword13: Related to the comment above, we do still support s3, just that we don't have a fast-path for it.

I think it's great to add better support for it. The only thing is I don't have a setup to test out things, so I am hesitant to do it with the current PR.

lib/ramble/ramble/util/web.py
rfbgo: OK makes sense, thanks. What does it fail now if we ask for a non supported protocol? It would be nice if that is as early as possible

lib/ramble/ramble/util/web.py
rfbgo: OK thanks, I agree!

lib/ramble/ramble/util/web.py
linsword13: That's a great suggestion. I re-structured the code so that it fails faster upon unsupported protocol.

rfbgo: Awesome, thank you for taking the time to do that! LGTM

rfbgo (dismissed):

rfbgo:
That's interesting, how did you get this break?!

```
/workspace/lib/ramble/docs/ramble.util.rst:277: WARNING: py:class reference target not found: ParseResult [ref.class]
/workspace/lib/ramble/docs/ramble.util.rst:277: WARNING: py:class reference target not found: ParseResult [ref.class]
```

Looks like gemini mislead you about some type hint? :) 

https://github.com/GoogleCloudPlatform/ramble/pull/1669/changes#diff-b9a947c2ae0ec5831397f0616e5388455e2c6d5045bbddcf192d2f45987f492fR120

linsword13:
> That's interesting, how did you get this break?!
> 
> ```
> /workspace/lib/ramble/docs/ramble.util.rst:277: WARNING: py:class reference target not found: ParseResult [ref.class]
> /workspace/lib/ramble/docs/ramble.util.rst:277: WARNING: py:class reference target not found: ParseResult [ref.class]
> ```
> 
> Looks like gemini mislead you about some type hint? :)
> 
> https://github.com/GoogleCloudPlatform/ramble/pull/1669/changes#diff-b9a947c2ae0ec5831397f0616e5388455e2c6d5045bbddcf192d2f45987f492fR120

Yeah looks like the doc build didn't like the plain `ParseResult` reference. Updated now to the full reference. Thanks for the catch!

rfbgo (approved):

PR #1672, opened 2026-07-21
fix issue where bootstaped spack was not used during workspace setup

rfbgo:
Resolves conflicts between Ramble's bootstrapped Spack (v0.22.0) and the user's home Spack
  installation:

(main fix)
  1. Fix Executable Path Resolution (command_runner.py): Updated CommandRunner to search for
  executables using the PATH provided in the env argument, rather than always falling back to the
  system PATH.

(caching issue)
  2. Invalidate Runner Cache (package_manager.py): Updated the runner property to invalidate and
  recreate the cached Spack runner when the execution environment changes (ensuring it switches to the
  bootstrapped path after the bootstrap phase).

### DIFF
--- a/lib/ramble/ramble/test/mock_spack_runner.py
+++ b/lib/ramble/ramble/test/mock_spack_runner.py
@@ -26,6 +26,7 @@ class MockPackageInfo:
 
 class MockSpackRunner:
     def __init__(self, **kwargs):
+        self.env = kwargs.get("env", None)
         self.env_path = None
         self.concretized = False
         self.dry_run = False
--- a/lib/ramble/ramble/util/command_runner.py
+++ b/lib/ramble/ramble/util/command_runner.py
@@ -44,11 +44,12 @@ class CommandRunner:
         if command is None:
             self.command = None
         else:
+            search_path = path if path is not None else (env.get("PATH") if env else None)
             try:
-                if path is None:
-                    self.command = which(command, required=required)
+                if search_path is not None:
+                    self.command = which(command, required=required, path=search_path)
                 else:
-                    self.command = which(command, required=required, path=path)
+                    self.command = which(command, required=required)
             except CommandNotFoundError:
                 raise RunnerError(f"Command {name} is not found in path") from None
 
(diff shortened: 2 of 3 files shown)


var/ramble/repos/builtin/utilities/spack/utility.py
```diff
@@ -53,6 +53,7 @@ class Spack(UtilityBase):
 
     env_source("{utility::spack::path}/share/spack/setup-env.sh")
     env_set("SPACK_USER_CONFIG_PATH", "{utility::spack::path}/../.spack")
+    env_set("SPACK_USER_CACHE_PATH", "{utility::spack::path}/../.spack")
```
douglasjacobsen: I actually don't actually think this is a good change. My read of this env-var is that it controlls where tar / git repos are cloned, and it should be organized based on SHA.

If this is stored in the workspace specific spack installation, the chace is basically useless across multiple installations.

douglasjacobsen (changes requested):

var/ramble/repos/builtin/utilities/spack/utility.py
rfbgo: OK sounds good, I can remove it. I agreed it seemed a bit odd but I think the bad read of ~/.spack seemed like a reasonable concern. I also wonder if we should bump the default version from 0.22 to >= 1.0

douglasjacobsen: Bumping the default is fine with me. If people haven't already installed spack, we should probably give them what we 're using mostly (so maybe even 1.0.2?)

rfbgo: Ya I have a slight preference that we would be close-ish to what someone gets on develop, so I'd prefer to be on the same major version (ie >=1.0) if possible 

I think mixing 0.22 and =>1.0 is more likely to give conflicts in shared config files, but I've not looked into it in detail

`SPACK_USER_CACHE_PATH` is now removed btw

douglasjacobsen (approved):

PR #1674, opened 2026-07-22
Heffte application

tteggelit:
This PR adds support for the [heFFTe library](https://github.com/icl-utk-edu/heffte) and its 4 benchmark utilities: `convolution`, `speed3d_c2c`, `speed3d_r2c`, and `speed3d_r2r`. It depends on [spack-packages PR #5262](https://github.com/spack/spack-packages/pull/5262) as that adds the `benchmarks` variant to the heFFTe spack package to install the benchmarks. 

heFFTe should be a useful application for multiple reasons:
- It supports both CPUs and GPUs
- It is both strong- and weak-scaling
- It can leverage multiple FFT backends for analysis - a stock builtin FFT, fftw, Intel MKL FFT, Intel oneAPI FFT, and cuFFT
- It can use multiple different MPI collectives

### DIFF
--- a/var/ramble/repos/builtin/applications/heffte/application.py
+++ b/var/ramble/repos/builtin/applications/heffte/application.py
@@ -34,6 +34,7 @@ class Heffte(ExecutableApplication):
         software_spec(
             "intel-mpi",
             pkg_spec="intel-oneapi-mpi@2021.17.2",
+            compiler="gcc14",
         )
 
         software_spec(
--- a/var/ramble/repos/builtin/applications/heffte/application.py
+++ b/var/ramble/repos/builtin/applications/heffte/application.py
@@ -1,0 +1,340 @@
+# Copyright 2022-2026 The Ramble Authors
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+from ramble.appkit import *
+
+
+class Heffte(ExecutableApplication):
+    """
+    The Highly Efficient FFT for Exascale (heFFTe) library delivers
+    algorithms for distributed fast-Fourier transforms on heterogeneous
+    systems.
+    """
+
+    name = "heffte"
+
+    tags("fft-benchmark", "communication-benchmark")
+
+    maintainers("tteggelit")
+
+    version("2.4.1", "Version 2.4.1 of the heFFTe library", preferred=True)
+    version("2.4.0", "Version 2.4.0 of the heFFTe library")
+    version("2.3.0", "Version 2.3.0 of the heFFTe library")
+    version("2.2.0", "Version 2.2.0 of the heFFTe library")
+    version("2.1.0", "Version 2.1.0 of the heFFTe library")
+
+    with when("package_manager_family=spack"):
+        define_compiler("gcc14", pkg_spec="gcc@14.2.0")
+
+        software_spec(
+            "intel-mpi",
+            pkg_spec="intel-oneapi-mpi@2021.17.2",
+        )
+
+        software_spec(
+            "heffte-{application::heffte::version}",
+            pkg_spec="heffte@{application::heffte::version} "
+            + "{application::variant::benchmarks} "
+            + "{application::variant::fftw} "
+            + "{application::variant::cuda} "
+            + "{application::variant::mkl}",
+            compiler="gcc14",
+        )
+
+        required_package("heffte")
+
+    variant(
+        "benchmarks",
+        values=[True, False],
+        default=True,
+        description="Install the heFFTe benchmarks. Required to run benchmarks",
+    )
+
+    variant(
+        "fftw",
+        default=False,
+        values=[True, False],
+        description="fftw support variant",
+    )
+
+    variant(
+        "cuda",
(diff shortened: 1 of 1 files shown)


var/ramble/repos/builtin/applications/heffte/application.py
```diff
@@ -0,0 +1,420 @@
...
+            group_name="error",
+            units="",
+            fom_type=FomType.INFO,
+        )
+
+        success_criteria(
+            "Within Tolerance",
+            mode="string",
+            anti_match=r"^ERROR: observed error after heFFTe benchmark exceeds the tolerance$",
+        )
```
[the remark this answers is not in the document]
tteggelit: Even though I've not seen tolerance use scientific notation or max error have a positive exponent, I made the re changes since the heFFTe code appears to leave these cases open to encounter. The mentioned FOMs and success criteria do not apply to the `convolution` workload so those need to remain behind `when()` clauses. Once `workload_groups` are valid when clauses, they can be cleaned up better.

var/ramble/repos/builtin/applications/heffte/application.py
```diff
@@ -0,0 +1,420 @@
...
+    version("2.2.0", "Version 2.2.0 of the heFFTe library", preferred=False)
+    version("2.1.0", "Version 2.1.0 of the heFFTe library", preferred=False)
+
+    with when("package_manager_family=spack"):
+        define_compiler("gcc14", pkg_spec="gcc@14.2.0")
+
+        software_spec(
+            "intel-mpi",
+            pkg_spec="intel-oneapi-mpi@2021.17.2",
+            compiler="gcc14",
```
linsword13: nit: impi doesn't need a compiler specified?

var/ramble/repos/builtin/applications/heffte/application.py
```diff
@@ -0,0 +1,420 @@
...
+    """
+
+    name = "heffte"
+
+    tags("fft-benchmark", "communication-benchmark")
+
+    maintainers("tteggelit")
+
+    version("2.4.1", "Version 2.4.1 of the heFFTe library", preferred=True)
+    version("2.4.0", "Version 2.4.0 of the heFFTe library", preferred=False)
```
linsword13: It's probably cleaner to leave the `preferred=False` out.

var/ramble/repos/builtin/applications/heffte/application.py
tteggelit: Removed. I'll make a bug to remove that from other applications for consistency.

var/ramble/repos/builtin/applications/heffte/application.py
tteggelit: Removed. I'll make a bug to remove that from other applications for consistency.

var/ramble/repos/builtin/applications/heffte/application.py
```diff
@@ -0,0 +1,419 @@
...
+    success_criteria(
+        "Time per Run",
+        mode="fom_comparison",
+        fom_name="Time per run",
+        formula="float({value}) > 0.0",
+    )
+
+    # TODO: Once workload_groups are valid when conditionals, collapse the following 3
+    with when("workload_name=c2c"):
+        figure_of_merit(
```
linsword13: Just curious, it seems some of these per-workload FOMs have the same definition for all 3 workloads. Would it make sense to have them shared across all these workloads?

tteggelit: They do, but the convolution workload doesn't produce those figure of merit. Once the feature to use workload_groups as when clauses is added, these could be collapsed into one when clause.

linsword13: I see. And yes the workload_group variant PR is in: https://github.com/GoogleCloudPlatform/ramble/pull/1675

tteggelit: If we want to wait to merge this until that PR is merged, I'm happy to do that.

linsword13: That PR is already merged. You can see it if you rebase.

tteggelit: Well I totally missed that. 🤦 I'll rebase now.

tteggelit: @linsword13 Newest commit should have the fix now.

tteggelit: @linsword13 OK. I had to reset my branch to get it clean, but I think it's now ready with your requested updates.

linsword13 (dismissed):

tteggelit:
@linsword13 It looks like the one failing test might be because the compiler was removed from the intel-mpi software definition. I believe all the listed applications have gcc14 compiler specified in their definitions. I can either add the compiler back or can also remove the compiler from these applications. I would vote for adding it back and then later removing it from all the applications when they can all be properly tested to ensure there's no unintended side effects from doing so. What do you think?

linsword13:
> @linsword13 It looks like the one failing test might be because the compiler was removed from the intel-mpi software definition. I believe all the listed applications have gcc14 compiler specified in their definitions. I can either add the compiler back or can also remove the compiler from these applications. I would vote for adding it back and then later removing it from all the applications when they can all be properly tested to ensure there's no unintended side effects from doing so. What do you think?

I see, makes sense. I agree that we can add that back in for the current PR, and then do the cleanup as a follow-up.

linsword13 (approved):

PR #1677, opened 2026-07-23
make edit type deducible

rfbgo:
Previously for ramble `ramble edit <name>` you had to specify type. This PR makes it so if you do `ramble edit modifier_name` it will attempt to find what you mean and deduce the type

### DIFF
--- a/lib/ramble/ramble/cmd/edit.py
+++ b/lib/ramble/ramble/cmd/edit.py
@@ -58,6 +58,97 @@ def edit_object(name, obj_type_name, repo_path, namespace):
         logger.die("No valid editor was found.")
 
 
+def normalize_type_name(type_name):
+    if not type_name:
+        return None
+
+    # Map aliases
+    aliases = {
+        "test": "test",
+        "tests": "test",
+        "command": "command",
+        "commands": "command",
+        "docs": "docs",
+        "doc": "docs",
+        "module": "module",
+        "modules": "module",
+    }
+    if type_name in aliases:
+        return aliases[type_name]
+
+    # Map singular to plural for ObjectTypes
+    norm_type = type_name.lower().replace("-", " ").replace("_", " ")
+    for obj_type in ramble.repository.ObjectTypes:
+        if norm_type == obj_type.name.lower().replace("_", " "):
+            return obj_type.name
+        if obj_type in ramble.repository.type_definitions:
+            singular = ramble.repository.type_definitions[obj_type]["singular"]
+            if norm_type == singular.lower().replace("_", " ").replace("-", " "):
+                return obj_type.name
+
+    return type_name
+
+
+def object_exists(name, obj_type_name, repo_path=None, namespace=None):
+    try:
+        obj_type = ramble.repository.ObjectTypes[obj_type_name]
+        if repo_path:
+            repo = ramble.repository.Repo(repo_path, object_type=obj_type)
+        elif namespace:
+            repo = ramble.repository.paths[obj_type].get_repo(namespace)
+        else:
+            repo = ramble.repository.paths[obj_type]
+        # Check if the filename for the object name exists
+        path = repo.filename_for_object_name(name)
+        return os.path.exists(path)
+    except Exception:
+        return False
+
+
+def non_object_exists(name, type_name):
+    type_to_path = {
+        "test": ramble.paths.test_path,
+        "command": ramble.paths.command_path,
+        "docs": os.path.join(ramble.paths.lib_path, "docs"),
+        "module": ramble.paths.module_path,
+    }
+    path = type_to_path[type_name]
+    # convert command names to python module name
+    if path == ramble.paths.command_path:
+        name = ramble.cmd.python_name(name)
+
+    path = os.path.join(path, name)
+    if os.path.exists(path):
+        return True
+
+    # Otherwise do a glob check (excluding backups/pyc)
+    files = glob.glob(path + ".*")
+    exclude_list = [".pyc", "~"]
+    files = list(filter(lambda x: all(s not in x for s in exclude_list), files))
+    return len(files) > 0
+
+
+def deduce_type(name, repo_path=None, namespace=None):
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/cmd/edit.py
```diff
@@ -58,51 +58,115 @@ def edit_object(name, obj_type_name, repo_path, namespace):
...
+
+    # Map singular to plural for ObjectTypes
+    for obj_type in ramble.repository.ObjectTypes:
+        # Check if type_name matches the enum member name (plural)
+        if type_name == obj_type.name:
+            return type_name
+        # Check if type_name matches the singular name defined in type_definitions
+        if obj_type in ramble.repository.type_definitions:
+            if type_name == ramble.repository.type_definitions[obj_type]["singular"]:
+                return obj_type.name
```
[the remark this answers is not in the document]
rfbgo: done

lib/ramble/ramble/cmd/edit.py
```diff
@@ -116,17 +180,41 @@ def edit(parser, args):
...
+    type_to_path = {
+        "test": ramble.paths.test_path,
+        "tests": ramble.paths.test_path,
+        "command": ramble.paths.command_path,
+        "commands": ramble.paths.command_path,
+        "docs": os.path.join(ramble.paths.lib_path, "docs"),
+        "doc": os.path.join(ramble.paths.lib_path, "docs"),
+        "module": ramble.paths.module_path,
+        "modules": ramble.paths.module_path,
+    }
```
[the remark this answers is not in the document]
rfbgo: done

lib/ramble/ramble/cmd/edit.py
```diff
@@ -58,51 +58,112 @@ def edit_object(name, obj_type_name, repo_path, namespace):
...
+        return True
+
+    # Otherwise do a glob check (excluding backups/pyc)
+    files = glob.glob(path + ".*")
+    exclude_list = [".pyc", "~"]
+    files = list(filter(lambda x: all(s not in x for s in exclude_list), files))
+    return len(files) > 0
+
+
+def deduce_type(name, repo_path=None, namespace=None):
```
douglasjacobsen: It feels like it would be nice if this warned if there were multiple matching objects rather than just returning the first one you hit.

And there should probably be a doc string here to describe the function's behavior.

lib/ramble/ramble/cmd/edit.py
```diff
@@ -58,51 +58,112 @@ def edit_object(name, obj_type_name, repo_path, namespace):
...
+        else:
+            repo = ramble.repository.paths[obj_type]
+        # Check if the filename for the object name exists
+        path = repo.filename_for_object_name(name)
+        return os.path.exists(path)
+    except Exception:
+        return False
+
+
+def non_object_exists(name, type_name):
```
douglasjacobsen: Same comment about warning if there are multiple matches. Maybe the warning should print the commands to edit each individually, but then also open whatever the default is based on the documented behvaior?

douglasjacobsen (changes requested):
Thanks, this is neat! A few comments to help the user experience but otherwise looks good.

lib/ramble/ramble/cmd/edit.py
rfbgo: Sure, sounds good. My thinking was that this was a strict improvement on the old behavior of opening apps (which is still the default in conflict) and the user would know pretty fast if they got the wrong object type. I can add a warning 

Agreed on doc string

lib/ramble/ramble/cmd/edit.py
rfbgo: Sounds good. I will try roll into one warning that explains all ambiguity in the match

lib/ramble/ramble/cmd/edit.py
rfbgo: I have an update that does this:

```
  $ ramble edit foo
    ==> Error: Multiple matches found for 'foo'. Please specify --type, --repo, or --namespace to disambiguate.
    Matches are:
      Type: applications, Repo: builtin, Path: ~/ramble/edit_auto/var/ramble/repos/builtin/applications/foo/application.py
      Type: test, Path: ~/ramble/edit_auto/lib/ramble/ramble/test/foo.py
```

douglasjacobsen: Two nits about this:

1) I was thinking it would be nice if this was more like a warning than an error.  And then it would still open the first match as the default.

2) For the warning text, I was thinking it would be nice if it was copy-pastable. So something like:

```
==> Warn: Multiple matches found for 'foo'. Defaulting to the first match to open. In order to open a different object, use the commands below:

  Builtin Application, Command: `ramble edit --type applications --repo builtin foo`
  Test: Command: `ramble edit --type test foo`
```

rfbgo: #1 is actually true now, it does something like:

```
==> Warning: Multiple matches found for 'config'. Opening highest-precedence match:
  ~/ramble/edit_auto/lib/ramble/ramble/test/config.py
Other matches found:
  Type: command, Path: ~/ramble/edit_auto/lib/ramble/ramble/cmd/config.py
  Type: module, Path: ~/ramble/edit_auto/lib/ramble/ramble/config.py
```

Happy to implement #2

rfbgo: Done, we now get something like:

```
 $ ramble edit config
==> Warning: Multiple matches found for 'config'. Opening highest-precedence match:
  /Users/user/ramble/edit_auto/lib/ramble/ramble/test/config.py
In order to open a different object, use the commands below:
  Type: command, Command: ramble edit --type command config
  Type: module, Command: ramble edit --type module config
```

douglasjacobsen: Awesome, thanks. Sorry about the warning bit. I thought from your comment with the example that it was a hard error.

douglasjacobsen (approved):

PR #1679, opened 2026-07-25
Tutorial and documentation fixes/updates

tteggelit:
Documentation and Tutorial fixes:

- Updates existing tutorials to fix problems when using current ramble, spack, and spack packages
- Adds a new `workspace manage experiments` tutorial
- Lots of documentation fixes especially regarding Return values and specifying they're Return Types
- Minor config change to allow sphinx `--fail-on-warning` to pass

There's still a lot of work that could be done to standardize formatting for documentation, but this PR addresses documentation that already exists in the expected sphinx format to be more compliant and explicit.

### DIFF
--- a/lib/ramble/docs/tutorials/Workspace_manage_experiments_command.rst
+++ b/lib/ramble/docs/tutorials/Workspace_manage_experiments_command.rst
@@ -42,7 +42,8 @@ To start, create a new workspace using ``ramble workspace create``:
 
    $ ramble workspace create -d manage-workspace
 
-This creates a workspace with the default ``ramble.yaml`` configuration:
+This creates an anonymous workspace (see note below) with the default ``ramble.yaml``
+configuration:
 
 .. code-block:: yaml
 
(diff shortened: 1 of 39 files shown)


linsword13 (dismissed):

linsword13:
Can you pull in the latest develop, just so the PR checks can run? Thanks!

tteggelit:
> Can you pull in the latest develop, just so the PR checks can run? Thanks!

Done!

lib/ramble/docs/tutorials/Workspace_manage_experiments_command.rst
```diff
@@ -0,0 +1,403 @@
...
+
+--------------------
+Creating a Workspace
+--------------------
+
+To start, create a new workspace using ``ramble workspace create``:
+
+.. code-block:: console
+
+   $ ramble workspace create -d manage-workspace
```
douglasjacobsen, line 43: Should probably discuss the `-d` flag here as well. Or at least refer to other documentation that discusses it and alternative paths for workspaces.

lib/ramble/docs/tutorials/Workspace_manage_experiments_command.rst
```diff
@@ -0,0 +1,403 @@
...
+        processes_per_node: 1
+    applications: {}
+    software: 
+        packages: {}
+        environments: {}
+
+And there are no experiments listed in the output from ``ramble -D manage-workspace workspace info``.
+
+.. note::
+   Throughout this tutorial, we will use the ``-D manage-workspace`` flag to direct
```
douglasjacobsen, line 85: Again, should probably call out that `-D ` expects a path to a workspace, not the name of a workspace. And that it can be provided as a relative or an absolute path.

lib/ramble/docs/tutorials/Workspace_manage_experiments_command.rst
```diff
@@ -0,0 +1,403 @@
...
+   commands to our new workspace without needing to activate it explicitly.
+
+-------------------------------
+Generating Baseline Experiments
```
douglasjacobsen: Maybe change `Baseline` here to `Default`. IMO, Baseline implies the thing you are comparing to, while default is more like "I don't want to change the parameters, just give me something that works."

I think in this section you mean more of the latter.

douglasjacobsen (changes requested):

lib/ramble/docs/tutorials/Workspace_manage_experiments_command.rst
tteggelit: See if 202b6e37f5c429f5a265e6a05451273a54e820d5 addresses this and the following comments.

douglasjacobsen (approved):

PR #1680, opened 2026-07-27
remove unused hpl vars

rfbgo:
I can't find where these are used. Can anyone confirm if they are needed and what I overlooked?


linsword13 (approved):
I think you are right, these are indeed unused.

PR #1697, opened 2026-07-31
Update pipeline documentation

tteggelit:
This:
- Adds a new Basics section to introduce a typical Ramble workspace workflow and the concepts of pipelines and phases
- Updates the Advanced section on pipelines with more information and diagrams to help explain the flow
- Updates the docs `conf.py` for graphviz to remove global rank direction and sets the fonts to something less Mac specific

Attaching PDFs of the new pages to make it a bit easier to see the results without people needing to checkout my branch and run Sphinx against it.
[Basics - Pipelines and Phases.pdf](https://github.com/user-attachments/files/30598866/Basics.-.Pipelines.and.Phases.pdf)
[Advanced - Pipelines and Builtins.pdf](https://github.com/user-attachments/files/30598868/Advanced.-.Pipelines.and.Builtins.pdf)

### DIFF
--- a/lib/ramble/docs/dev_guides/advanced_topics.rst
+++ b/lib/ramble/docs/dev_guides/advanced_topics.rst
@@ -51,7 +51,7 @@ command and targets a specific stage of the experiment lifecycle. The most
      → :py:class:`~ramble.pipeline.ArchivePipeline`
 
 There are more pipelines that Ramble can use to perform complex actions
-on a workspace, which can be seen below in :ref:`Additional Pipelines
+on a workspace, which can be seen in the :ref:`Additional Pipelines
 <additional-pipelines>` section below.
 
 Phase Graphs
--- a/lib/ramble/docs/dev_guides/advanced_topics.rst
+++ b/lib/ramble/docs/dev_guides/advanced_topics.rst
@@ -28,13 +28,34 @@ types supported in Ramble.
 Experiment Pipelines and Phases
 -------------------------------
 
-Ramble has a concept of ``pipeline``, which represent full actions that can
-be taken on a workspace. Some of the common ``pipelines`` that are used are the
-``setup`` (driven with ``ramble workspace setup``) and ``analyze`` (driven with
-``ramble workspace analyze``) pipelines.
+Architecture Overview
+=====================
 
-There are several more pipelines that Ramble uses to perform complex actions
-on a workspace, which can be seen in :mod:`ramble.pipeline`.
+Ramble organizes workspaces into a modular **Pipeline** and **Phase Graph** architecture.
+
+Pipelines
+---------
+
+Ramble has a concept of a ``pipeline`` which represents full actions that can
+be taken on a workspace. Each ``pipeline`` has a corresponding workspace
+command and targets a specific stage of the experiment lifecycle. The most 
+:ref:`common pipelines <standard-pipelines>` are:
+
+   * :ref:`ramble workspace setup <ramble-workspace-setup>`
+     → :py:class:`~ramble.pipeline.SetupPipeline`
+   * :ref:`ramble on <ramble-on>`
+     → :py:class:`~ramble.pipeline.ExecutePipeline`
+   * :ref:`ramble workspace analyze <ramble-workspace-analyze>`
+     → :py:class:`~ramble.pipeline.AnalyzePipeline`
+   * :ref:`ramble workspace archive <ramble-workspace-archive>`
+     → :py:class:`~ramble.pipeline.ArchivePipeline`
+
+There are more pipelines that Ramble can use to perform complex actions
+on a workspace, which can be seen below in :ref:`Additional Pipelines
+<additional-pipelines>` section below.
+
+Phase Graphs
+------------
 
 Pipelines are built out of phases. In Ramble, a ``phase`` represents a specific
 step along the path of completing the action defined by the ``pipeline``.
@@ -42,9 +63,35 @@ Examples of phases include ``get_inputs`` (for downloading input files needed
 by a workload) and ``software_install`` (for performing software installation
 using a package manager).
 
-Phases can be defined in a variety of locations. Some base classes define phases for specific pipelines. Additionally,
-instances of application, modifier, or package manager definitions can define
-phases as well. Each phase is defined in two parts. The first part is to define
+For each pipeline, Ramble builds a directed acyclic graph
+(:py:class:`ramble.graphs.PhaseGraph`) containing all phases registered for that pipeline.
+
+   * **Constraint Declaration**: Phase dependency constraints are declared within object class
+     definitions using ``run_before=['phase_name']`` or ``run_after=['phase_name']``.
+   * **Order Resolution**: When a pipeline initializes, :py:class:`ramble.graphs.PhaseGraph`
+     resolves phase dependencies using a topological sort. If cyclical dependencies exist,
+     Ramble raises a :py:class:`~ramble.error.RambleError`.
+
+Phase Registration
+==================
+
+Phases can be defined in a variety of locations. Some base classes define phases for specific
+pipelines. Additionally, Applications, Modifiers, and Package Managers can all define and
+register their own phases to build more complex pipelines for specific usecases.
+
+   * :doc:`Applications <../application_list>` define core command generation, input downloading,
+     license inclusion, and analysis hooks.
+   * :doc:`Package Managers <../package_managers>` (e.g., Spack, Pip) inject software environment
+     creation and installation phases.
+   * :doc:`Modifiers <../modifier_list>` (e.g., containers, profilers) inject pre-exec and post-exec
+     hooks.
+   * :doc:`Workflow Managers <../workflow_managers>` (e.g., Slurm, PBS, Google Batch) define batch
+     submission logic and job execution wrappers.
+   * :doc:`Systems <../systems>` and :doc:`Platforms <../platforms>` contribute hardware-specific
+     variables, CPU architecture parameters, and compiler configurations expanded during pipeline
+     setup.
+
+Each phase is defined in two parts. The first part is to define
 a class method on a object definition. Phase names need to begin with an
 underscore, and they have the following signature:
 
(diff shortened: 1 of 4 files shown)


rfbgo (commented):
@dcoapp recheck

tteggelit:
The Docker image won't get rebuilt based on the new Dockerfiles until the PR is merged, so the doc build test is still using the old image that doesn't have graphviz included. Verified that adding graphviz does allow the doc build test to complete. I can split the Dockerfile updates into their own PR to force an image rebuild and then have this one re-run the tests after that if that is more desirable.

tteggelit:
I'm assuming since #1787 has been merged that the images should now have graphviz. Can I get another gcbrun done on this to make sure it now passes all actions?

lib/ramble/docs/dev_guides/advanced_topics.rst
```diff
@@ -28,23 +28,70 @@ types supported in Ramble.
...
+     definitions using ``run_before=['phase_name']`` or ``run_after=['phase_name']``.
+   * **Order Resolution**: When a pipeline initializes, :py:class:`ramble.graphs.PhaseGraph`
+     resolves phase dependencies using a topological sort. If cyclical dependencies exist,
+     Ramble raises a :py:class:`~ramble.error.RambleError`.
+
+Phase Registration
+==================
+
+Phases can be defined in a variety of locations. Some base classes define phases for specific
+pipelines. Additionally, Applications, Modifiers, and Package Managers can all define and
```
douglasjacobsen: Probably say "Object types (applications, modifiers, etc...)" instead of listing only some of the objects.

lib/ramble/docs/dev_guides/advanced_topics.rst
```diff
@@ -28,23 +28,70 @@ types supported in Ramble.
...
+register their own phases to build more complex pipelines for specific use cases.
+
+   * :doc:`Applications <../application_list>` define core command generation, input downloading,
```
douglasjacobsen: This section seems out of place. It seems more like this is describing general object types, and their purposes, rather than phases that exist.

lib/ramble/docs/dev_guides/advanced_topics.rst
```diff
@@ -70,8 +117,400 @@ phase. Phase registration can also define a ``run_after`` list of phases to exec
...
+            p1 -> p2 -> p3 -> p4 -> p5 [color="#28a745", penwidth=2.0, weight=10];
+        }
+
+        // Horizontal rank alignment between Classes and Phases
+        { rank=same; mod; p1; }
+        { rank=same; pm; p2; }
+        { rank=same; app; p3; }
+
+        // Classes --> Phases
+        mod -> p1 [
```
douglasjacobsen: It's unclear to me from this diagram if this is just intended to be an example, or if this is intended to be authoritative about the ordering.

Regardless, the mapping is incorrect (utilities are not injected from modifiers). I think that a diagram of this could be super helpful, but I also think it'd be good if we made it more exhaustive and covered a simple case completely.

douglasjacobsen: Also, this makes me wonder if we should have some command within a workspace to create both the DAG and the flattened order of all of the phases...

lib/ramble/docs/dev_guides/advanced_topics.rst
```diff
@@ -70,8 +117,400 @@ phase. Phase registration can also define a ``run_after`` list of phases to exec
...
+        ];
+        app:s -> p5:w [
+            label=<<i>register_phase('make_experiments')</i>>,
+            color="#0056b3"
+        ];
+    }
+
+.. _standard-pipelines:
+
+Standard Pipelines
```
douglasjacobsen: I don't think we should have the diagrams in here, mostly because they are not exhaustive and are not 100% guaranteed to represent the phase ordering in a random experiment.

Any object can register phases into any of these pipelines. So these can change quickly.

However, having the pipelines defined, how they map to commands, and what their purposes are is really helpful.

lib/ramble/docs/dev_guides/advanced_topics.rst
```diff
@@ -70,8 +117,400 @@ phase. Phase registration can also define a ``run_after`` list of phases to exec
...
+        ];
+        edge [fontsize=10, color="#862e9c", penwidth=1.3];
+
+        ar1 [label="archive_experiments\n(Registered Phase: collect logs, templates & FOM files)"];
+        ar2 [label="create_tarball / upload_archive\n(Generate archive.latest.tar.gz & upload)"];
+
+        ar1 -> ar2;
+    }
+
+Standard Pipeline Phases
```
douglasjacobsen: I also probably would remove this list. It's going to be difficult to maintain this, and it would probably be better to just give developers the command to see what the phases currently are on the CLI.

Something like:
```
ramble info --type base_classes application-base --attrs registered_phases -v
```

But that also doesn't show the ordering.

lib/ramble/docs/dev_guides/advanced_topics.rst
```diff
@@ -70,8 +117,400 @@ phase. Phase registration can also define a ``run_after`` list of phases to exec
...
+     - Computes summary statistics (mean, stddev, min, max) for repeated experiment runs.
+   * - ``archive_experiments``
+     - Archive
+     - Application Base
+     - Copies logs, rendered templates, FOM output files, and inventory metadata into the workspace
+       archive directory.
+
+.. _additional-pipelines:
+
+Additional Pipelines
```
douglasjacobsen: I think you could just merge this list with the list above (with the diagrams removed) to make it comprehensive.

lib/ramble/docs/pipelines.rst
```diff
@@ -0,0 +1,104 @@
...
+        ];
+        edge [fontsize=10, color="#333333", penwidth=1.5];
+
+        step1 [
+            label=<<b>1. Workspace Creation</b><br/>(ramble workspace create)>,
+            fillcolor="#f8f9fa",
+            color="#495057"
+        ];
+        step2 [
+            label=<<b>2. YAML &amp; Template Configuration</b><br/>(ramble workspace edit)>,
```
douglasjacobsen: <mark for cycle start>

lib/ramble/docs/pipelines.rst
```diff
@@ -0,0 +1,104 @@
...
+            fillcolor="#e9ecef",
+            color="#495057"
+        ];
+        step3 [
+            label=<<b>3. Workspace Concretization</b><br/>(ramble workspace concretize)>,
```
douglasjacobsen: This should be noted as optional (or maybe one time at best?). It might be better to name this node as something like `Software configuration` and list that it can possibly be done with concretization.

lib/ramble/docs/pipelines.rst
```diff
@@ -0,0 +1,104 @@
...
+            fillcolor="#d4edda",
+            color="#28a745"
+        ];
+        step5 [
+            label=<<b>5. Execute Pipeline</b><br/>(ramble on)>,
+            fillcolor="#cce5ff",
+            color="#004085"
+        ];
+        step6 [
+            label=<<b>6. Analyze Pipeline</b><br/>(ramble workspace analyze)>,
```
douglasjacobsen: We should add a cycle here, because this can be a loop. The loop probably needs to call out something like "Decide if there are further explorations needed" and then continue to re-configure the workspace.

lib/ramble/docs/pipelines.rst
```diff
@@ -0,0 +1,104 @@
...
+        step8 [
+            label=<<b>8. Archive Pipeline</b><br/>(ramble workspace archive)>,
+            fillcolor="#eebefa",
+            color="#862e9c"
+        ];
+
+        step1 -> step2 -> step3 -> step4 -> step5 -> step6 -> step7 -> step8;
+    }
+
+1. **Creation**: Initialize a workspace using :ref:`ramble workspace create
```
douglasjacobsen: Same comments here, but in words instead of the diagram.

douglasjacobsen (changes requested):
Thanks for working on this (and sorry it took so long to review). It's great to make the documentation better!

lib/ramble/docs/dev_guides/advanced_topics.rst
tteggelit: Each of these are examples of phases that these objects register. Perhaps explicitly stating that these are examples and mentioning the actual phase name and pipeline would make it better?

lib/ramble/docs/dev_guides/advanced_topics.rst
tteggelit: It says the diagram illustrates an example of how classes register specific phases. Perhaps it should say "... illustrates a ***hypothetical*** `ramble.pipeline.SetupPipeline` example of how
classes ***can*** register ..." to make it a bit more clearly?

Also, maybe replace `Profilers` and `bootstrap_utilities` with `Docker` and `pull_container`? The reason being this is to show that essentially any object can register a phase with any pipepline (but perhaps that's incorrect?).

lib/ramble/docs/dev_guides/advanced_topics.rst
tteggelit: I think it's helpful to illustrate how the phases are chained together within a pipeline. Perhaps caveating that this just an illustrative example of a typical (or some other phrase) experiment but that any given experiment can and will have different phases depending on how it is configured.

lib/ramble/docs/dev_guides/advanced_topics.rst
tteggelit: The list wasn't intended to be a complete list, just to highlight what phases are common for "typical" experiments. I tried to select ones that seemed to be common with lots of common experiment setups and part of the core code. They were listed to help give developers some sign posts of how to further explore the code regarding this.

lib/ramble/docs/pipelines.rst
tteggelit: Was this an errant comment?

lib/ramble/docs/pipelines.rst
tteggelit: How about something like:

Image

lib/ramble/docs/dev_guides/advanced_topics.rst
tteggelit: See if f505bd62f addresses your comment.

lib/ramble/docs/pipelines.rst
tteggelit: See if f505bd62f addresses your comment.

lib/ramble/docs/pipelines.rst
tteggelit: See if f505bd62f addresses your comment.

tteggelit:
> Thanks for working on this (and sorry it took so long to review). It's great to make the documentation better!

Just pushed some updates and here's re-generated PDFs to make it easier to view:
[advanced_topics.pdf](https://github.com/user-attachments/files/32616522/advanced_topics.pdf)
[pipelines.pdf](https://github.com/user-attachments/files/32616524/pipelines.pdf)

PR #1709, opened 2026-08-07
Allow namespaced specs

dapomeroy:
Allows namespaced specs (e.g., builtin.app.wrf) to be used in place of object names and arguments indicating repo and type.

### DIFF
--- a/lib/ramble/docs/dev_guides/shared/repository_create.rst
+++ b/lib/ramble/docs/dev_guides/shared/repository_create.rst
@@ -104,7 +104,8 @@ Common object type abbreviations include:
 * ``sys`` or ``system`` for systems
 * ``plat`` or ``platform`` for platforms
 
+Base objects can be referenced with the ``base_<type_abbrev>`` prefix.
+
 These namespaced specs can be used in CLI commands (e.g., ``ramble info <spec>``,
 ``ramble edit <spec>``, ``ramble create <spec>``) and in workspace configuration files
 (``ramble.yaml``) under the ``applications:`` and ``environments:`` sections.
-
--- a/lib/ramble/docs/dev_guides/shared/repository_create.rst
+++ b/lib/ramble/docs/dev_guides/shared/repository_create.rst
@@ -82,3 +82,29 @@ when multiple exist with the same name. Each repository has a namespace, and
 these namespaces can be used to refer to specific instances of each object
 definition.
 
+Referencing Objects with Namespaces
+-----------------------------------
+
+When multiple repositories contain objects with the same name, or when you want
+to be explicit about which repository an object comes from, you can use fully-qualified
+namespaced specs.
+
+Namespaced specs take the form:
+
+* ``<namespace>.<object_name>`` (e.g., ``tutorial-repo.hostname``, ``builtin.wrf``)
+* ``<namespace>.<type_abbrev>.<object_name>`` (e.g., ``tutorial-repo.app.hostname``, ``builtin.mod.my_modifier``)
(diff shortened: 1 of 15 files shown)


lib/ramble/ramble/spec.py
```diff
@@ -16,31 +17,71 @@
...
+    # Strip any version suffix if present (e.g., app@1.0)
+    spec_like = spec_like.partition("@")[0]
+
+    parts = spec_like.split(".")
+
+    import ramble.repository
+
+    type_map = ramble.repository.get_object_type_map()
+
+    if len(parts) >= 3 and parts[-2] in type_map:
```
rfbgo, line 34: nit: do we need to do anything here to make sure the `parts` is case insensitive (assuming that is desirable)? It feels like we may as well support `App.hostname` if it's a simple as a `to_lower` here?

lib/ramble/ramble/cmd/workspace.py
```diff
@@ -902,19 +903,33 @@ def workspace_info(args):
...
-            experiment_template_name,
-        )
-        experiment_index_map[key].append(exp_name)
+        app_ns_raw = app_inst.variables[app_inst.keywords.application_namespace]
+        if app_ns_raw not in app_names_cache:
+            app_ns_spec = ramble.spec.Spec(app_ns_raw)
+            app_ns = app_ns_spec.fullname
+            app_name = app_inst.variables[app_inst.keywords.application_name]
+            app_ns_no_type = (
+                f"{app_ns_spec.namespace}.{app_ns_spec.name}"
```
rfbgo: I haven't had chance to play with this, but I was surprised this will get a different result depending on his `namespace` is passed.

Am I right in thinking when there is no namespace this is the same as app_name? Would it be more robust if we always injected the namespace so a lookup in `app_names_cache` would always be consistent / hold all 3 permutations regardless of what is initially passed?

rfbgo (commented):
Two random things caught my eye. 

The edit portion specifically LGTM, thanks for incorporating my type deduction changes.

It might be good to have @douglasjacobsen review too

lib/ramble/ramble/spec.py
dapomeroy: Good suggestion. I can't think of any reason we'd want it to be case sensitive. Updated.

lib/ramble/ramble/cmd/workspace.py
dapomeroy: That's right. Updated this to inject the namespace.

lib/ramble/docs/dev_guides/shared/repository_create.rst
```diff
@@ -82,3 +82,29 @@ when multiple exist with the same name. Each repository has a namespace, and
...
+to be explicit about which repository an object comes from, you can use fully-qualified
+namespaced specs.
+
+Namespaced specs take the form:
+
+* ``<namespace>.<object_name>`` (e.g., ``tutorial-repo.hostname``, ``builtin.wrf``)
+* ``<namespace>.<type_abbrev>.<object_name>`` (e.g., ``tutorial-repo.app.hostname``, ``builtin.mod.my_modifier``)
+* ``<namespace>.<type_abbrev>.<object_name>@<version>`` (e.g., ``builtin.app.wrf@4.2``, ``builtin.app.wrf@{version}``)
+
+Common object type abbreviations include:
```
douglasjacobsen, line 98: It might be good to say that `base` can be prefixed on any of these as well.

douglasjacobsen (changes requested):

lib/ramble/docs/dev_guides/shared/repository_create.rst
dapomeroy: Added this

douglasjacobsen (approved):

PR #1711, opened 2026-08-10
Agents updates

tteggelit:
This update `AGENTS.md` to make it smaller (to reduce context consumption) and move aspects to relevant skills as opposed to being in the agent context. It updates the existing `ramble-workspace-wizard` to address some changes to `workflow_manager` (`config:` vs. `variants:`) and adds several new skills that can be used in more agentic workflows as needed.


rfbgo (commented):
@dcoapp recheck

PR #1719, opened 2026-08-12
Ensure model name does not override machine type

linsword13:


lib/ramble/ramble/test/uploader.py
```diff
@@ -389,3 +389,56 @@ def test_fom_validation_fails_with_none():
...
+    contexts4 = [
+        {"foms": [{"name": "Model name", "value": "Intel(R) Xeon(R) Gold 6268CL CPU @ 2.80GHz"}]}
+    ]
+    determine_node_type(exp4, contexts4)
+    assert exp4.node_type == "Intel(R) Xeon(R) Gold 6268CL CPU @ 2.80GHz"
+
+    exp5 = types.SimpleNamespace(node_type=default_node_type_val)
+    contexts5 = [{"foms": [{"name": "other_fom", "value": "123"}]}]
+    determine_node_type(exp5, contexts5)
+    assert exp5.node_type == default_node_type_val
```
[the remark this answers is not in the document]
linsword13: Good point, updated.

lib/ramble/ramble/uploader.py
```diff
@@ -237,18 +237,16 @@ def determine_node_type(experiment, contexts):
...
-        for fom in context["foms"]:
+        for fom in context.get("foms", []):
             if "machine-type" in fom["name"]:
                 experiment.node_type = fom["value"]
-                continue
-            elif "Model name" in fom["name"]:
-                experiment.node_type = fom["value"]
-                continue
+                return
+            elif "Model name" in fom["name"] and node_type == default_node_type_val:
```
tteggelit, line 246: Does `machine-type` and `Model name` only come from modifiers like `lscpu` and `gcp-metadata`? If so, is there anything special or protected about these names? Does that mean if those modifiers are changed for some reason this code would break?

linsword13: Yeah this is currently only a best-effort for getting the machine type.

rfbgo (approved):

PR #1724, opened 2026-08-13
abstract key normlization and support type plurals

rfbgo:
This allows for things like `ramble edit openfoam --type base_application` (normally `--type base_applications` but it's not really plural in that context..)

### DIFF
--- a/lib/ramble/ramble/repository.py
+++ b/lib/ramble/ramble/repository.py
@@ -101,6 +101,7 @@ type_definitions = {
         "config_section": "package_manager_repos",
         "accepted_configs": ["package_manager_repo.yaml", unified_config],
         "singular": "package manager",
+        "aliases": ["pkg", "package"],
     },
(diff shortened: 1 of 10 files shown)


lib/ramble/ramble/repository.py
```diff
@@ -205,8 +205,57 @@
 }
 
 
+def _normalize_type_key(key):
+    return str(key).lower().replace("-", "_").replace(" ", "_")
+
+
+_TYPE_ALIASES = {
```
linsword13: More like a nit: I wonder if it'd be better to have this inside the `type_definitions`, just so we have all these object-related information defined in a single place?

rfbgo: Yea that's a good idea

rfbgo: implemented. Thanks for suggestion, I do think this way is nicer now :)

douglasjacobsen (approved):

PR #1725, opened 2026-08-14
Add `allow_external` arg when requiring utilities

douglasjacobsen:
This merge adds the `allow_external` arugment on the `requires_utility` directive. This can be used to allow (or disallow) experiments from using an external utilty. For example, this can be used to force experiments to always bootstrap an internal spack instead of allowing a user installed spack to be used.

Additionally, this merge fixes an issue when running some ramble commands without having spack in your path.

### DIFF
--- a/var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
+++ b/var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
@@ -799,6 +799,7 @@ class SpackRunner(CommandRunner):
                 os.path.dirname(self.spack.exe[0])
             )
         except RunnerError:
+            self.shell = shell
             self.spack = None
             self.installer = None
             self.concretizer = None
@@ -810,6 +811,8 @@ class SpackRunner(CommandRunner):
             script = "setup-env.csh"
         elif self.shell == "fish":
             script = "setup-env.fish"
+        else:
+            script = "setup-env.sh"
         self.source_script = os.path.join(
             self.spack_dir, "share", "spack", script
         )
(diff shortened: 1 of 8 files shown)


var/ramble/repos/builtin/package_managers/spack-lightweight/package_manager.py
```diff
@@ -764,24 +765,44 @@ def __init__(self, shell="bash", dry_run=False, env=None):
...
+                    self.spack.add_default_arg(arg)
+
+            self.spack_dir = os.path.dirname(
+                os.path.dirname(self.spack.exe[0])
+            )
+        except RunnerError:
+            self.spack = None
+            self.installer = None
+            self.concretizer = None
+            self.spack_dir = os.path.join("missing", "path")
```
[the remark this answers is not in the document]
douglasjacobsen: I don't think we want this actually...

I basically want the spack runner to think it's active, and let the failure happen later in the execution (when trying to actually do spack things).

var/ramble/repos/builtin/utilities/spack/utility.py
```diff
@@ -35,7 +35,7 @@ class Spack(UtilityBase):
     )
     variable(
         "spack_version",
-        default="v0.22.0",
+        default="v1.2.2",
```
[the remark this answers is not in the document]
douglasjacobsen: This isn't true...

linsword13, line 38: Are we concerned about the "`--fresh` install issue" that you recently discovered on this version of Spack?

linsword13 (dismissed):

var/ramble/repos/builtin/utilities/spack/utility.py
douglasjacobsen: No, we have the `--first` in already that fixes the issue. In the future, we can look at removing it (there's a PR into spack to fix the issue)

linsword13: Ah makes sense

linsword13 (approved):

PR #1727, opened 2026-08-17
Convert FOM extract to strategy pattern [AI]

rfbgo:
This is some nice OOP around fom extraction, that allows for different methods of extraction

In some situations (such as no contexts) this allows us to have an optimized extract that swap out  the full scan "forward last wins" and instead read the file backwards and stop early

```
class BackwardsAnalysisStrategy(AnalysisStrategyBase):
    """Optimized analysis strategy that reads logs backwards and stops early."""
```


rfbgo (commented):
@dcoapp recheck

douglasjacobsen (approved):

PR #1729, opened 2026-08-18
allow parents to be queried by  `--where "'{experiment_status}' ==`

rfbgo:
hoist parent status calculation to allow parents to be queried by  `--where "'{experiment_status}' == FAILED`

### DIFF
--- a/lib/ramble/ramble/test/success_criteria_functionality/repeat_success_strict.py
+++ b/lib/ramble/ramble/test/success_criteria_functionality/repeat_success_strict.py
@@ -11,6 +11,7 @@ import os
 import pytest
 
 from ramble.main import RambleCommand
+from ramble.util import json_util
 from ramble.util.foms import SummaryFoms
 
 # everything here uses the mock_workspace_path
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/test/success_criteria_functionality/repeat_success_strict.py
```diff
@@ -85,3 +86,164 @@ def test_repeat_success_strict(mock_applications, make_workspace_from_config):
         assert "SUCCESS" not in data
         assert f"summary::{SummaryFoms.N_TOTAL.value}" not in data
         assert f"summary::{SummaryFoms.N_SUCCESS.value}" not in data
+
+
+def test_repeat_analyze_where_experiment_status(mock_applications, make_workspace_from_config):
+    # Test strict mode (default: repeat_success_strict is True)
+    test_config = """
+ramble:
```
douglasjacobsen: Can this be converted to use `workspace manage experiments`? It looks like the only thing you want to set that isn't easily set-able in that is `n_repeats`, but I think you could do something like:

```
with ramble.config.override("config:n_repeats:2"):
  ...
```

lib/ramble/ramble/test/success_criteria_functionality/repeat_success_strict.py
```diff
@@ -85,3 +86,164 @@ def test_repeat_success_strict(mock_applications, make_workspace_from_config):
...
+        assert "basic.working_wl.test_exp.1" not in exp_names
+        assert "basic.working_wl.test_exp.2" in exp_names
+
+
+def test_repeat_analyze_where_experiment_status_loose(
+    mock_applications, make_workspace_from_config
+):
+    # Test loose mode (repeat_success_strict is False)
+    test_config = """
+ramble:
```
douglasjacobsen: Also can this be converted to use `workspace manage experiments`?

douglasjacobsen (changes requested):
Thanks, this is awesome. I added two small comments, but otherwise LGTM.

lib/ramble/ramble/test/success_criteria_functionality/repeat_success_strict.py
rfbgo: nice idea, ill look into it

rfbgo: done :)

lib/ramble/ramble/test/success_criteria_functionality/repeat_success_strict.py
rfbgo: done :)

douglasjacobsen (approved):

PR #1741, opened 2026-08-24
Embedded scripts cleanup

rfbgo:
Previously we had python files (eg `_ramble_cleaner.py`) hiding as comments which we then echo to file. This PR promotes them to real files and allows them to be copied instead

This means we can do things like lint and reason about them more easily


linsword13 (dismissed):

linsword13 (commented):
Great stuff!

One nit question: are there other checks (like `mypy`) that should be enabled for these `/share/ramble/scripts/**` files?

rfbgo (commented):
@dcoapp recheck

douglasjacobsen (approved):

PR #1747, opened 2026-08-26
Add explicit version bootstrapping

douglasjacobsen:
This merge adds the ability for `requires_utility` to define an explict version that should be used instead of just a range. Additionally, now any variables for the utility can be passed into the `requires_utility` directive as a kwarg to make it more easily set.

For example, the spack utility can be required with something like:

```
requires_utility("spack", path="/spack-path", hash="1234abc")
```

which will control the hash and the installation path for the installed spack.


var/ramble/repos/builtin/base_classes/application-base/base_class.py
```diff
@@ -2620,10 +2620,33 @@ def _bootstrap_utilities(self, workspace, app_inst=None):
 
                                     fetch_kwargs[var_name] = var_info.default
 
-                        if ext_dep_name in ws_ext_deps:
-                            fetch_kwargs.update(ws_ext_deps[ext_dep_name])
+                        conf_to_merge = (
```
rfbgo, line 2623: This used to be a merge now it's an overwrite, does it mean we might be losing some object data?

lib/ramble/ramble/test/end_to_end/package_manager_provenance.py
```diff
@@ -59,8 +59,16 @@ def test_spack_package_manager_provenance_zlib(mock_applications, workspace_name
 
         workspace("setup", global_args=global_args)
 
-        spack_yaml = os.path.join(ws.software_dir, pm, "zlib", "spack.yaml")
-        spack_lock = os.path.join(ws.software_dir, pm, "zlib", "spack.lock")
+        spack_yaml = os.path.join(
+            __import__("glob").glob(os.path.join(ws.software_dir, f"{pm}*"))[0],
```
rfbgo, line 63: There is a bunch of these, should they be using a top level import?

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
```diff
@@ -99,18 +99,200 @@ def runner(self):
...
+                            util_name,
+                            ramble.repository.ObjectTypes.utilities,
+                        )
+
+                        conf_to_merge = merged_configs.get(
+                            util_name, ext_dep_conf
+                        )
+
+                        version_str = (
+                            conf_to_merge.get("commit")
```
rfbgo, line 190: Is there some logic somewhere to guard what subset of these the users can set? At a glance it looks like they can accidentally ask for a conflicting commit and branch and we silently take them in the order defined here?

var/ramble/repos/builtin/base_classes/application-base/base_class.py
douglasjacobsen: Well, the previous logic was actually the incorrect semantic. I think it should actually be an overwrite rather than a merge to make sure that objects can control what they get out of the utility they require (that's ultimately what this PR is supposed to fix).

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
douglasjacobsen: That's a good point. Let me add an error if more than one is passed in here.

var/ramble/repos/builtin/base_classes/application-base/base_class.py
rfbgo: OK, makes sense, thanks!

lib/ramble/ramble/test/end_to_end/package_manager_provenance.py
douglasjacobsen: Yes, definitely. I fixed these. Good catch. :)

var/ramble/repos/builtin/base_classes/package-manager-base/base_class.py
rfbgo: Thanks!

lib/ramble/ramble/test/end_to_end/package_manager_provenance.py
rfbgo: Thanks!

rfbgo (dismissed):
Thanks for the changes!

PR #1748, opened 2026-08-26
add type cast protection to wm type casting

rfbgo:
Doug hit some issue where he got a bad int cast, this is just to tighten up some safety around that issues

```
    end_time = int(f.read().strip("\n"))
ValueError: invalid literal for int() with base 10: ''
```

### DIFF
--- a/var/ramble/repos/builtin/workflow_managers/slurm/test/slurm.py
+++ b/var/ramble/repos/builtin/workflow_managers/slurm/test/slurm.py
@@ -359,4 +359,5 @@ ramble:
         f.write(job_id_content)
 
     # Analyze should complete without crashing due to empty or invalid files
-    workspace("analyze", global_args=["-D", ws.root])
+    out = workspace("analyze", "-p", global_args=["-D", ws.root])
+    assert "Status = FAILED" in out
--- a/var/ramble/repos/builtin/workflow_managers/slurm/test/slurm.py
+++ b/var/ramble/repos/builtin/workflow_managers/slurm/test/slurm.py
@@ -314,3 +314,49 @@ ramble:
             filename,
         )
         assert os.path.isfile(archived_path)
+
+
+@pytest.mark.parametrize(
+    "end_time_content,job_id_content",
+    [
+        ("", ""),
+        ("   \n", "12345"),
+        ("not_a_float", "12345"),
+        ("1700000000", ""),
+    ],
+)
+def test_slurm_analyze_empty_or_invalid_files(
+    make_workspace_from_config, end_time_content, job_id_content
+):
+    test_config = """
+ramble:
+  variants:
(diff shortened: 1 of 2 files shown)


var/ramble/repos/builtin/workflow_managers/slurm/workflow_manager.py
```diff
@@ -222,26 +219,28 @@ def _get_job_termination_overhead(self):
...
                 "-j", job_id, "-X", "-n", "--format=End", output=str
             ).strip()
             # sacct by default reports timestamp in local timezone
-            # TODO: can use more convenient method with python 3.7+
             job_end_time = datetime.datetime.strptime(
                 job_end_time_str, "%Y-%m-%dT%H:%M:%S"
             ).timestamp()
             duration = int(job_end_time - script_end_time)
-        except (ProcessError, ValueError) as e:
+        except (ProcessError, ValueError, OSError) as e:
```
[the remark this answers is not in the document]
rfbgo: done

var/ramble/repos/builtin/workflow_managers/slurm/test/slurm.py
```diff
@@ -314,3 +314,49 @@ def test_slurm_archive_patterns(make_workspace_from_config):
...
+    end_time_path = os.path.join(experiment_dir, ".slurm_script_end_time")
+    with open(end_time_path, "w", encoding="utf-8") as f:
+        f.write(end_time_content)
+
+    job_id_path = os.path.join(experiment_dir, ".slurm_job")
+    with open(job_id_path, "w", encoding="utf-8") as f:
+        f.write(job_id_content)
+
+    # Analyze should complete without crashing due to empty or invalid files
+    workspace("analyze", global_args=["-D", ws.root])
```
linsword13: nit: it feels a bit weird that the test has no assertions. Maybe some basic checks on the analyze output?

rfbgo: Yea as written the test is mostly "dont crash if empty", I was OK with since that is mostly my intnet here (over say, checking for a file), but I agree we can add something like:

```
  assert os.path.exists(os.path.join(ws.root, "results.latest.txt"))
```

I can do that now for completeness

linsword13: Or can we assert on the status?

rfbgo: pushed :) thanks for the feedback!

rfbgo: Ah sure, good idea. I was thinking that would require a file read but I can probably pass `-p`

linsword13 (approved):

PR #1753, opened 2026-09-01
Replace live gcs fetch test with mocked

rfbgo:
Previously we were hitting a live end point out of our control (`gs://hpc-toolkit-demos/build_cache/index.json`), which means our unit tests required internet.  This addresses those 2 issues


rfbgo (commented):
@dcoapp recheck

linsword13 (approved):

PR #1756, opened 2026-09-01
Implement lazy directives for Ramble

linsword13:
This is inspired by Spack's lazy directive (https://github.com/spack/spack/pull/51881).

* Add in the dict descriptor (as a non-data descriptor) to support the lazy initiailization
* Remove the `convert_class_attributes` since now the lazy execution allows us to perform copy on first access
* Update the various directives to support the new semantics. Specifically, it now requires listing out all mutated dicts of a directive
* Also unify the various language meta directives, so now they are all essentially DirectiveMeta
* Update the mirror test since before it implicitly relied on the mutation happening both at the class and the instance level


lib/ramble/ramble/language/language_base.py
```diff
@@ -315,49 +349,37 @@ def _warn_impossible(obj):
...
-                    else:
-                        # Remove directives args from the exec queue
-                        remove = next((d for d in directives if d is arg), None)
-                        if remove is not None:
-                            directives.remove(remove)
+                    elif callable(arg):
+                        DirectiveMeta._directives_to_be_executed = [
+                            (n, fn)
+                            for n, fn in DirectiveMeta._directives_to_be_executed
+                            if fn != arg
```
rfbgo: Can this be expressed safely as  `is not`?

lib/ramble/ramble/language/language_base.py
```diff
@@ -68,234 +68,268 @@ def _impossible_when_warning(directive_name, obj_type, obj_name, message, args,
...
+    @classmethod
+    def directive(
+        cls: Type["DirectiveMeta"],
+        dicts: Union[Sequence[str], str, None] = None,
+        init_value: Any = _UNSET,
+        language_type: str = "shared",
+    ) -> Callable[..., Any]:
+        """Decorator for Ramble directives."""
+        if dicts is None or dicts == ():
+            dicts_tuple: Tuple[str, ...] = ()
```
rfbgo, line 403: I don't really understand what this is doing, can you help me understand? Elsewhere this stuff is keyed on `func_name`, right? It feels like this could accidentally turn stuff into a noop?

var/ramble/repos/builtin/base_classes/object-mixin/base_class.py
```diff
@@ -58,11 +59,40 @@ def name(self):
...
+    @property
+    def preferred_version(self) -> Optional[ObjectVersion]:
+        _ = getattr(self, "known_versions", None)
+        return getattr(self, "_preferred_version", None)
+
+    @preferred_version.setter
+    def preferred_version(self, value: Optional[ObjectVersion]):
+        self._preferred_version = value
+
+    def _copy_evaluated_directives(self, target):
```
rfbgo, line 88: Is this able to retain `preferred_version` too?

lib/ramble/ramble/language/language_base.py
```diff
@@ -68,234 +68,268 @@ def _impossible_when_warning(directive_name, obj_type, obj_name, message, args,
...
-    return DirectiveMeta._when_constraints_from_context.pop()
-
-
-def _push_default_args(default_args: Dict[str, Any]) -> None:
-    DirectiveMeta._default_args.append(default_args)
-
-
-def _pop_default_args() -> dict:
-    return DirectiveMeta._default_args.pop()
+_UNSET = object()
```
rfbgo, line 69: Something odd and interesting is leaking in an info print, I (well.. gemini...) thinks it's this:

```
ramble info --type modifier ...
...
default_mode:
    <object object at 0x1052e1e50>
```

rfbgo (commented):
Two random questions: 

1. The bot reports a slow down from this. Is it it true? If so I think there are some places we might be able to improve perf slightly (eg optimized copies)
2. I asked gemini to review this PR wrt "leaking" (it says it will leak). I kind of expect this is something you were already careful about, but I wanted to double check we believe this is indeed safe?

lib/ramble/ramble/language/language_base.py
linsword13: Yep good call, done.

lib/ramble/ramble/language/language_base.py
linsword13: Yeah you are right. Now the code is changed to evaluate eagerly when the dicts are not specified. Hopefully it now makes a bit more sense.

var/ramble/repos/builtin/base_classes/object-mixin/base_class.py
linsword13: Indeed, this was problematic. Now this is changed to read directly from `known_versions`, which is a "registered" dict, so that gets copied properly.

lib/ramble/ramble/language/language_base.py
linsword13: Ah good find, it was a mistake of using `_default_usage_mode` instead of `default_usage_mode`. Corrected.

linsword13:
> > Two random questions:
> > 
> > 1. The bot reports a slow down from this. Is it it true? If so I think there are some places we might be able to improve perf slightly (eg optimized copies)
> > 2. I asked gemini to review this PR wrt "leaking" (it says it will leak). I kind of expect this is something you were already careful about, but I wanted to double check we believe this is indeed safe?
> 
> Great questions:
> 
> For 1. it does seem to cause a slowdown on some perf tests. One reason is that previously the directive initialization cost are part of the startup, which wasn't counted by the benchmarking. And now with the lazy init, these init times are now part of the operations that are being benchmarked. So this is more or less an accounting thing. Other factors include the more works that are needed during copy (as we can no longer rely on the class-level attributes.) I should do a more thorough analysis though to see if any improvements are possible, and at least have a better breakdown of the timing.
> 
> For 2. I did try to look for leaking, but I will do another pass to see where I missed.

The perf has been improved. The leaking should hopefully be addressed as well.

rfbgo:
> The perf has been improved. The leaking should hopefully be addressed as well.

Awesome, thank you for being willing to take those in. I know those were none trivial changes but I think the code is stronger now

rfbgo (approved):

PR #1759, opened 2026-09-02
Add compact repo list printing

rfbgo:
The existing format is great but gets very long:

```
==> 4 applications repositories.
builtin.mock    /Users/robertbird/ramble/injconf/var/ramble/repos/builtin.mock
extra       /Users/robertbird/extra
builtin         /Users/robertbird/ramble/var/ramble/repos/builtin
==> 4 modifiers repositories.
builtin.mock    /Users/robertbird/ramble/injconf/var/ramble/repos/builtin.mock
builtin         /Users/robertbird/ramble/var/ramble/repos/builtin
==> 3 package_managers repositories.
extra      /Users/robertbird/extra
builtin        /Users/robertbird/ramble/var/ramble/repos/builtin
==> 3 workflow_managers repositories.
extra      /Users/robertbird/extra
builtin        /Users/robertbird/ramble/var/ramble/repos/builtin
==> 3 systems repositories.
...
```

This offers a compact alternative:
```
 $ ramble repo list --format=compact
==> 4 repositories.
NAMESPACE     TYPES                    PATH
builtin       all                      /Users/robertbird/ramble/var/ramble/repos/builtin
builtin.mock  applications, modifiers  /Users/robertbird/ramble/injconf/var/ramble/repos/builtin.mock
extra     all (non-utility)        /Users/robertbird/extra
```

Flips the default output to compact, as requested

### DIFF
--- a/lib/ramble/ramble/cmd/repo.py
+++ b/lib/ramble/ramble/cmd/repo.py
@@ -69,8 +69,8 @@ def setup_parser(subparser):
     list_parser.add_argument(
         "--format",
         choices=["full", "compact"],
-        default="full",
-        help="format to be used to print the output (default: full)",
+        default="compact",
+        help="format to be used to print the output (default: compact)",
     )
     ramble.cmd.common.arguments.add_common_arguments(list_parser, ["repo_type"])
 
--- a/lib/ramble/ramble/cmd/repo.py
+++ b/lib/ramble/ramble/cmd/repo.py
@@ -13,6 +13,7 @@ import sys
 import ramble.cmd.common.arguments
 import ramble.config
 import ramble.repository
+import ramble.util.colors as color
 from ramble.util.logger import logger
 
 description = "manage Ramble repositories"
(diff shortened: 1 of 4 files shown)


lib/ramble/ramble/cmd/repo.py
```diff
@@ -65,6 +66,19 @@ def setup_parser(subparser):
         default=ramble.config.default_list_scope(),
         help="configuration scope to read from",
     )
+    list_parser.add_argument(
+        "-c",
+        "--compact",
+        action="store_true",
+        default=False,
```
douglasjacobsen: Is there a reason to not make this default to True?

douglasjacobsen (commented):
One question, otherwise there are merge conflicts that need resolving.

lib/ramble/ramble/cmd/repo.py
rfbgo: Only backwards compatibility. I was thinking we add this and see how folks like it, then flip the switch if they do. Happy to switch it now if you want (but I guess then the arg is more like "verbose")

rfbgo: I think I will drop `--compact` and we can rely on `--format`, and we can discuss making `--format=compact` the default?

rfbgo: rebased and pushed, ptal

douglasjacobsen: Yeah, sounds good. I think we should just make compact the default (because it's crazy wordy with full). I don't think anyone should realisticly be broken by changing the way the repo list is printed out anyway.

rfbgo: Sounds good to me, good idea!

douglasjacobsen (approved):

PR #1761, opened 2026-09-03
workspace concretize: define missing variables and expand compiler specs

chinglung880218:
Fixes an issue where running `ramble workspace concretize` failed with a schema ValidationError (`CommentedMap is not of type 'string'`) when compiler specs contained variable templates (e.g., `{system_compiler_spec}`).

### Background
During compiler definition, calling `ramble.config.add()` is unnecessary because compiler packages are already written in bulk via `update_config(full_software_dict)`. In addition, passing unquoted `{...}` templates into this call causes YAML 1.2 to parse them as inline maps rather than strings, triggering the schema validation error.

### Key Changes
- Remove the redundant `ramble.config.add()` call during concretize, allowing unexpanded spec templates to be cleanly written to `ramble.yaml`.
- Update regression unit tests in `workspace_concretize.py` to assert unexpanded templates are retained.


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -1542,8 +1546,6 @@ def concretize(self, force=False, quiet=False):
 
                     compiler_packages[comp] = False
                     packages_dict[comp].update(info.to_dict(apply_prefix=force_prefix))
-                    for conf in info.config_opts():
-                        ramble.config.add(conf, scope=self.ws_file_config_scope_name())
 
             logger.debug(f"Trying to define packages for {env_name}")
             app_packages = []
```
douglasjacobsen, line 1549: If we remove this here, do you think we need the code for the `config_opts` method in the SoftwareSpec class?

chinglung880218: Good catch! `config_opts` was only used for this purpose, so it is no longer needed after removing previous two lines. I have removed `config_opts` from `SoftwareSpec` in the latest commit.

PR #1765, opened 2026-09-08
Fix (potential) POSIX 8-bit exit code overflow in

rfbgo:
I don't think this is a big deal in practice, but an exit code should not overly 255

### DIFF
--- a/lib/ramble/ramble/cmd/software_definitions.py
+++ b/lib/ramble/ramble/cmd/software_definitions.py
@@ -11,6 +11,7 @@ from typing import Any, Dict, List
 
 from llnl.util.tty.colify import colify
 
+import ramble.error
 import ramble.repository
 import ramble.util.colors as color
 from ramble.util.logger import logger
@@ -53,7 +54,7 @@ def collect_definitions():
         for obj_name in obj_path.all_object_names():
             try:
                 obj_inst = obj_path.get(obj_name)
-            except Exception:
+            except ramble.error.RambleError:
                 continue
             obj_repo = obj_path.repo_for_obj(obj_inst.name)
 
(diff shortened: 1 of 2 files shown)


lib/ramble/ramble/cmd/software_definitions.py
```diff
@@ -50,7 +50,11 @@ def collect_definitions():
 
     for object_type in types_to_print:
         obj_path = ramble.repository.paths[object_type]
-        for obj_inst in obj_path.all_objects():
+        for obj_name in obj_path.all_object_names():
+            try:
+                obj_inst = obj_path.get(obj_name)
+            except Exception:
```
linsword13: Can this be narrowed?

rfbgo: Done!

lib/ramble/ramble/cmd/software_definitions.py
```diff
@@ -50,7 +51,11 @@ def collect_definitions():
 
     for object_type in types_to_print:
         obj_path = ramble.repository.paths[object_type]
-        for obj_inst in obj_path.all_objects():
+        for obj_name in obj_path.all_object_names():
+            try:
+                obj_inst = obj_path.get(obj_name)
+            except ramble.error.RambleError:
```
linsword13: Mostly curious: is this needed because `all_object_names` could be "lying", in the sense that it may return non-existent `obj_name`?

rfbgo: Yea this is pretty unlikely to get tripped, but the idea is that no one single broken object will take it down (here at least, hopefully handled more gracefully elsewhere). Looking at it now I don't know why I no longer have a warning in there, I will add it now

rfbgo: I think this is all round improved now. Thanks for the catch and PTAL

linsword13 (approved):

dapomeroy (approved):

PR #1769, opened 2026-09-08
Lazy load command dependencies

linsword13:
The motivation is that certain dependencies from `ramble.reports` are causing the free-threading to be disabled (for 3.14 and up). Dynamically loading the dependencies help with both startup speed, as well as avoiding the free-threading disablement when running with other unrelated commands.


rfbgo (approved):

rfbgo (approved):
Small note: can we somehow test this functionality to avoid an accidental regression of your fix?

PR #1777, opened 2026-09-10
Add more type checking to Ramble

linsword13:


lib/ramble/ramble/repository.py
```diff
@@ -1015,15 +1016,22 @@ def check(condition, msg):
...
-            if os.path.exists(config_file):
-                self.config_name = config
-                self.config_file = config_file
-        check(self.config_file, "No valid config file found")
-        check(os.path.isfile(self.config_file), f"No {self.config_name} found in '{root}'")
+            candidate = os.path.join(self.root, config)
+            if os.path.exists(candidate):
+                config_name = config
+                config_file = candidate
+                break
```
rfbgo, line 1026: Is this early exit definitely desirable? I'm having a hard time thinking through it but it feels like it could change the precedence ?

lib/ramble/ramble/reports.py
```diff
@@ -677,7 +683,9 @@ def write_inventory(self):
         with open(self.get_inventory_path(), "w+", encoding="utf-8") as f:
             syaml.dump(self.inventory, stream=f)
 
-    def draw(self, perf_measure, scale_var, series, pdf_report, y_label=None):
+    def draw(self, perf_measure, scale_var, series, *args, **kwargs):
+        pdf_report = args[0] if len(args) > 0 else kwargs.get("pdf_report")
+        y_label = args[1] if len(args) > 1 else kwargs.get("y_label", None)
```
rfbgo, line 688: This line gives me pause. Is it definitely the right to fix it? I don't have a concrete suggestion, but it makes me think like we have the wrong interface? I almost prefer this function is renamed if we aren't able to meet the parents api "contract" (over hiding the arg changes). What do you think ?

PR #1779, opened 2026-09-11
tidy up error behavior of `manage modifiers`

rfbgo:
This would previously expose an `IndexError` and now only gives something human readable 

Example of old odd behavior:

```
    $ ramble -D ws workspace manage modifiers --remove --mod-index 99
    ==> Error: Modifier index 99 is outside of the range of modifiers.Use `ramble worksapce manage modifiers --list` to see indices
    ==> Error: list index out of range
```

New:

```
    ==> Error: Modifier index 99 is outside of the range of modifiers. Valid indices are 0-1. Use `ramble workspace manage modifiers --list` to see indices.
```


lib/ramble/ramble/workspace/workspace.py
```diff
@@ -2426,16 +2426,25 @@ def remove_modifier(
...
+                raise RambleWorkspaceError(
+                    f"Cannot remove modifier index {remove_index}. "
+                    "This workspace contains no modifiers."
+                )
+
+            if remove_index < 0 or remove_index >= len(mod_list):
+                raise RambleWorkspaceError(
+                    f"Modifier index {remove_index} is outside of the range of modifiers. "
+                    f"Valid indices are 0-{len(mod_list) - 1}. "
+                    "Use `ramble workspace manage modifiers --list` to see indices."
```
chinglung880218, line 2447: Thanks for cleaning up this error behavior! The explicit exception handling and type guarding make the UX much clearer.

One minor nitpick: when the workspace contains only a single modifier (`len(mod_list) == 1`), the error message currently formats as:
> `... Valid indices are 0-0. ...`

We could polish the phrasing and singular/plural grammar for that single-element edge case:

```suggestion
            if remove_index < 0 or remove_index >= len(mod_list):
                valid_str = (
                    "Valid index is 0."
                    if len(mod_list) == 1
                    else f"Valid indices are 0-{len(mod_list) - 1}."
                )
                raise RambleWorkspaceError(
                    f"Modifier index {remove_index} is outside of the range of modifiers. "
                    f"{valid_str} "
                    "Use `ramble workspace manage modifiers --list` to see indices."
                )
```

*(Optional)* If you think this suggestion is reasonable and decide to adopt this, it might also be nice to add a test case for `len(mod_list) == 1` to ensure the single-item branch is explicitly covered. 

Overall, great fix!

rfbgo: Thanks for the suggestion and the feedback, I incorporated into to the PR. PTAL

chinglung880218 (approved):
LGTM! Thanks for providing the thoughtful commit to tidy up this error behavior.

douglasjacobsen (approved):

PR #1784, opened 2026-09-14
Add variable/vector support to `range`

rfbgo:
Adds support for dynamic vector generation through `range`, eg:

```
    variables:
      foo: [1, 2, 4]
      bar: range(1, {foo})
```

Also adds `pow2_range` 

(this PR is very test heavy [85%], so is not as crazy as the +900 makes it look)


lib/ramble/ramble/expander.py
```diff
@@ -230,11 +223,134 @@ def _maybe(expander, var_name, default=""):
...
+    """
+    if not isinstance(val, str):
+        return False
+    if not supported_list_function_pointers:
+        return False
+    func_names = "|".join(re.escape(fn) for fn in supported_list_function_pointers)
+    return bool(re.search(rf"\b(?:{func_names})\s*\(", val))
+
+
+def extract_var_refs(expr) -> Set[str]:
```
douglasjacobsen: Is there a way we can just use the expansion graph for this? It seems very similar to the parsing we are doing there.

douglasjacobsen (changes requested):
This is awesome, thanks for working on this! It's really cool to have dynamic length experiments without needing to add where statements.

PR #1785, opened 2026-09-15
add spack retry to CI

rfbgo:
This will help us be more resilient to GH flakiness by being able to retry:

```
Step #2 - "ramble-unit-tests": us-central1-docker.pkg.dev/ramble-eng/ramble-repo/ramble-rockylinux-8-spackv1.0.0-python3.7.17:latest
Step #2 - "ramble-unit-tests": branch 'develop' set up to track 'origin/develop'.
Step #2 - "ramble-unit-tests": fatal: unable to access 'https://github.com/spack/spack-packages.git/': Empty reply from server
Step #2 - "ramble-unit-tests": ==> Warning: Error constructing repository '/builder/home/.spack/package_repos/fncqgg4': Failed to clone repository builtin
```

It also DRYs the env setup into a shared script to


share/ramble/cloud-build/ramble-pr-docs.yaml
```diff
@@ -32,22 +32,7 @@ steps:
       - '-c'
       - |
         set -e
```
linsword13, line 34: A bit outside the scope, but I wonder if this `set -e` would defeat the retry somehow?

share/ramble/cloud-build/setup-ci-env.sh
```diff
@@ -0,0 +1,57 @@
...
+# Ensure develop branch exists for comparison in PR builds
+if ! git rev-parse --verify develop >/dev/null 2>&1; then
+  git branch develop origin/develop 2>/dev/null || true
+fi
+
+. /opt/spack/share/spack/setup-env.sh
+
+# Pre-seed cache from container image if available to avoid cloning over network
+if [ -d /root/.spack ] && [ ! -d "$HOME/.spack" ]; then
+  cp -rn /root/.spack "$HOME/.spack" 2>/dev/null || true
```
linsword13: On my local Spack (1.1.0.dev0), calling `setup-env.sh` creates the `$HOME/.spack` directory. I wonder if that means the pre-seed should happen before invoking the script?

share/ramble/cloud-build/ramble-pr-docs.yaml
rfbgo: That is a very good question. My first reaction when I saw this failure was "oh I'm surprised we don't have `set -e`? It should have saved me?". I think you're right here and `set -e` will defeat my intention in this specific yaml

I guess there is an implied choice here between "fail fast" (`-e`) and "retry and fail clearly". Any preference?

share/ramble/cloud-build/setup-ci-env.sh
rfbgo: That's a great question, let me look into it

rfbgo: Thanks for catching this, you are correct I have an order of operations error here. i can fix and push

PR #1787, opened 2026-09-15
Add graphviz to the image builds

tteggelit:
Graphviz has been included in the sphinx config since 2023 but hasn't actually been used in the documentation yet. #1697 does use graphviz and currently fails PR actions because the image doesn't include graphviz. Regardless of whether that PR is accepted, graphviz should be included in the image since it's allowable per the sphinx config. This PR will add it to the cloud-build images.


linsword13 (changes requested):
It might not be sufficient to add the graphviz only to the cloud-build images. The cloud-build images are used by our PR check to validate the doc build. However the actual readthedocs build configuration resides in .readthedocs.yaml, and I think the graphviz dependency might need to be added in there as well, to ensure the actual doc build can work.

tteggelit:
> It might not be sufficient to add the graphviz only to the cloud-build images. The cloud-build images are used by our PR check to validate the doc build. However the actual readthedocs build configuration resides in .readthedocs.yaml, and I think the graphviz dependency might need to be added in there as well, to ensure the actual doc build can work.

I'm not as familiar with .readthedocs.yaml but it looks like it would need to be added to [`build.apt_packages`](https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-apt-packages). Is that right?

linsword13:
> > It might not be sufficient to add the graphviz only to the cloud-build images. The cloud-build images are used by our PR check to validate the doc build. However the actual readthedocs build configuration resides in .readthedocs.yaml, and I think the graphviz dependency might need to be added in there as well, to ensure the actual doc build can work.
> 
> I'm not as familiar with .readthedocs.yaml but it looks like it would need to be added to [`build.apt_packages`](https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-apt-packages). Is that right?

Yes, that looks right

linsword13 (approved):

